PR #11, opened 2017-06-18
Add Travis CI

danrabbit:


benwaffle:
Why `docker pull` and `docker run`? `docker run` will pull images it doesn't have locally

danrabbit:
:man_shrugging: just copying the way we did it in other repos

codygarver:
I think there was a reason I used docker pull, like I couldn't pass all the arguments I wanted. Not sure. But you're welcome to prove me wrong.

Issue #1, opened 2017-06-18
Copying files is slow and slows to a crawl over time for large numbers of files [$200]


elementaryBot:
Copying a lot of files via Pantheon Files becomes slower and slower over time.

I've created 250,000 100-byte files on tmpfs for testing, and kicked off copying to another tmpfs. It started off at speeds over 100Kb/s but halfway through it's just 4Kb/s (!) and dropping.

Profiling with sysprof shows that all this time is spent in g_list_last(),  which probably means that we're abusing a linked list somewhere and that it has to walk the entire list of already copied files, one by one, for each next file copied.

Testcase:
mkdir ~/created-files ~/copy-here
sudo mount -t tmpfs -o size=1G,mode=0777 tmpfs ~/created-files
sudo mount -t tmpfs -o size=1G,mode=0777 tmpfs ~/copy-here
cd created-files
split -b 100 SOME-BIG-FILE
# open Pantheon Files and copy "created-files" folder into "copy-here"

This is a synthetic test case, but I had over 250,000 files during my last backup for OS reinstallation, so this *is* a real-life scenario.

ProblemType: Bug
DistroRelease: elementary OS 0.3
Package: pantheon-files 0.1.5.1+r1680+pkg35~ubuntu0.3.1 [origin: LP-PPA-elementary-os-daily]
ProcVersionSignature: Ubuntu 3.13.0-43.72-generic 3.13.11.11
Uname: Linux 3.13.0-43-generic x86_64
ApportVersion: 2.14.1-0ubuntu3.6
Architecture: amd64
CrashDB: pantheon_files
CurrentDesktop: Pantheon
Date: Sun Dec 21 04:42:10 2014
ExecutablePath: /usr/bin/pantheon-files
GsettingsChanges:

InstallationDate: Installed on 2014-12-10 (10 days ago)
InstallationMedia: elementary OS 0.3 "Freya" - Daily amd64 (20141209)
SourcePackage: pantheon-files
UpgradeStatus: No upgrade log present (probably fresh install)

Launchpad Details: [#LP1404588](https://bugs.launchpad.net/bugs/1404588) Sergey "Shnatsel" Davidoff - 2014-12-21 01:53:22 +0000

elementaryBot:
pantheon-files-daemon has a lot of memory used too (250Mb)

[...]

PR #16, opened 2017-06-20
Improve performance with large number of files.

vjr:
Attempt to address file operation performance issue as described in #1 by using a hashmap instead of list in clipboard manager.


vjr:
Jeremy asked me to elaborate on what testing I'd done, while he is working on the new github Files repo for better testing and to be able to accept (major-ish) PRs, so here is a brief note about it:

I just copied over my compiled pantheon-files binary to /usr/bin after making a backup of the original and tried copying a large number of files, both in a single window with source and destination tabs and also two separate windows. While the original binary becomes unresponsive, my build the copy completes "very fast". Note that I tried with about 10,000 files and then 180,000 files. When I tried with about 360,000 (yes 360 thousand) files, both binaries the folder list view would take a while (spinner) then display "Operation timed out" kind of message in the view. Maybe another area to fix in a different branch at a later time?

Note that I was using keyboard shortcuts CTRL+A, CTRL+C then CTRL+V to select-all, copy then paste because my compiled binary if I right-click it would exit/crash but I'm unsure if that is due to my code change or just some mismatch in my binary with other parts of Files/modules/libraries.

This is on a real disk (SSD) with the ext4 filesystem, not in a ramdisk/tmpfs or anything like that. Also note that the console "rm" command would complain about "too many arguments" if I tried to delete those large number of files with wildcard parameter.

Well this is all I have to post, sorry I didn't/can't do more sophisticated performance timing or memory/speed testing.

jeremypw:
Thanks for the extra information Vishal, that is very helpful.

jeremypw:
I tried compiling and installing your branch in the usual fashion (sudo make install) - it is not normal to just manually copy the binary across.  I could not demonstrate any speed up I am afraid with a couple of tests.  First copied 25K files totalling about 100Mb from one directory to another (both subdirectories of the same folder).  This took 1:02 +- 1 second in each case. Next I tried 14.2K very small files totalling 14.3k.  This took 33+-1 second in each case.  Tests were carried out with a real SSD using the development version of Loki and Files.

I noticed that moving this number of files, rather than copying was impossibly slow.  There are a number of things involved e.g. undo stack, progress indication, updating sidebar available space, the ctags plugin that do not scale well and I think it unlikely that the tweaking clipboard alone will speed things up dramatically.  I based on these results I suggest this issue focus on moving rather than copying files - the target is at least comparability with other common filemanagers (see comments to Launchpad bug).

jeremypw:
To clarify, the tests were carried out using keyboard shortcuts and two tabs each in ListView mode.

vjr:
@jeremypw can you try copying with larger sized files to simulate real world scenario. You mention the total size of your 25k files was just 100 mb?

Lets say someone copying over (making a manual backup) of a large library of music files. I didn't earlier mention that while I tried with around 180k files it was a split of a 3.6 GiB file so each one was about 20 KiB.

I'm not claiming to even have attempted to resolve the whole issue #1 just address one use case - that is, if you are able to repro the copy speed issue with larger individual files.

jeremypw:
OK, I'll try with different conditions as suggested.  I was trying to determine to what extent the problem was related to the number of files or their size (or both). Initial tests indicated it was more closely related to the number of files.

jeremypw:
I have labelled this as "Incomplete" as it is no quantitative performance comparisons are given and it is not clear how the changes made to the clipboard would solve the progressive slow down indicated in the original bug or whether the clipboard processes constitute a substantial percentage of the copy time.

jeremypw:
Would it be faster to use GLib.File.hash () as a key rather than file_to_string ()?

src/ClipboardManager.vala
```diff
@@ -254,8 +256,9 @@ namespace Marlin {
         private string file_list_to_string () {
             var sb = new StringBuilder ("");
             uint count = 0;
-            uint file_count = files.length ();
-            foreach (var file in files) {
+            GLib.List<GOF.File> l_files = files.get_values ();
+            uint file_count = l_files.length ();
```
jeremypw: Slow for large numbers

jeremypw:
This appears to be abandoned so closing for now.

Issue #36, opened 2017-06-22
Files doesn't always show what's actually there (caching issue)


keyilan:
Files does not always show the actual contents of a folder. This happens frequently when removing files via Terminal, or when a folder's contents are changed remotely.

For example, I have a usb thumb drive. It has 10 files on it. I close the window, take the drive out of the computer, take to another computer, put on an 11th file, then bring back to the original computer. When I open the window again, it still only shows 10 files.  `ls` in Terminal it shows 11, and I can open the file that way, but meanwhile the window (which has been closed and opened again) still shows only 10 files.

vjr:
I can't seem to reproduce this issue. You say "not always" - meaning sometimes it works for you and sometimes not? What filesystem and disk device do you have? Any special filesystem or kernel version/configuration apart from the default elementary OS? I actually have an ubuntu mainline kernel version 4.11.5 running at the moment.

keyilan:
Boot volume is ext4. Thumb drive is fat32. Happens on both, and it's consistently reproducible on my end.  When I said "not always" I mean it's not always the case that the folder is accurate, as in sometimes it's not. But I have no trouble reproducing it on this end.

To recreate it, I have computer A and computer B. A has the thumb drive in, with the window open, I see 10 files. I click the eject icon in the sidebar to unmount the volume, then remove the thumb drive, put it in Computer B, copy over a text file, eject it, put it back in computer A, the drive shows up in the sidebar again, I click on it to open that window, and there are 10 files still. In Terminal if I run `ls` all 11 show up.

I just tested it again right now and got the same result. It happens at least every time I try to recreate it.

jeremypw:
Files should not be caching the contents of the thumbdrive under these circumstances. When it ejected, the corresponding directory cache should be destroyed (the view should change to show the home folder).  When it is remounted, the directory cache is recreated and repopulated via gvfs.  I assume you are running an up to date version of FIles on Loki?  I'll try to reproduce this.

jeremypw:
I did not reproduce carrying out the steps you described using an 8GB FAT32 thumbdrive.  Computer A was Elementaryos, computer B was Lubuntu.

vjr:
I was able to repro and I believe I might have a small code fix, will submit a PR shortly for review. As Jeremy, you say, the directory cache should be destroyed so I am calling the `purge_dir_from_cache ()` function in the `Sidebar.vala do_unmount ()` method.

Djaler:
[...]

PR #38, opened 2017-06-22
Fix caching issue with directory listing.

vjr:
Fix for issue #36 - caching issue with directory listing.


vjr:
Was able to reproduce as described in https://github.com/elementary/files/issues/36#issuecomment-310287634 and with this code change the issue appears to be resolved.

jeremypw:
Thanks for the proposed fix, Vishal.  I would prefer to understand a bit more why this is (sometimes) happening, although I have not been able to reproduce it yet.  Then a more general solution may be possible rather than fixing this particular case.  In particular, the directory should be automatically removed from the cache when the reference count drops to 1 (the toggle ref), so it sounds like there is a reference counting problem somewhere.  If you could describe exactly how you reproduced this problem it would help.

jeremypw:
Looks like the directory representing the mounted USB is not necessarily cleared from the cache if it is unmounted externally to Files.  Not sure your fix covers that case. Probably need to use purge_dir_from_cache () anyway when a device is unmounted.

vjr:
Thanks for looking at this Jeremy. In reply to you most recent two comments, the steps to reproduce this works for me as described here: https://github.com/elementary/files/issues/36#issuecomment-310287634

Also I just now checked with two windows of Files open and am able to repro on either one. Plus, the patch resolves (covers the case) of unmounting externally to Files - if I unmount from one of the Files windows the other detects it or if I do a `sudo umount /dev/sda1` in the terminal then both windows appear to detect and purge the directory alright. Looks like the `GLib.VolumeMonitor` and its connected signals are working as expected.

vjr:
Jeremy, maybe you are closing the Files window while trying to reproduce, hence unable to see the issue? Closing the window, of course, clears the directory cache anyway (I don't believe it's being persisted anywhere, is it?). I leave the Files window(s) open while reproducing the issue here.

Also, to your point about understanding the general case, I believe Files directory cache is updated fine if the folder contents are modified (either in another Files window or in the terminal) while mounted, but when unmounted and changed on another machine, there is no way for the VolumeMonitor on the first machine to know. Plus, when plugging the USB stick back in, it mounts to the same mount point (location) which is they key on which the directory is cached, as I understand it.

jeremypw:
This appears to be abandoned so closing for now.

PR #43, opened 2017-06-23
Fix opening URIs and paths via daemon.

vjr:
Escape URIs and enclose in double-quotes in pantheon-files-daemon to handle paths with spaces.


vjr:
Note that I only tested with space and double-quote in the file path via dbus-send command. Space works but file name with double-quote in it still fails - not sure if I am sending the wrong command string format or the `PF.FileUtils.escape_uri ()` method needs tweaking.

vjr:
Switching to using `sanitize_path ()` instead of `escape_uri ()` now gets paths with spaces and double-quotes in them to work, in my minimal testing.

pantheon-files-daemon/CMakeLists.txt
```diff
@@ -34,9 +35,13 @@ PACKAGES
     gio-2.0
     gee-0.8
     sqlite3
+    pantheon-files-core
+    pantheon-files-core-C
 OPTIONS
+    --vapidir=${CMAKE_SOURCE_DIR}/libcore/
```
jeremypw: This line is not required for successful compilation, although it does not harm.

pantheon-files-daemon/FileManager1.vala
```diff
@@ -43,7 +43,7 @@ public class FileManager1 : Object {
         string cmd = "pantheon-files -t";
 
         foreach (string s in uris) {
-            cmd += (" " + s);
+            cmd += (" \"" + PF.FileUtils.sanitize_path (s) + "\"");
```
davidmhewitt: I believe this should actually be single quotes based upon the bash docs:
https://www.gnu.org/software/bash/manual/html_node/Single-Quotes.html#Single-Quotes
https://www.gnu.org/software/bash/manual/html_node/Double-Quotes.html#Double-Quotes

> Enclosing characters in double quotes (‘"’) preserves the literal value of all characters within the quotes, with the exception of ‘$’, ‘`’, ‘\’, and, when history expansion is enabled, ‘!’.

> Enclosing characters in single quotes (‘'’) preserves the literal value of each character within the quotes.

Again, I still haven't tested this yet, but it looks as though there's still a risk of bash interpreting some special characters if they happen to be in the path string.

jeremypw: Bearing mind that the malfactor does not have physical access and the uri is that of a file that has been downloaded with e.g. FireFox I am not sure how much of a risk this is.  If specific examples of dangerous "uris" can be given then maybe the sanitise function can be strengthened.

jeremypw:
https://wiki.gnome.org/HowDoI/DBusApplicationLaunching  might give an indication for another solution, but appears to mean that pantheon-files would have to be started with the --gapplication-service flag on system startup as well as making other significant changes.  I think the currently proposed solution is no less secure than the present situation and should not necessarily be blocked if improvement would mean extensive re-writing of the FileManager1 service code.  A more complete solution can be provided in due course.

pantheon-files-daemon/FileManager1.vala
davidmhewitt: I agree that it's something that's unlikely to be easily exploited. But if there's a way to reduce an attack surface by replacing double quotes with single quotes, then we probably should do it.

jeremypw: Yes, I agree about the single quotes - it just sounded you were questioning the whole approach.

vjr: Thanks for the review comments folks. Let me test both using David's code for replacing single quotes with the percent codes along with the existing methods from `FileUtils`. It may take me a couple of days to follow up.

vjr:
Ah I see the priority has been bumped, I can get back to taking a look at this again, unless one of you has already fixed it, seeing as it's a few lines of code but tricky to get it right?

jeremypw:
I can work on it a bit if you are busy, but I cannot push to the branch I don't think.  I think the solution should be incorporated into FileUtils if it is applicable elsewhere.

vjr:
@jeremypw and @davidmhewitt - what do you guys think of the latest code? it uses a combination of david's suggested code and jeremy's advice to use the `sanitize_path ()` utility method, so i made the unescaping optional (defaults to true for existing behaviour) in the method.

pantheon-files-daemon/FileManager1.vala
```diff
@@ -43,7 +43,8 @@ public class FileManager1 : Object {
         string cmd = "pantheon-files -t";
 
         foreach (string s in uris) {
-            cmd += (" " + s);
+            s = s.replace ("'", "%27").replace ("%", "%%");
+            cmd += (" \'" + PF.FileUtils.sanitize_path (s, null, false) + "\'");
```
jeremypw: Is it necessary to to this before PF.FileUtils.sanitize_path?.   If you do it after, do you need to have the extra parameter stopping the unescaping?  See my PR

vjr: Ah, your code looks more robust and doesn't mess with the existing sanitize method. I didn't realise there needed to be a null check because the return type of the existing sanitize method is `string` and not `string?`. I think we can just close this PR once your PR gets merged to master.

vjr: Would the modified sanitize method with the added `unescape` parameter be useful in the future for other places that need unescaped paths but not necessarily enclosing them in single quotes, hence not requiring the appinfo step of adding? I guess that could be done if the time comes / need arises.

jeremypw: As you say, probably best to modify it as and when the need arises.

jeremypw:
I have blocked this PR for now until the FileManager1 test PR has been reviewed, as that might be merged instead as it also includes unit testing.

jeremypw:
See alternative solution in PR #201 which only changes one file.

jeremypw:
I am closing this as an adequate solution was merged at 493e5d59b1aa773d663b3644bcd8a7917fdd84ce.

PR #45, opened 2017-06-23
Update signal connections on drag/drop tab.

### LINKED ISSUE #40: Home folder artifacts
Sometimes I have a strange bug. In various folder I can see files from ~. You can view example in attached screencast
[screencast.zip](https://github.com/elementary/files/files/1096082/screencast.zip)

jeremypw:
* When a tab is moved between windows, the view container is not destroyed so its reference to the containing window and its signals must be updated.

May fix #40.

Related to https://bugs.launchpad.net/pantheon-files/+bug/1695729.


src/View/ViewContainer.vala
```diff
@@ -29,7 +29,22 @@ namespace Marlin.View {
...
         public bool can_show_folder = false;
-        public Marlin.View.Window window;
+        private Marlin.View.Window _window;
+        public Marlin.View.Window window {
+            get {
+                return _window;
+            }
+
+            set {
+                if (window != null) {
```
matzipan: In this case it doesn't make much difference, but shouldn't this refer to `_window`?

src/View/Window.vala
```diff
@@ -342,6 +342,11 @@ namespace Marlin.View {
                 });
             });
 
+            tabs.tab_added.connect ((tab) => {
```
matzipan: Can you clarify why this was not present in the code before and why it is needed now, and what it actually does?

src/View/ViewContainer.vala
jeremypw: You are probably right.

matzipan:
I tested the branch and it does not fix #1695729

src/View/Window.vala
jeremypw: When tabs are created other than by dropping them onto the notebook, window already knows about it.  Only on the implementation of drag/drop of tabs between windows was it necessary to pay attention to this signal so that the back reference from ViewContainer to window could be updated.

matzipan:
Just realized I was testing master. Whoopsie. Retesting now.

PR #48, opened 2017-06-24
Ignore leave notify when renaming

### LINKED ISSUE #47: Cannot rename bookmark under some circumstances
If the mouse pointer is over the text of the bookmark that the treeview cursor is on then pressing F2 causes renaming to start and immediately end.  Moreover the context menu rename option fails to work for similar reasons.

Seems to be due to recent changes aimed at improving accessibility to keyboard users.

jeremypw:
* Avoid problems renaming with mouse over text due to spurious leave event being generated
* Strip trailing blanks in SideBar.vala

Fixes #47

### DIFF
--- a/src/View/Sidebar.vala
+++ b/src/View/Sidebar.vala
@@ -120,7 +120,7 @@ namespace Marlin.Places {
             }
         }
 
-        /* For cancelling async tooltip updates when update_places re-entered */ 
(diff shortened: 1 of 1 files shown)


matzipan:
I can't seem to be able to trigger rename with f2 right after I have clicked a bookmark (a custom bookmark, not the default ones). I have to hover my mouse over the file view and then back on the sidebar and press f2 to do this.

Issue #46, opened 2017-06-23
tests: failure without elementary-icon-theme


codygarver:
`hicolor-icon-theme` does not suffice, but it should.

jeremypw:
The test code in Files only mentions the following icon specifically: "file:///usr/share/icons/hicolor/16x16/apps/system-file-manager.svg"

Moreover, marlin-icon-info.c returns "hicolor" as the icon theme if the system default icon theme is unset.

It is not immediately clear therefore why the elementary theme is needed.

danrabbit:
Is there a reason that we're testing for the existence of this icon? `system-file-manager` is covered in the freedesktop icon name spec, so it should theoretically always be provided

jeremypw:
The test is not for the existence of the icon (it is assumed to be present).  It is used to test the functionality of marlin-icon-info.  We could alternatively use a built-in icon I suppose, but that would not reproduce normal use as well.

danrabbit:
Oh I see. Yeah that icon wouldn't be provided in hicolor. Hicolor only stores third party icons. You would need either elementary or Adwaita or some other fdo icon set

jeremypw:
Closing as should be fixed by linked merged PR

PR #52, opened 2017-06-25
Vala iconinfo test

jeremypw:
Replace historical marlin-iconinfo test in C with a similar one in Vala.  

* Implement a test data folder and put some images in it.
* Use one builtin image for marlin-iconinfo test.

relates to #46


libcore/tests/MarlinIconInfoTests/MarlinIconInfoTests.vala
```diff
@@ -0,0 +1,46 @@
...
+* You should have received a copy of the GNU General Public
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+* Boston, MA 02111-1307, USA.
+*
+* Authored by: Jeremy Wootten <jeremy@elementaryos.org>
+*/
+
+void add_icon_info_tests () {
+    Test.add_func ("/MarlinIconInfo/goffile_icon_update", goffile_icon_update_test);
```
matzipan: I've noticed that this is what the other tests do as well, is there a reason why this can't be called directly in main?

jeremypw: Only because I prefer to keep main () as simple as possible.

PR #59, opened 2017-06-27
Plugins update visible only

### LINKED ISSUE #58: Pantheon-files daemon maxes out for a long time after loading large folder
After loading  folder containing ca 100,000 files, the pantheon-files daemon is maxed out for several minutes afterwards as well as the files app and the dbus daemon having a high usage.  This is due to the whole folder being compared, one file at a time, against the ctags database.   This is unnecessary; only files that are actually visible need to be updated.

jeremypw:
Fixes #58 

This is first of a series of branches aimed at improving handling of large folders by avoiding applying operations to the whole folder one file at a time when not necessary.

Plugins are only updated for files that are visible or within 50 files of being visible, in the same way thumbnailing now works.

TO TEST:
* create a test folder populate it with a large number of small files by using split -b 1000 [path to a file of size ~100MB] 
* Open files and switch to column view
* Monitor %cpu of pantheon-files and its daemon processes with system-monitor or otherwise.
* Navigate to the test folder
In trunk the daemon process is maxed out for long after the folder is loaded,  In branch this is not the case.

Close the tab with many files as it very slow to navigate away from it due to a different issue.


plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -318,14 +295,23 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
     }
 
     private async void set_color (GLib.List<unowned GOF.File> files, int n) throws IOError {
+        if (files == null) {
```
donadigo: Not sure why this check is needed here. Instead I would expect callers of this method to always pass a valid list of `files`.

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -210,26 +190,25 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
                                                   return false;
                                                   });
+            }
         }
     }
 
     private async void rreal_update_file_info (GOF.File file) {
         return_if_fail (file != null);
-
+warning ("rreal update file info %s", file.uri);
```
donadigo: Fix identation here and don't use `warning ()` here, it's better to use something like `debug ()`.

jeremypw: Looks like left over development code - will remove.

plugins/pantheon-files-ctags/plugin.vala
jeremypw: You are right, the calling function checks this. Will remove.

PR #60, opened 2017-06-27
Better handling of selection changes.

jeremypw:
* Throttle multiple rapid selection changes better.
* Stop selection change handler being called many times during clearing of model.

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -653,9 +653,15 @@ namespace FM {
                 style_context.remove_class (MESSAGE_CLASS);
 
             cancel ();
-            clear ();
+
+            /* Stop unneeded signal handling of old_dir while clearing */
             disconnect_directory_handlers (old_dir);
+            is_frozen = true;
+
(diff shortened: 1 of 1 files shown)


matzipan:
Can you please provide a testcase or information on how to test that this branch does what it does?

jeremypw:
The main effect is to speed up navigating away from a folder with a  very large number of files (10's thousand). The main effect is due to lines 659 and 664. 

The other changes reduce the number of selection_change events emitted and processed but it is questionable whether the extra complexity in on_view_selection_changed () us worthwhile. It was already throttling the events when selections were rapidly changing.  It was just that the original version always called after_selected_files_changed () twice, even with only one event.  It was also found that duplicate events were being emitted due to multiple calls to connect_signals during view creation.

The benefits would mainly be felt with slower hardware.

jeremypw:
Pull request #63 provide a different (better) way of solving the clearing large folder problem anyway, so the benefits of this change would only be noticed if there are other situations where very many rapid selection change events are emitted (and handled).

PR #61, opened 2017-06-27
Do not append to potentially long glist

jeremypw:
* Review where GList is appended and, except where only one or two files long, prepend instead.
* Reverse GList where necessary.

Related to lp:1409468, which is fixed.


libcore/gof-directory-async.vala
```diff
@@ -1030,8 +1033,8 @@ public class Async : Object {
             GLib.File from = pair.index (0);
             GLib.File to = pair.index (1);
 
-            list_from.append (from);
-            list_to.append (to);
+            list_from.prepend (from);
+            list_to.prepend (to);
         }
```
jeremypw: The order of these lists is not important; the files listed are removed to and added from structures (HashTable, ListModel) where the sequence of changes have no effect.

src/View/AbstractDirectoryView.vala
```diff
@@ -765,6 +765,8 @@ namespace FM {
                 list.prepend (file);
             });
 
+            list.reverse ();
+
```
jeremypw: This fixes lp:1409468.

src/View/AbstractDirectoryView.vala
```diff
@@ -2566,8 +2568,10 @@ namespace FM {
                 selected_files.@foreach ((file) => {
                     var goffile = GOF.File.get_by_uri (file.get_display_target_uri ());
                     goffile.query_update ();
-                    action_files.append (goffile);
+                    action_files.prepend (goffile);
                 });
+
+                action_files.reverse ();
```
jeremypw: Prepend followed by reverse () is equivalent to append () as action_files starts off empty.

src/View/Widgets/BreadcrumbsEntry.vala
```diff
@@ -319,7 +319,7 @@ namespace Marlin.View.Chrome {
             if (!drop_data_ready) {
                 drop_file_list = null;
                 foreach (var uri in selection_data.get_uris ()) {
-                    drop_file_list.append (File.new_for_uri (uri));
+                    drop_file_list.prepend (File.new_for_uri (uri));
                     drop_data_ready = true;
```
jeremypw: Dropping onto a breadcrumb adds files to a directory which is not sensitive to the order, so we can prepend without reversing.

jeremypw:
I have commented on the changes (see above).  To test you can check the following still work:

Drag drop selections between folders
Cut & Paste selections between folders
Drag & drop files onto breadcrumbs

and check linked bug is fixed by sending a selection of files to an app that is sequence sensitive using both the context menu and drag & drop - the order of files in the receiving app should be the same as in the source view.

PR #63, opened 2017-06-28
Fix clear large folder

jeremypw:
For folders with a large number of files (currently set to 1000, although that is open to change) when the model needs clearing (e.g. when changing from "show hidden" to "do not show hidden" or when navigating within the same view), instead of using the model·clear () function, which gets very slow for large folders, the model is destroyed and rebuilt.

Steps are taken to avoid unexpected change in the sort order when this done (also fixes bug in trunk where sort order may change on reload folder).

TO TEST:
(A)
1) Create a folder containing a large number of files (e.g ~100,000), using "split -b 100 SOME-10MB-FILE" or otherwise.
2) Show this folder (in ListView preferably for speed).
3) Monitor process cpu usage using gnome-system-monitor, htop or otherwise.
4) When cpu usage drops back to zero, switch on "Show Hidden".
5) When cpu usage drops back to zero, switch off "Show Hidden".

IN TRUNK:  When switching "Show Hidden" off, Files 100% of a cpu core for a significant length of time and may hang (or appear to).
IN BRANCH: When switching "Show Hidden" off, Files uses 100% of cpu core for a much shorter period of time and does not hang.

A similar difference is seen when reloading the folder.

(B)
1) Show two tabs with the same location and sort order (use ListView for ease).
2) Change the sort order in one tab.
3) Switch to other tab and press reload.

IN TRUNK:  The reloaded tab changes sort order.
IN BRANCH: The reloaded tab does not change sort order.


src/View/AbstractDirectoryView.vala
```diff
@@ -312,7 +312,17 @@ namespace FM {
...
+
+            in_trash = slot.directory.is_trash;
+            in_recent = slot.directory.is_recent;
+
             model = GLib.Object.@new (FM.ListModel.get_type (), null) as FM.ListModel;
+            if (in_recent)
+                model.set_sort_column_id (get_column_id_from_string ("modified"), Gtk.SortType.DESCENDING);
+            else if (slot.directory.file.info != null) {
+                model.set_sort_column_id (slot.directory.file.sort_column_id, slot.directory.file.sort_order);
+            }
```
matzipan: Is this change related to this pull request?

src/View/AbstractDirectoryView.vala
```diff
@@ -312,7 +312,17 @@ namespace FM {
                     view.queue_draw ();
                 }
             });
+
+            in_trash = slot.directory.is_trash;
+            in_recent = slot.directory.is_recent;
+
             model = GLib.Object.@new (FM.ListModel.get_type (), null) as FM.ListModel;
+            if (in_recent)
+                model.set_sort_column_id (get_column_id_from_string ("modified"), Gtk.SortType.DESCENDING);
```
matzipan: Missing brackets

src/View/AbstractDirectoryView.vala
```diff
@@ -663,18 +677,44 @@ namespace FM {
...
             connect_directory_handlers (new_dir);
+
+            /* If navigating to new folder reset the sort order as appropriate */
+            if (new_dir.file.uri != old_dir.file.uri) {
+                in_trash = new_dir.is_trash;
+                in_recent = new_dir.is_recent;
+                if (in_recent)
+                    model.set_sort_column_id (get_column_id_from_string ("modified"), Gtk.SortType.DESCENDING);
+                else if (new_dir.file.info != null) {
+                    model.set_sort_column_id (slot.directory.file.sort_column_id, slot.directory.file.sort_order);
```
matzipan: Is this change also related?

src/View/AbstractDirectoryView.vala
```diff
@@ -663,18 +677,44 @@ namespace FM {
             clear ();
             disconnect_directory_handlers (old_dir);
             connect_directory_handlers (new_dir);
+
+            /* If navigating to new folder reset the sort order as appropriate */
+            if (new_dir.file.uri != old_dir.file.uri) {
+                in_trash = new_dir.is_trash;
+                in_recent = new_dir.is_recent;
+                if (in_recent)
```
matzipan: Brackets here?

src/View/AbstractDirectoryView.vala
jeremypw: It's a while since I wrote this but ... I think it is related to the second paragraph of the PR description (unexpected sort order change on clearing)

src/View/AbstractDirectoryView.vala
jeremypw: See comment above

src/View/AbstractDirectoryView.vala
jeremypw: Fixed

src/View/AbstractDirectoryView.vala
jeremypw: Fixed

src/View/AbstractDirectoryView.vala
jeremypw: FIxed

PR #69, opened 2017-07-01
Select all per view

### LINKED ISSUE #66: Select-all shortcut does not work properly when tabs or path changed
The select-all action state is not updated when that active view changes, meaning sometimes no effect on first key press.

jeremypw:
Fixes #66 by having each view handle <Ctrl-a> key press and record its own "all-selected" status.


src/View/AbstractDirectoryView.vala
```diff
@@ -3510,18 +3533,33 @@ namespace FM {
...
             select_all ();
 
-            foreach (Gtk.TreeRowReference r in selected_row_refs) {
-                var p = r.get_path ();
-                if (p != null)
-                    unselect_path (p);
+            if (selected_row_refs != null) {
+                foreach (Gtk.TreeRowReference r in selected_row_refs) {
+                    var p = r.get_path ();
+                    if (p != null)
```
donadigo: Since we modify that code, can you insert braces here?

PR #70, opened 2017-07-01
Fix #65 unselect individual

### LINKED ISSUE #65: Cannot unselect with helper under some circumstances
The behaviour of the helper seems to be different after selecting ALL the files in a folder using Ctrl-A.  It does not deselect the item clicked on.  Affects all views.

jeremypw:
Fixes #65.

The code determining what part of the fileitem was clicked on is reworked and no longer relies on helper coordinates obtained from iconrenderer (which may not refer to the current item when there is a multiple selection).

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -2900,7 +2900,6 @@ namespace FM {
                 return true;
 
             click_zone = get_event_position_info ((Gdk.EventButton)event, out path, false);
-            GOF.File? file = path != null ? model.file_for_path (path) : null;
 
             if (click_zone != previous_click_zone) {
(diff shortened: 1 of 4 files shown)


src/View/AbstractDirectoryView.vala
```diff
@@ -2931,6 +2930,8 @@ namespace FM {
                 if (slot.directory.is_local || NetworkMonitor.get_default ().get_network_available ()) {
                     /* cannot get file info while network disconnected. */
                     GOF.File? target_file;
+                    GOF.File? file = path != null ? model.file_for_path (path) : null;
```
jeremypw: This line was moved here so the function is only called when required instead of at every event.

PR #72, opened 2017-07-03
Filemanager1 refactoring

### LINKED ISSUE #42: Opening paths containing spaces via org.freedesktop.FileManager1 fails.
A properly URI encoded path such as the following sent to the `ShowItems` method causes Files to try and open an incorrect path.
`file:///home/david/Downloads/Torrent%20Downloads/podcast.mp3`

For the example above Files tries and fails to open the following path:
`/home/david/Downloads/Torrent0Downloads/podcast.mp3/`

I have tested with Nautilus to check that the path I'm sending is well formed and it had no issue.

jeremypw:
Fixes #42 

Refactor FileManager1 interface to simplify it.  Avoids use of command line  by having the interface provided by the main app and not the daemon.  If not already running, the app is launched automatically by DBus and then adds a tab for each uri received by the interface.  

An additional launch option is provided to avoid an extra "Home" tab appearing when launched by org.freedesktop.FileManager1 interface.

It was necessary to scrap the unit tests as it is no longer possible to run the interface independently of the app. Ways of reimplementing these will be investigated for a future PR.

The sanitizing already included in the main app fixes the linked issue, avoiding an additional function


src/main.vala
```diff
@@ -25,6 +44,12 @@ public static int main (string[] args) {
     Environment.set_application_name (APP_NAME);
     Environment.set_prgname (APP_NAME);
 
+    Bus.own_name (BusType.SESSION, "org.freedesktop.FileManager1", BusNameOwnerFlags.REPLACE,
```
donadigo: Can we get this into the already available `dbus_register` of the `Marlin.Application` method?
The `GLib.Application` provides a way to register any needed objects in the application itself implementing the `dbus_register` and `dbus_unregister` methods like that:

```
public override bool dbus_register (DBusConnection connection, string object_path) throws Error {
    base.dbus_register (connection, object_path);

    // own the FileManager1 interface here
}
```

```
public override void dbus_unregister (DBusConnection connection, string object_path) {
    // Unown the FileManager1 interface here if it exists

    base.dbus_unregister (connection, object_path);
}

jeremypw: I tried this but either I did something wrong or it does not work unless the app is already running.  If not, the app get launched OK but misses the dbus message.  The FM1 service needs to be up and running (in main ()) before the GApplication launches, I think.

matzipan:
Jeremy Wootten [11:39 AM] 
@matzipan: Donadigo said on slack (>10,000 messages ago I think) he was going to look into reimplementing PR#72.

jeremypw:
Waiting for merging of preparatory branches.  No longer high priority as a quick fix for #42 has been proposed.

jeremypw:
Quick fix may be sufficient.

PR #74, opened 2017-07-04
Remove About dialog

danrabbit:


danrabbit:
Yeah that'll have to be followed up in Granite.Application I guess

PR #78, opened 2017-07-08
Review and test themed icon caching

jeremypw:
This branch implements a test for proper operation of caching of themed icons, checking that they are dropped from the cache appropriately, at least under test conditions. Some changes to the caching code was required to achieve this.

This is intended as first part of a series related to memory consumption.

In addition, the code in marlin-icon-info.c is tidied a little as it may be converted to Vala later.


tintou:
Looks sane

PR #77, opened 2017-07-07
Fix #76 - Compile with option

jeremypw:
The compile options "MODULE_ONLY" and "LIB_ONLY" were previously provided but now fail to compile.  This branch modifies some CMakeLists to allow compilation to complete.

Some missing lines from the CMakeLists.txt for MarlinIconInfoTests.vala were also inserted (although under normal circumstances this did not lead to failure to build, e.g. in Travis CI, the linked issue indicated that failure was possible under some circumstance).

Fixes #76 (to be confirmed on a system where failure occurs)

PR #81, opened 2017-07-09
add po folder to cmake config

btkostner:
Adds the `po` directory to cmake, enabling translations again.


danrabbit:
hm looks like it was a causality of https://github.com/elementary/files/pull/77/files 

I guess maybe this should be inside that if eh @jeremypw ?

PR #88, opened 2017-07-21
Add filemanager1 ctest

jeremypw:
This adds ctesting back to the filemanager1-test branch and should therefore not be merged before that one.

A limited interface to the app is used instead of the app itself, allowing a simple dummy app to be substituted by the unit test when DBus is not available (on Travis CI).


jeremypw:
Will not refactor at this stage.

PR #92, opened 2017-07-22
Fix segfaults in Column view when there are no selected files

### LINKED ISSUE #91: Segfaults in Column view
This seems to happen randomly while opening folders from the Column view.
Here's the stacktrace:
```
Thread 1 "pantheon-files" received signal SIGSEGV, Segmentation fault.
0x0000000000441530 in fm_column_view_real_handle_primary_button_click ()
(gdb) bt
#0  0x0000000000441530 in fm_column_view_real_handle_primary_button_click ()
#1  0x00000000004510ab in fm_abstract_directory_view_real_on_view_button_press_event ()
#2  0x00007ffff6876fac in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#3  0x00007ffff59c3fa5 in g_closure_invoke () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#4  0x00007ffff59d5fc1 in ?? () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#5  0x00007ffff59de7f9 in g_signal_emit_valist () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#6  0x00007ffff59df08f in g_signal_emit () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#7  0x00007ffff69b4c3c in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#8  0x00007ffff68743be in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#9  0x00007ffff68761bc in gtk_main_do_event () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#10 0x00007ffff63e3d92 in ?? () from /usr/lib/x86_64-linux-gnu/libgdk-3.so.0
#11 0x00007ffff56ed197 in g_main_context_dispatch () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#12 0x00007ffff56ed3f0 in ?? () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#13 0x00007ffff56ed49c in g_main_context_iteration () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#14 0x00007ffff7820d6d in g_application_run () from /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#15 0x00007ffff7b20bd5 in granite_application_run () from /usr/lib/x86_64-linux-gnu/libgranite.so.3
#16 0x000000000042bdfd in _vala_main ()
#17 0x00007ffff496f830 in __libc_start_main (main=0x426f80 <main>, argc=1, argv=0x7fffffffddd8, 
    init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffddc8)
    at ../csu/libc-start.c:291
#18 0x0000000000426fb9 in _start ()
```

donadigo:
Fixes #91.

Check if we have any selected files before doing any operations on them. It looks like this crash was happening because of unfocusing the ColumnView and making the selection `null`. There is still some issues with the column view though that should be resolved.

### DIFF
--- a/src/View/ColumnView.vala
+++ b/src/View/ColumnView.vala
@@ -120,40 +120,45 @@ namespace FM {
         }
 
         protected override bool handle_primary_button_click (Gdk.EventButton event, Gtk.TreePath? path) {
-            unowned GOF.File file = selected_files.data;
-            bool is_folder = file.is_folder ();
-
-            selected_folder = null;
-
-            if (!is_folder || !Preferences.settings.get_boolean ("single-click")) {
-                return base.handle_primary_button_click (event, path);
-            }
-
-            selected_folder = file;
             bool result = true;
 
-            if (event.type == Gdk.EventType.BUTTON_PRESS) {
-                /* Ignore second GDK_BUTTON_PRESS event of double-click */
-                if (awaiting_double_click)
-                    result = true;
-                else {
-                    /*  ... store clicked folder and start double-click timeout */
-                    awaiting_double_click = true;
-                    is_frozen = true;
-                    double_click_timeout_id = GLib.Timeout.add (drag_delay, () => {
-                        not_double_click (event, path);
-                        return false;
-                    });
+            if (selected_files != null) {
+                unowned GOF.File file = selected_files.data;
+                bool is_folder = file.is_folder ();
+
+                selected_folder = null;
+
+                if (!is_folder || !Preferences.settings.get_boolean ("single-click")) {
+                    return base.handle_primary_button_click (event, path);
                 }
-            } else if (event.type == Gdk.EventType.@2BUTTON_PRESS) {
-                should_activate = false;
-                cancel_await_double_click ();
 
-                if (selected_folder != null)
-                    load_root_location (selected_folder.get_target_location ());
+                selected_folder = file;
+                
 
-                result = true;
+                if (event.type == Gdk.EventType.BUTTON_PRESS) {
+                    /* Ignore second GDK_BUTTON_PRESS event of double-click */
+                    if (awaiting_double_click)
+                        result = true;
+                    else {
+                        /*  ... store clicked folder and start double-click timeout */
+                        awaiting_double_click = true;
+                        is_frozen = true;
+                        double_click_timeout_id = GLib.Timeout.add (drag_delay, () => {
+                            not_double_click (event, path);
+                            return false;
+                        });
+                    }
+                } else if (event.type == Gdk.EventType.@2BUTTON_PRESS) {
+                    should_activate = false;
+                    cancel_await_double_click ();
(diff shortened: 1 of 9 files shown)


src/View/ColumnView.vala
```diff
@@ -120,40 +120,45 @@ namespace FM {
...
+                bool is_folder = file.is_folder ();
 
-            if (!is_folder || !Preferences.settings.get_boolean ("single-click")) {
-                return base.handle_primary_button_click (event, path);
-            }
+                selected_folder = null;
 
-            selected_folder = file;
-            bool result = true;
-
```
jeremypw: This is (I think) a simpler solution:
```
 protected override bool handle_primary_button_click (Gdk.EventButton event, Gtk.TreePath? path) {
            unowned GOF.File? file = null;
            if (selected_files != null) {
                file = selected_files.data;
            }

            selected_folder = null;

            if (file == null || !file.is_folder () || !Preferences.settings.get_boolean ("single-click")) {
                return base.handle_primary_button_click (event, path);
            }

            selected_folder = file;
            bool result = true;

```

the rest of the function being unaltered.  The base function just returns true so a null selection there does not matter.

src/View/ColumnView.vala
```diff
@@ -120,40 +120,45 @@ namespace FM {
...
+                            not_double_click (event, path);
+                            return false;
+                        });
+                    }
+                } else if (event.type == Gdk.EventType.@2BUTTON_PRESS) {
+                    should_activate = false;
+                    cancel_await_double_click ();
+
+                    if (selected_folder != null)
+                        load_root_location (selected_folder.get_target_location ());
```
jeremypw: Could insert braces here while we are fixing this function

src/View/ColumnView.vala
```diff
@@ -120,40 +120,45 @@ namespace FM {
...
-                if (selected_folder != null)
-                    load_root_location (selected_folder.get_target_location ());
+                selected_folder = file;
+                
+
+                if (event.type == Gdk.EventType.BUTTON_PRESS) {
+                    /* Ignore second GDK_BUTTON_PRESS event of double-click */
+                    if (awaiting_double_click)
+                        result = true;
+                    else {
```
jeremypw: Could insert braces here while we are fixing this function.

src/View/ColumnView.vala
```diff
@@ -154,6 +158,8 @@ namespace FM {
 
                 result = true;
             }
+
+
```
jeremypw: Extra line

PR #106, opened 2017-08-04
Replace org.pantheon.files with io.elementary.files

jeremypw:
* Affects settings schemas
* Affects DBus names
* Affects App ID

First of two related PRs


danrabbit:
It should be noted that changing the gsetting schema means that all settings will effectively be reverted to defaults. That's probably okay since it doesn't seem we're saving anything that could really be considered data loss, but I thought I'd mention it.

danrabbit:
Okay so it sounds like the plan is that @codygarver is going to create patches in the Loki packaging to keep the .desktop name from breaking there. So this is gonna be `Blocked` until he can do that

Issue #114, opened 2017-08-10
Crash when drag/drop file into folder with auto open


jeremypw:
To reproduce:

Drag a file over a folder in the same view and hover until that folder opens.
Drop the file
Files crashes.

Does not occur when dragging e.g. into another tab.

PR #115, opened 2017-08-10
Fix drop auto open

jeremypw:
See #114 

A null check is used in DndHandler to prevent similar bugs (dropping an empty file list) causing a crash.
As an extra check, a drop is not acted on unless the "drop_data_ready" is true.

The root cause in this case is fixed by moving a couple of lines from the "drag leave" handler to the "drag end" handler.

See linked issue for test procedure.


jeremypw:
Marking high priority because of possible crash.

PR #123, opened 2017-08-21
Clear large folder rapidly

### LINKED ISSUE #54: Slow to navigate away from view with many files.
If displaying a folder with many files (i.e. 10's of thousands) it is very slow to navigate away from this folder while keeping the same view.  This is because the view keeps the same model and this is very inefficiently cleared.

jeremypw:
Fixes #54 

Instead of deleting every row when clearing the model, the model is recreated.  This greatly speeds up reloading and navigating away from large folders (10's or 100's thousands of files) as well as switching "show hidden" off.

For simplicity the same is done for small folders as well as there seems no discernible downside.

TO TEST:
Monitor FIles cpu usage using system-monitor or htop.  Do not start a new step until cpu usage drops to zero.

1) Create a folder containing several 10's of thousand or 100,000 files (recommend using ListView for > 10,000 files).  Some of the files should be hidden,
2) Navigate into folder 
3) Reload the folder
4) Switch "show-hidden" from "on" to "off"
5) navigate away from folder while monitoring cpu usage.

Compared to trunk these operations are quicker and use less cpu.

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -637,7 +637,7 @@ namespace FM {
             freeze_tree ();
          }
 
-        protected void after_loading_directory (GOF.Directory dir) {
+        protected void after_loading_directory (GOF.Directory.Async dir) {
             dir.file_loaded.disconnect (on_directory_file_loaded);
             dir.done_loading.disconnect (on_directory_done_loading);
         }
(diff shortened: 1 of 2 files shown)


matzipan:
Hey @jeremypw  you think we can set up some test codes which do their own setup and metrics by themselves so we can quantify these improvements? I think it's gonna make it more easy to get changes liek this merged.

jeremypw:
I'll see what I can do.

jeremypw:
Closed pending merging of branches introducing changes impacting on this.

PR #143, opened 2017-09-25
Spelling: WebDAV, as administrator, download folder

comradekingu:
As per https://en.wikipedia.org/wiki/WebDAV


jeremypw:
I suspect at least some of these changes should be made by correcting the translation(s) as required rather than changing the message IDs.  Seeking advice of i18n team.

jeremypw:
In any case you would need to change the text in code not in the pot file.

PR #144, opened 2017-09-25
Enable compilation under vala 0.38.1

### LINKED ISSUE #139: will not compile under vala 0.38.1 +
plugin.vala:111:2: error: non-void function 'marlin_plugins_ctags_add_entry' should return a value [-Wreturn-type]
        g_return_if_fail (_tmp0_ != NULL);

### LINKED ISSUE #157: Tag colors not always displayed initially
For some folders, when first navigated into the tag colors are not displayed until the view is redrawn through e.g. being focussed in.  Other folders show the tag colors immediately even if focussed out.

It seems this is related to whether or not the folder contains a thumbnailable icon and indicates a race between initial drawing and updating of the file items by the plugin.

jeremypw:
Fixes #139. Fixes #157

* Make add_entry() a void function and simplify.
* Use a `GenericArray<Variant>`
* Signal file icon change after color update.
* Queue redraw after file icon change
* Throttle redraw calls.

### DIFF
--- a/plugins/pantheon-files-ctags/plugin.vala
+++ b/plugins/pantheon-files-ctags/plugin.vala
@@ -107,16 +107,18 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
         ignore_dir = f_ignore_dir (directory.uri);
     }
 
-    private Variant add_entry (GOF.File gof) {
+    private void add_entry (GOF.File gof, ref GenericArray<Variant> entries) {
         return_if_fail (gof != null);
 
-        char* ptr_arr[4];
-        ptr_arr[0] = gof.uri;
-        ptr_arr[1] = gof.get_ftype ();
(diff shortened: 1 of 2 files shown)


benwaffle:
What's the compilation error under 0.38?

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -107,16 +107,18 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
         ignore_dir = f_ignore_dir (directory.uri);
     }
 
-    private Variant add_entry (GOF.File gof) {
+    private void add_entry (GOF.File gof, ref GenericArray<Variant> entries) {
```
tintou: You don't need the `ref` here, GenericArray is a pointer already

PR #151, opened 2017-10-05
Multi-window refreshing (Fix #96)(Fix #148)

### LINKED ISSUE #96: Auto strange refreshing folder view while in other windows file manager
I discover that when I go back from one folder in _window1_ then _window2_ will refresh with home files + _windows2_ real files. Not always and not always in the same way.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/47703621-auto-strange-refreshing-folder-view-while-in-other-windows-file-manager?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #148: Crash when refreshing multiple windows rapidly
To reproduce:

Open Files showing Home folder in one tab.
Duplicate the tab
Drag the second tab to create a new window
Return to the first window
Press "Reload" button twice or more

FIles will produce critical warnings and may crash.

jeremypw:
* Cache Async earlier during creation
* All cache handling in static methods
* Cache lookups return unowned.
* Discontinue separate GOF.File cache.
* Consequential changes to gof-file.c and AbstractDirectoryView
* Reloading controlled by Async only; simplify Slot.
* Reformat code touched

The interconnected design of Files resulting in reference counting errors and race conditions made it difficult to fix these issues without introducing regressions that needed further fixing.

Part of the solution was to have GOFDirectoryAsync solely responsible both for caching GOFDirectoryAsync GOFFile objects, and reloading, and making sure GOFFiles were not needlessly destroyed and recreated during reloading.

The opportunity was taken to remove some old commented out and unused code.

TO TEST: See https://github.com/PerfectCarl/elementary-files-testing/blob/master/multiwindow.md
tests 01 and 02.


libcore/tests/GOFFileTests/GOFFileTests.vala
```diff
@@ -153,7 +122,9 @@ void new_symlink_local_test () {
 
     file.query_update ();
     assert (file.info != null);
-    assert (file.get_symlink_target () == path);
+
+    string target = file.get_symlink_target ().dup ();
```
tintou: you don't have to explicitly use .dup (), assigning an `unowned string` to a `string` should do it automatically

tintou:
Looks good :)

libcore/tests/GOFFileTests/GOFFileTests.vala
jeremypw: Yes, now that I have corrected the .vapi for GOFFile, I guess you are right. I'll change it and check for any regressions.

jeremypw:
"Regression" in row expand in debug mode found due to stricter assertion.  Investigating.

jeremypw:
* Removed unnecesssary dup ()
* Null files in model expected if there are expanded empty directories in list view.

jeremypw:
Master merges have caused regression in ViewContainer destruct - to be fixed.

jeremypw:
Will split out separate branches were possible

PR #156, opened 2017-10-10
Changed colours for file tags

### LINKED ISSUE #155: File tagging should use elementary colours.
The file tagging colours should be from the elementary colour palette and not tango's.
file-highlights

TraumaD:
Fixes #155 

Changed file tag colours to colours from elementary's palette.
screenshot from 2017-10-10 16 44 37
screenshot from 2017-10-10 16 39 31

### DIFF
--- a/libcore/gof-preferences.vala
+++ b/libcore/gof-preferences.vala
@@ -21,7 +21,7 @@ namespace GOF {
 
     public class Preferences : Object {
 
-        public const string TAGS_COLORS[10] = { null, "#fce94f", "#fcaf3e", "#997666", "#8ae234", "#729fcf", "#ad7fa8", "#ef2929", "#d3d7cf", "#888a85" };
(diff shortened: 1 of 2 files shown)


danrabbit:
I wonder if it would be better to use Material brown as a stopgap. It would be closer to what users are expecting. It looks like you're using 100 to make sure the black text is legible. Maybe Material Brown 100 (#D7CCC8) or 200 (#BCAAA4) would work here

TraumaD:
That sounds good to me. I'll play around with material brown. I can make the changes and update the PR when I get back to my computer.

TraumaD:
Does anyone have a preference on which brown to use?
tag-colours
tag-dialog

danrabbit:
200 is probably more distinctly brown and not just grey looking

TraumaD:
I was thinking the same. 100 looks a little too grey. just wanted to run it by someone else before going for it.

jeremypw:
@TraumaD   Thanks for your contribution! May it be the first of many ;-)

PR #179, opened 2017-10-20
Focus rectangle - adjust padding; fix icon clearance & occasional truncation.

### LINKED ISSUE #82: Excessive padding in focus rectangle
Since r2581, the padding in the focus rectangle (Icon View) has increased and this causes touching/overlap with the icon for some icons. Padding should be reduced to original value.

jeremypw:
Fixes #82 
* Reduce number of variables used by TextRenderer, included deprecated.
* Fix some truncation problems in ListView for long files names
* Fix clearance between focus rectangle and icon for tall or wide icons.
* Reduce padding between text and focus rectangle.
* Remove variable text size code - it has no effect under default conditions.

TO TEST
Verify that there is no truncation or misposition of focus rectangle and text for all views, zoom-levels and name lengths in both LTR and RTL environments.

Verify that name editing as unaffected.  NOTE: There is a pre-existing bug for renaming in List View in RTL environment which is not fixed in this branch.


danrabbit:
I can confirm that this works as expected on HiDPI in icon, list, and column views with varying title lengths in LTR and RTL. Nice work :)

src/TextRenderer.vala
```diff
@@ -20,34 +20,49 @@ namespace Marlin {
...
+
+        private Marlin.ZoomLevel _zoom_level;
+        public Marlin.ZoomLevel zoom_level {
+            get {
+                return _zoom_level;
+            }
+
+            set {
+                var icon_size = Marlin.zoom_level_to_icon_size (value);
+                border_radius = 5 +  icon_size / 40;
```
danrabbit: accidental double space here

PR #194, opened 2017-11-04
Ignore expanded folders except in ListView

### LINKED ISSUE #158: Folder icon can be shown as "expanded" in icon view.
To reproduce:

* Open two windows at the same location which has subfolders.
* Have one IconView, the other ListView.
* Expand one subfolder in ListView
* Hover over that subfolder in the IconView
* The icon shows in the "expanded" form.
* In the ListView, navigate away from the folder without first collapsing the subfolder.
* The folder continues to show as "expanded" in IconView even if refreshed.

jeremypw:
Fixes #158


jeremypw:
Withdrawn in favour of simpler solution  #215

PR #196, opened 2017-11-04
hidpi icons in location bar

markodp:
This fixes blurry icons in location bar, similar to my previous PR https://github.com/elementary/files/pull/192, but for hidpi.

At 2x scale, icons are loaded with twice the size to pixbuf. When those icons are drawn, cairo.context is scaled appropriately.

### DIFF
--- a/libwidgets/Chrome/BreadcrumbIconList.vala
+++ b/libwidgets/Chrome/BreadcrumbIconList.vala
@@ -33,11 +33,13 @@ namespace Marlin.View.Chrome {
 
         private GLib.List<BreadcrumbIconInfo?> icon_info_list = null;
         private Granite.Services.IconFactory icon_factory;
-        Gtk.StyleContext context;
+        private Gtk.StyleContext context;
+        public int scale { public get; private set; }
 
         public BreadcrumbIconList (Gtk.StyleContext _context) {
             context = _context;
             icon_factory = Granite.Services.IconFactory.get_default ();
+            scale = context.get_scale ();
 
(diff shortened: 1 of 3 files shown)


libwidgets/Chrome/BreadcrumbIconList.vala
```diff
@@ -33,11 +33,13 @@ namespace Marlin.View.Chrome {
...
         private GLib.List<BreadcrumbIconInfo?> icon_info_list = null;
         private Granite.Services.IconFactory icon_factory;
-        Gtk.StyleContext context;
+        private Gtk.StyleContext context;
+        public int scale { public get; private set; }
 
         public BreadcrumbIconList (Gtk.StyleContext _context) {
             context = _context;
             icon_factory = Granite.Services.IconFactory.get_default ();
+            scale = context.get_scale ();
```
tintou: I wouldn't cache the scale value, because it might change (the stylecontext is always updated)

markodp: That's why it's cached, because i wasn't sure if it can change. If i load icon with 2x scaling, i have to draw it with the same scaling it was loaded with. Caching value just makes sure icon isn't drawn too small or too big.

Do you know how scaling could change while app is running? If that's possible then i would have to detect it and reload icons.

jeremypw: Manually editing org.gnome.desktop.interface.scaling-factor is possible while the app is running. Or running the app under the environment GTK_DEBUG=interactive allows rescaling.

markodp:
Ok, i have managed to reload icons when scaling factor changes.

Additionally i had to change style context state when loading icons because i noticed they are loaded in gray color when scale factor changes and window is in background - they should be black.

PR #199, opened 2017-11-06
Prepare to refactor FileManager1 interface (pt 1)

jeremypw:
* Add extra check to sanitize_path in FileUtils

Aimed at reducing the diff for the main PR #72, facilitating review


jeremypw:
Quick fix now merged is sufficient for now.

PR #200, opened 2017-11-06
Prepare for FileManager1 interface refactor (pt2)

jeremypw:
* Add new option used by FileManager1 interface.

Aimed at reducing diff for main PR #72


jeremypw:
May be submitted later

PR #201, opened 2017-11-07
Handle escaped uris for creating AppInfo in FileManager1.vala

### LINKED ISSUE #42: Opening paths containing spaces via org.freedesktop.FileManager1 fails.

jeremypw:
Fixes #42 

To test:

After installing and killing any existing Files daemon, open a terminal and enter:

```
dbus-send --session --print-reply --reply-timeout=120000 --type=method_call \
--dest='org.freedesktop.FileManager1'   '/org/freedesktop/FileManager1' \
org.freedesktop.FileManager1.ShowFolders   array:string:"[path to be tested]" string:"TESTID"
```
Replace `[path to be tested]` with the path to an existing file or folder with spaces in the name.  The spaces may or may not be escaped (%20).

The branch should also cope with other unusual characters in the name although it cannot be guaranteed that all weird paths will be handled properly.

### DIFF
--- a/pantheon-files-daemon/FileManager1.vala
+++ b/pantheon-files-daemon/FileManager1.vala
@@ -43,7 +43,7 @@ public class FileManager1 : Object {
         StringBuilder sb = new StringBuilder ("io.elementary.files -t");
         foreach (string s in uris) {
                 sb.append (" ");
-                sb.append (prepare_uri_for_appinfo_create (s));
+                sb.append (FileManager1.prepare_uri_for_appinfo_create (s));
         }
 
         try {
@@ -59,7 +59,7 @@ public class FileManager1 : Object {
         }
     }
 
-    private string prepare_uri_for_appinfo_create (string uri, bool allow_utf8 = true) {
+    private static string prepare_uri_for_appinfo_create (string uri, bool allow_utf8 = true) {
         string rc = reserved_chars.replace("#", "").replace ("*","");
         string? escaped_uri = Uri.escape_string ((Uri.unescape_string (uri) ?? uri), rc , allow_utf8);
         return (escaped_uri ?? "").replace ("%", "%%");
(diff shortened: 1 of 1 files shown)


pantheon-files-daemon/FileManager1.vala
```diff
@@ -39,29 +40,28 @@ public class FileManager1 : Object {
...
 
-        if (pf_app_info != null) {
-            try {
-                pf_app_info.launch (null, null);
-            } catch (Error e) {
-                var msg = "Unable to open item or folder with command %s. %s".printf (cmd, e.message);
-                throw new IOError.FAILED (msg);
-            }
-        }
+    private string prepare_uri_for_appinfo_create (string uri, bool allow_utf8 = true) {
```
donadigo: This should be static.

pantheon-files-daemon/FileManager1.vala
```diff
@@ -18,6 +18,7 @@
 
 [DBus (name = "org.freedesktop.FileManager1")]
 public class FileManager1 : Object {
+    const string reserved_chars = (GLib.Uri.RESERVED_CHARS_GENERIC_DELIMITERS + GLib.Uri.RESERVED_CHARS_SUBCOMPONENT_DELIMITERS);
```
donadigo: `reserved_chars` should be uppercase: `RESERVED_CHARS` for constants.

jeremypw:
Since we are having to fix at least two reports we might as well fix as many corner cases as possible, so long as it does not significantly complicate the code.  There is a custom RESERVED_CHAR string already in the main code but the daemon does not have access to that.

pantheon-files-daemon/FileManager1.vala
jeremypw: modified to static string with replacements already in place.

jeremypw:
I have tested this with a (legal) filename containing spaces, Japanese characters, and punctuation including %.

PR #207, opened 2017-11-09
Add error dialog for failed mount from Sidebar.

### LINKED ISSUE #204: No information when I can't mount a drive
If for some reason I cannot mount a drive (i.e. it's an exfat partition and I don't have exfat support installed, it's a Windows partition that is locked, etc.), I do not get any information as to why I cannot mount it and Files silently fails.

jeremypw:
FIxes #204 

Also corrects construction of similar message for start drive error dialog.

### DIFF
--- a/src/View/Sidebar.vala
+++ b/src/View/Sidebar.vala
@@ -1353,7 +1353,8 @@ namespace Marlin.Places {
                     }
                 }
                 catch (GLib.Error error) {
-                    warning ("Error mounting volume %s: %s", volume.get_name (), error.message);
+                    var primary = _("Error mounting volume %s").printf (volume.get_name ());
+                    Eel.show_error_dialog (primary, error.message, null);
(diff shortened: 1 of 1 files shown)


davidmhewitt:
Can confirm this adds an error dialog and I'm happy with the code.

I'm wondering if this is a candidate for having the error details in an expander and possible "Terminal Styled" like the [recent AppCenter PR](https://github.com/elementary/appcenter/pull/530) as the error dialog I just got was quite unwieldy:

image

jeremypw:
Thanks for the comments.  My only concern is that changing just this dialog to one based on Granite.MessageDialog would result in inconsistency with Eel dialogs used in many other places. It would mean committing to a series of PRs replacing them all.

jeremypw:
I have created a project to upgrade Files Eel dialogs.

PR #215, opened 2017-11-13
Use folder-open as an icon when Gtk.CellRendererState is EXPANDED

### LINKED ISSUE #195: ListView can show open-folder icon when not expanded
To reproduce:

* Open two windows at the same location, with subfolders, both in ListView
* In one window expand the subfolder in the first row
* In the other other window, the subfolder remains unexpanded but the icon shows "folder-open"

For some reason this only affects the first row.

* Open the same subfolders in the two windows
* Open a subfolder of that subfolder in one window
* In the other window, the sub-sub-folder remains unexpanded but shows a "folder-open" icon

obyknovenius:
resolve #195

### DIFF
--- a/src/IconRenderer.vala
+++ b/src/IconRenderer.vala
@@ -102,8 +102,11 @@ namespace Marlin {
                 flags |= Gtk.CellRendererState.PRELIT;
                 special_icon_name = "folder-drag-accept";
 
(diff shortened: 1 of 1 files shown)


jeremypw:
Also fixes #158 
Will remove corresponding PR as this is simpler!
Thanks! :-)

PR #216, opened 2017-11-14
Simplify ViewContainer path/mode/container changes

### LINKED ISSUE #236: Remove autoswitch to Icon mode
Files is intended to switch the view mode automatically to IconView when navigating to a directory with the folder name "/icons/" or /.icons/" in the path, although this seems to be broken at the moment and no issue has been raised about it.

I propose to remove this feature since it has some flaws and it is uncertain how useful it is. This would simplify the code to some extent.  Fixing the flaws would significantly complicate the code.

Main flaws:
* Going back in history does not reverse the switch.
* The switch occurs whether or not the folder being viewed actually contains icons
* It is not possible for the user to switch this behaviour off.

jeremypw:
Fixes #236 
* Improve some variable & function names
* Handle path, container and automagical mode change in one function.
* Use PF.FileUtils function to detect icon directory
* Modify existing handlers to call new function

Note: Additional lines are due to formatting and comments.

### DIFF
--- a/libcore/FileUtils.vala
+++ b/libcore/FileUtils.vala
@@ -552,6 +552,10 @@ namespace PF.FileUtils {
                 return false;
         }
     }
+
+    public bool is_icon_path (string path) {
+        return "/icons" in path || "/.icons" in path;
+    }
 }
 
 namespace Marlin {
--- a/libcore/gof-directory-async.vala
+++ b/libcore/gof-directory-async.vala
@@ -40,7 +40,7 @@ public class Async : Object {
     public int icon_size = 32;
 
     /* we're looking for particular path keywords like *\/icons* .icons ... */
-    public bool uri_contain_keypath_icons;
+    public bool uri_contain_keypath_icons = false;
 
     /* for auto-sizing Miller columns */
     public string longest_file_name = "";
@@ -444,8 +444,7 @@ public class Async : Object {
         set_confirm_trash ();
 
         if (can_load) {
-            uri_contain_keypath_icons = "/icons" in file.uri || "/.icons" in file.uri;
-
+            uri_contain_keypath_icons = PF.FileUtils.is_icon_path (file.uri);
             if (file_loaded_func == null && is_local) {
                 try {
                     monitor = location.monitor_directory (0);
@@ -459,7 +458,6 @@ public class Async : Object {
                     }
                 }
             }
-
 
             if (is_trash) {
                 connect_volume_monitor_signals ();
--- a/libcore/gof-directory-async.vala
+++ b/libcore/gof-directory-async.vala
@@ -40,7 +40,7 @@ public class Async : Object {
     public int icon_size = 32;
 
     /* we're looking for particular path keywords like *\/icons* .icons ... */
-    public bool uri_contain_keypath_icons = false;
+    public bool uri_contains_keypath_icons = false;
 
     /* for auto-sizing Miller columns */
     public string longest_file_name = "";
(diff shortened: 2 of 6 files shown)


libcore/FileUtils.vala
```diff
@@ -552,6 +552,10 @@ namespace PF.FileUtils {
                 return false;
         }
     }
+
+    public bool is_icon_path (string path) {
```
donadigo: This should be static.

libcore/gof-directory-async.vala
```diff
@@ -40,7 +40,7 @@ public class Async : Object {
     public int icon_size = 32;
 
     /* we're looking for particular path keywords like *\/icons* .icons ... */
-    public bool uri_contain_keypath_icons;
+    public bool uri_contain_keypath_icons = false;
```
donadigo: Can we change this variable name to `uri_contain`**`s`**`_keypath_icons` while we're here?

src/View/ViewContainer.vala
```diff
@@ -296,36 +302,55 @@ namespace Marlin.View {
...
+                           make_root ? Marlin.OpenFlag.NEW_ROOT : Marlin.OpenFlag.DEFAULT,
+                           allow_mode_change);
         }
 
-        private void on_slot_new_container_request (GLib.File loc, Marlin.OpenFlag flag = Marlin.OpenFlag.NEW_ROOT) {
+        private void open_location (GLib.File loc,
+                                    Marlin.OpenFlag flag = Marlin.OpenFlag.NEW_ROOT,
+                                    bool allow_mode_change = true) {
+
+            bool change_to_icon_mode = false;
```
donadigo: I'm not seeing this change anywhere else removed, can we get an explanation why, it's only present here in this diff?

src/View/ViewContainer.vala
```diff
@@ -296,36 +302,55 @@ namespace Marlin.View {
...
+
+            if (allow_mode_change &&
+                view_mode != Marlin.ViewMode.ICON &&
+                PF.FileUtils.is_icon_path (loc.get_uri ())) {
+
```
donadigo: Remove this newline.

libcore/FileUtils.vala
jeremypw: As per Slack discussion, methods outside classes are implicitly static.  If required for elementary code style, then I will submit a separate PR explicitly designating all the FileUtils methods to be static.

libcore/gof-directory-async.vala
jeremypw: Will do

src/View/ViewContainer.vala
jeremypw: I always insert a blank line after a multi-line if clause as it improves readability.

jeremypw: As per Slack discussion I will leave this blank line.

src/View/ViewContainer.vala
```diff
@@ -296,36 +302,57 @@ namespace Marlin.View {
             refresh_slot_info (slot.location);
         }
 
-        public void user_path_change_request (GLib.File loc, bool allow_mode_change = true, bool make_root = true) {
+        private void user_path_change_request (GLib.File loc, bool allow_mode_change = true, bool make_root = true) {
```
donadigo: It seems that this is no longer used in the code.

```
/src/View/ViewContainer.vala:305.9-305.45: warning: method `Marlin.View.ViewContainer.user_path_change_request' never used
        private void user_path_change_request (GLib.File loc, bool allow_mode_change = true, bool make_root = true) {
```

PR #223, opened 2017-11-15
Open executable scripts in text editor by default (#4)

### LINKED ISSUE #4: Automatic execution of scripts on click or double click cannot be changed [$15]
The executable scripts are always executed if clicked in Pantheon Files, even if the default application to be used is changed for them to a text editor under *Properties-Open with*.

I have noticed this behavior only for executable scripts (while the opposite happens for desktop files after February 2017 update:  http://elementaryos.stackexchange.com/q/10952/170, which by default are open in a text editor). 

This is also the cause why the system will not ask before running script from Pantheon Files (http://elementaryos.stackexchange.com/q/4256/170) even if a zenity script is added in a desktop file  used for that purpose as default application (http://elementaryos.stackexchange.com/a/10358/170) (which works with other file managers).

This behavior could be dangerous (given the reason why desktop files are not automatically executed after  February 2017 update). 

Automatic execution of scripts on click or double click should not be the default action or at least it should be possible to change it.

Launchpad Details: [#LP1669996](https://bugs.launchpad.net/bugs/1669996) cipricus - 2017-03-04 11:56:03 +0000

obyknovenius:
Fixes #4

### DIFF
--- a/libcore/gof-file.c
+++ b/libcore/gof-file.c
@@ -1372,8 +1372,9 @@ gof_file_is_executable (GOFFile *file)
             /* check if the content type is save to execute, we don't use
              * g_content_type_can_be_executable() for unix because it also returns
              * true for "text/plain" and we don't want that */
-            if (g_content_type_is_a (content_type, "application/x-executable"))
+            if (g_content_type_is_a (content_type, "application/x-executable")) {
                 can_execute = TRUE;
+            }
         }
     }
 
--- a/libcore/gof-file.c
+++ b/libcore/gof-file.c
@@ -1366,13 +1366,13 @@ gof_file_is_executable (GOFFile *file)
     {
         /* get the content type of the file */
         content_type = gof_file_get_ftype (file);
+        printf("Content type: %s\n", content_type);
         if (G_LIKELY (content_type != NULL))
         {
             /* check if the content type is save to execute, we don't use
              * g_content_type_can_be_executable() for unix because it also returns
              * true for "text/plain" and we don't want that */
-            if (g_content_type_is_a (content_type, "application/x-executable")
-                || g_content_type_is_a (content_type, "application/x-shellscript"))
+            if (g_content_type_is_a (content_type, "application/x-executable"))
                 can_execute = TRUE;
         }
     }
--- a/libcore/gof-file.c
+++ b/libcore/gof-file.c
@@ -1366,7 +1366,7 @@ gof_file_is_executable (GOFFile *file)
     {
         /* get the content type of the file */
         content_type = gof_file_get_ftype (file);
-        printf("Content type: %s\n", content_type);
+
         if (G_LIKELY (content_type != NULL))
         {
             /* check if the content type is save to execute, we don't use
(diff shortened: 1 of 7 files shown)


libcore/gof-file.c
```diff
@@ -1366,13 +1366,13 @@ gof_file_is_executable (GOFFile *file)
     {
         /* get the content type of the file */
         content_type = gof_file_get_ftype (file);
+        printf("Content type: %s\n", content_type);
```
jeremypw: Please use g_debug () (if necessary; avoid excessive use).  Also space before bracket.

obyknovenius: Forgot to remove it before push. Sorry

libcore/gof-file.c
```diff
@@ -1366,13 +1366,13 @@ gof_file_is_executable (GOFFile *file)
...
+
         if (G_LIKELY (content_type != NULL))
         {
             /* check if the content type is save to execute, we don't use
              * g_content_type_can_be_executable() for unix because it also returns
              * true for "text/plain" and we don't want that */
-            if (g_content_type_is_a (content_type, "application/x-executable")
-                || g_content_type_is_a (content_type, "application/x-shellscript"))
+            if (g_content_type_is_a (content_type, "application/x-executable"))
                 can_execute = TRUE;
```
jeremypw: By convention, we convert lines touched to elementaryos style even if rest of file does not comply.

Please use braces with single line if clause.

```
if () {

}
```

src/View/AbstractDirectoryView.vala
```diff
@@ -824,7 +824,12 @@ namespace FM {
                     if (file.is_root_network_folder ()) {
                         load_location (location);
                     } else if (file.is_executable ()) {
-                        file.execute (screen, null, null);
+                        var content_type = file.get_ftype ();
+
+                        if (g_content_type_is_a (content_type, "text/plain"))
+                            open_file (file, screen, default_app);
+                        else
+                            file.execute (screen, null, null);
```
jeremypw: Please use braces with single line if and else clauses.

```
if () {

} else {

}
```

src/View/AbstractDirectoryView.vala
```diff
@@ -824,7 +824,12 @@ namespace FM {
                     if (file.is_root_network_folder ()) {
                         load_location (location);
                     } else if (file.is_executable ()) {
-                        file.execute (screen, null, null);
+                        var content_type = file.get_ftype ();
+
+                        if (g_content_type_is_a (content_type, "text/plain"))
```
jeremypw: Please use Vala style:

```
GLib.ContentType.is_a (content_type, "text/plain");
```

libcore/gof-file.c
obyknovenius: I'm a newcomer. This is me second pull request. Thanks you for comments!

jeremypw:
Don't forget to claim your bounty! https://www.bountysource.com/issues/46338007-automatic-execution-of-scripts-on-click-or-double-click-cannot-be-changed

PR #230, opened 2017-11-16
Improve DnD handling.

### LINKED ISSUE #225: Incorrect drop file on second drag between two folders
To reproduce:

* Set up a source and destination folders in two tabs
* Place two files in the source.
* Drag one file from source to destination
* Drag the second file from source to destination
* An error dialog is produced indicating that the second drag is dropping a non-existent file (if moving) or that the file already exists (if copying).

### LINKED ISSUE #229: Possible to drag & drop within trash
Regression?  It should not be possible to drag and drop files around inside the trash.

jeremypw:
Fixes #229

* Use Gee.LinkList<> for drop_file_list
* Replace some related C functions with Vala in FileUtils.
* Reduce unnecessary checking in accepts_drop function.
* Improve trash directory detection.
* Share some AbstractDirectoryView and DndHandler code with BreadcrumbEntry.
* Use DndHandler.dnd_perform () in Sidebar.


jeremypw:
Has been split into separate branches

PR #231, opened 2017-11-16
Better trash detection

### LINKED ISSUE #229: Possible to drag & drop within trash

jeremypw:
Fixes #229 

* New Vala PF.FileUtils.function location_is_in_trash ()
* Use new function throughout.

Still detects file is in trash even if the scheme is not trash:// but the actual location of the system or user trash folder.

### DIFF
--- a/libcore/FileUtils.vala
+++ b/libcore/FileUtils.vala
@@ -555,8 +555,9 @@ namespace PF.FileUtils {
 
     public bool location_is_in_trash (GLib.File location) {
         var uri = location.get_uri ();
+        var scheme = Uri.parse_scheme (uri);
 
-        return (uri.has_prefix ("trash") ||
+        return (scheme != null && scheme.has_prefix ("trash") ||
                 uri.contains (GLib.Path.DIR_SEPARATOR_S + ".Trash-") ||
                 (uri.contains (GLib.Path.DIR_SEPARATOR_S + ".local") &&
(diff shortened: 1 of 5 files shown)


libcore/FileUtils.vala
```diff
@@ -552,6 +552,15 @@ namespace PF.FileUtils {
                 return false;
         }
     }
+
+    public bool location_is_in_trash (GLib.File location) {
```
donadigo: This should be static.

libcore/FileUtils.vala
```diff
@@ -552,6 +552,15 @@ namespace PF.FileUtils {
...
+        var uri = location.get_uri ();
+
+        return (uri.has_prefix ("trash") ||
```
donadigo: Shouldn't that test if the file has an URI scheme, like the C code did?

jeremypw: I guess so - this is just a bit quicker.

jeremypw:
I found that the uri passed to DnD can lose the trash:// scheme e.g. by opening folders in trash. If trashed files are not properly detected then the permitted DnD operations will not be correct.  I guess this could be regarded as a "work around" and we should try and make sure that only trash:// scheme uris are generated.

Is there a downside to keeping the extra checks, at least temporarily?

PR #235, opened 2017-11-18
Reduce slot window dependency

jeremypw:
* Remove some direct references to Marlin.View.Window, either by moving the funcrion to Window or emitting a signal.

* Get a reference to the app directly

### DIFF
--- a/src/View/Window.vala
+++ b/src/View/Window.vala
@@ -220,7 +220,7 @@ namespace Marlin.View {
 
             /* Toggle focus between sidebar and view using Tab key, unless location
              * bar in focus. */
-            key_press_event.connect ((event) => {
+            key_press_event.connect_after ((event) => {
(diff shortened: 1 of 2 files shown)


src/View/Window.vala
```diff
@@ -220,7 +220,7 @@ namespace Marlin.View {
             });
 
             undo_manager.request_menu_update.connect (undo_redo_menu_update_callback);
-            button_press_event.connect (on_button_press_event);
+            button_press_event.connect_after (on_button_press_event);
```
donadigo: Hm, any reason here, for handling this after the actual event?

jeremypw: The existing code handles the "find" function in AbstractDirectoryView after it has handled the other events, so now the "find" code is moved to Windows we connect this way so that the order of processing is unchanged.

PR #237, opened 2017-11-20
Handle dummy row null file entries better

jeremypw:
TO TEST:

Check for regressions and terminal warnings when displaying expanded empty directories in ListView and carrying out operations such as resorting, copy/paste etc.

### DIFF
--- a/libcore/PluginManager.vala
+++ b/libcore/PluginManager.vala
@@ -224,9 +224,17 @@ public class Marlin.PluginManager : Object {
             plugin.update_sidebar (widget);
     }
 
-    public void update_file_info (GOF.File file) {
-        foreach (var plugin in plugin_hash.values)
+    public void update_file_info (GOF.File? file) {
+        /* Plugin interface demands non-null file.  A null file can
+         * arise from a dummy row in list view */
+
+        if (file == null) {
+            return;
+        }
+
+        foreach (var plugin in plugin_hash.values) {
             plugin.update_file_info (file);
+        }
     }
 
     public Gee.List<string> get_available_plugins () {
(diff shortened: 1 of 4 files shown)


libcore/PluginManager.vala
```diff
@@ -224,9 +224,17 @@ public class Marlin.PluginManager : Object {
             plugin.update_sidebar (widget);
     }
 
-    public void update_file_info (GOF.File file) {
-        foreach (var plugin in plugin_hash.values)
+    public void update_file_info (GOF.File? file) {
```
donadigo: Shouldn't that method have the file not nullable by design? Every object that calls this should just verify if the file is actually a non-null object. Why is that change?

libcore/fm-list-model.c
```diff
@@ -533,15 +533,20 @@ fm_list_model_file_entry_compare_func (gconstpointer a,
...
         result = gof_file_compare_for_sort (file_entry1->file, file_entry2->file,
                                             model->details->sort_id,
                                             TRUE,
                                             (model->details->order == GTK_SORT_DESCENDING));
-    } else if (file_entry1->file == NULL) {
-        return -1;
     } else {
-        return 1;
+        /* Dummy rows representing expanded empty directories have null files */
+        if (file_entry1->file == NULL || file_entry1->file->location == NULL) {
```
donadigo: Shouldn't that check be just moved to the `else if` clause on diff line 541?

libcore/PluginManager.vala
jeremypw: It seemed more efficient to check once here rather than having to all calling objects ensure that no null files are passed.

PR #244, opened 2017-11-23
Vala accepts drop

jeremypw:
Transfer and refactor some code from gof_file.c to PF.FileUtils.vala.  Reduces line count.


libcore/FileUtils.vala
```diff
@@ -562,6 +562,144 @@ namespace PF.FileUtils {
...
+            actions = Gdk.DragAction.DEFAULT;
+        }
+
+        if (actions == Gdk.DragAction.DEFAULT) { // No point asking if no other valid actions
+            return Gdk.DragAction.DEFAULT;
+        } else if (location_is_in_trash (target_location)) { // cannot copy or link to trash
+            actions &= ~(Gdk.DragAction.COPY | Gdk.DragAction.LINK);
+        }
+
+        if ((suggested_action & actions) != 0) {
```
donadigo: Can we use the native Vala `in` syntax here?
```if (suggested_action in actions) { ... }```

File: libcore/FileUtils.vala [hunk 2]
```diff
@@ -562,6 +562,144 @@ namespace PF.FileUtils {
...
+
+                break;
+            }
+        }
+
+        /* Modify Gtk suggested COPY action to MOVE if source is trash or dest is in
+         * same filesystem and if MOVE is a valid action.  We assume that it is not possible
+         * to drop files both from remote and local filesystems simultaneously
+         */
+        if ((valid_actions & (Gdk.DragAction.COPY | Gdk.DragAction.MOVE)) != 0 &&
```
donadigo: Same here.

libcore/FileUtils.vala
jeremypw: Thanks - I did not know that was valid for enums!

PR #254, opened 2017-11-29
Avoid some slow list actions

jeremypw:
As the selection list can be very long we should not use operations that are O(n), like length (), where not necessary.


donadigo:
Has unresolved conflicts.

jeremypw:
Conflicts resolved.

PR #266, opened 2017-12-20
Always redraw on thumbnail finished signal (#214)

### LINKED ISSUE #214: [Bug] If there is no thumbnail cache, image thumbnails only show if you scroll or move the mouse to side bar.
Hi.  

If there is no thumbnail cache ( /home/user/.cache/thumbnails ), the image thumbnails only show if you scroll or move the mouse to the sidebar (shortcuts).

jeremypw:
Fixes #214


davidmhewitt:
It looks as though you may also be able to remove the `thumbnail_request` variable?

PR #268, opened 2017-12-20
Remove public cancel methods

jeremypw:
Simplify by combining close () and cancel () public methods - there is no need for the latter.


src/View/Slot.vala
```diff
@@ -393,21 +393,13 @@ namespace Marlin.View {
             }
         }
 
-        public override void cancel () {
+        public override void close () {
             cancel_timeouts ();
 
             if (directory != null) {
```
donadigo: This condition is now checked twice in the same `close` method: https://github.com/elementary/files/blob/be976c8b7fe63d354496427fc82d7a99aacc0edc/src/View/Slot.vala#L399

PR #286, opened 2018-01-11
Fix find regression

jeremypw:
In ListView (and ColumnView), direct typing into the view (without pressing Ctrl-F) has reverted to showing the native Gtk.TreeView search box.

This fixes that.


jeremypw:
This was caused by recent changes in Files - the order of key event handlers was changes to fix a different problem.  See https://github.com/elementary/files/commit/7f94e32a61235baab30c1c191091766023509b68

PR #297, opened 2018-01-19
Window action cleanup

jeremypw:
* Replace remaining binding entries with existing action accelerators
* Create a new action for "edit-path" and replace that binding entry.
* Remove some unused functions, actions and signals.
* Incorporate a small function into the constructor.

There is no intention to make key bindings variable via a config file so there seems no reason to maintain a Gtk.BindingSet.

It was decided not to reinstate BackSpace so the branch is now mis-named.

### DIFF
--- a/src/View/Window.vala
+++ b/src/View/Window.vala
@@ -39,8 +39,7 @@ namespace Marlin.View {
             {"info", action_info, "s"},
             {"view_mode", action_view_mode, "s", "'MILLER'"},
             {"show_hidden", null, null, "false", change_state_show_hidden},
-            {"show_remote_thumbnails", null, null, "false", change_state_show_remote_thumbnails},
-            {"show_sidebar", null ,  null, "false", change_state_show_sidebar}
+            {"show_remote_thumbnails", null, null, "false", change_state_show_remote_thumbnails}
         };
 
         public GLib.SimpleActionGroup win_actions;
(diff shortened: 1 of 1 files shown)


danrabbit:
I’m :-1: on adding backspace as a go back key. As far as I remember we removed this intentionally

jeremypw:
If it was intentional it was done badly.  The key binding was still present but was broken because the the virtual signal it linked to was missing.   There was also a keybinding for the "XF86Back" and "XF86Forward" keys still present but broken.  This is specifically intended for navigation (although only present on some keyboards) so should be supported?

danrabbit:
Yeah we should definitely support those keys and we’ve had reports recently about back and forward mouse buttons in other apps

jeremypw:
Note that the "BackSpace" still has a special action when in Column View.  It is used to close the current column and go back up the tree (maintaining the current root).  Pressing `<Alt>LeftArrow` in Column View has a different action - the previous history entry is opened as a new root  in the view.

A decision will have to be made whether to maintain this; if so whether to use a different shortcut, and if not,what the effect of `<Alt>LeftArrow` should be.

danrabbit:
It looks like this branch no longer respects the `show-sidebar` key in gsettings. Is that intentional?

jeremypw:
Yes, there was no way of affecting this through the UI anymore (the shortcut was withdrawn deliberately a fair while back).  There does not seem to be any good reason to allow the sidebar to be hidden _as a preference_ by manually editing the gsetting.  I have not (yet)  changed the schema as this PR was limited to Windows.vala.

danrabbit:
Okay cool. I'm down with killing that setting. We'll have to remember to remove the settings schema. Looks good though :+1:

PR #299, opened 2018-01-19
TextRenderer.vala: use render_background

danrabbit:
`get_background_color ()` is deprecated and we can replace a whole lot of stuff here with `render_background ()`

So far the only regression I've noticed is with the border radius, but I'm not sure this is really Files' fault. I'm actually not sure how to select this widget with CSS since Gtk.IconView is opaque to the inspector. I'm guessing what we should do here is add a style class, but I figured I'd leave it open to suggestions.


jeremypw:
TextRenderer is not a widget but you can use CSS like this (may not be the best way):

```
        private string CSS_DATA = """
        * {
            border-radius: %ipx;
        }

        """;
....
public override void render (Cairo.Context cr,
                                     Gtk.Widget widget,
                                     Gdk.Rectangle background_area,
                                     Gdk.Rectangle cell_area,
                                     Gtk.CellRendererState flags) {
...
            var style_context = widget.get_parent ().get_style_context ();
            style_context.save ();
            style_context.set_state (state);

            var css = new Gtk.CssProvider ();
            css.load_from_data (CSS_DATA.printf (border_radius));
            style_context.add_provider (css, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
...
```

This reproduces the variable border radius of the original.  If you forgo that, then some simplication is possible.

jeremypw:
I am not sure it is necessary to save, modify and restore the widget style context on each render - may be worth investigating whether the border radius can be set once on the parent widget.

jeremypw:
Needs decision regarding whether and how to reproduce curved corners.

jeremypw:
As per conversation on Slack, this can be merged pending resolution of styling before next release.

PR #311, opened 2018-01-26
Refactor opening multiple tabs or windows

### LINKED ISSUE #310: Open in Tab / Open in New Window broken for multiple selection
Opening a multiple selection in new tab or window from context menu should open multiple tabs/windows.
Only the first selected opens.

jeremypw:
Fixes #310 

* Multiple selections open correctly in multiple tabs and windows.
* Remove unnecessary Timeouts and Idles.
* Show dialog if too many files activated at once (Max number made a const)
* Correct formatting
* Reduce nested logic.

Checking for regressions.


jeremypw:
Will split

PR #319, opened 2018-02-08
Distinguish unfocused views by reducing opacity.

jeremypw:
In master, views do not indicate whether they are focused visually. Particularly in Column VIew, it is not easy to see which view is in focus when there is no cursor.

This branch dims views that are not in focus.  In order for this to appear in Column View the same way as in other views it was necessary to darken the background of the Miller slot.  Existing Gtk classes were used but UX team input required as to the most appropriate classes (if accepted in principle).


tintou:
Is there no way to tell a widget that it is currently not active ?
I was thinking that it's maybe possible to assign the whole view the state flag BACKDROP for it to view as if the window was in background https://valadoc.org/gtk+-3.0/Gtk.StateFlags.BACKDROP.html

danrabbit:
There is a way to determine focus in a list view and that's whether the selection is grey or blue. If this isn't clear enough for Files' column view, it's probably not clear enough for other kinds of list views, so I think this should be a stylesheet issue

jeremypw:
There is already a function for selecting the first item in a view if there is no other selection but this is only used when making a new root slot in Column View.  I seem to recall it was discontinued in other situations as it interfered with other desired behaviour like selecting the folder you have just come from when navigating up a directory tree using Alt-UpArrow.  I do now notice that when opening a new child slot in Column View, the first item has the cursor on it but it is not selected so that it appears grey, even though that slot is active.

So whether the cursor is grey or blue is not a infallible indication of whether a slot is active.  It can mean either that the item is focused but not selected or that it is selected but not focused.  So I think a more global change of appearance of a view when it is not active still has some merit.

I will however see if I can implement "select first when empty" again without losing other desired behaviours.

jeremypw:
Closing as there does not seem to be any support for this behaviour.

PR #321, opened 2018-02-10
Select first item of new view if no other selection

jeremypw:
See conversation on #319 .

This PR ensures that, if no other selection is made when opening a folder, the first item is selected.


danrabbit:
Hm I wonder if we should do this only on list view when you're opening a new column.

Also I think we have to be careful to focus, but not select. Otherwise (especially in column view) we can create a really large selection on accident

jeremypw:
Do you mean "... in Column View (aka Miller View) when you are opening a new column"?  I call call the three view modes Icon View, List View and Column View corresponding to Ctrl-1, Ctrl-2 and Ctrl-3 shortcuts.

I am not sure how a large selection can easily be made accidentally - could you explain?  Maybe this is why this was not implemented (or was removed) before.

The point about selecting the first item was so that something is blue when the view is active and grey when not.  This prevents confusion in Column as to which column is active under some circumstances.

Maybe the stylesheet could be modified so that the colours of both the cursor and selected items are distinguishable both when the view is focused and not focused?

danrabbit:
Er sorry, you're right. I meant to say column view :)

Yeah looking at it again, this might just be an unrelated visual bug. I was noticing that while navigating with the keyboard that everything in my current path had a selection badge when nothing should be "selected" in terms of making a copy/paste/move selection

I'm not sure what you mean by "the cursor" in this context. Are you referring to the item that has keyboard focus?

jeremypw:
Yes, the cursor shows the position of the item with keyboard focus.  This is not *necessarily* selected although the code does in fact mostly select the item under the cursor and deselect again when the cursor moves (unless the Shift key is held down).  You can see there is an issue if you hold the control key down and then move the cursor with the arrow keys.  The position of focus shows in that the icon is highlighted.  In Icon View the cursor is not visible.  In List and Column View the cursor appears to turns grey and does not move (although this could just be the selected but unfocused color).   I think the position of the cursor (focused item) should be easily visible and independent of the selected status.  

Needs design input to decide on optimum behaviour to aim at.

jeremypw:
I could not reproduce the "everything selected in current path" issue, but I noticed while navigating around by various methods that the first item is occasionally not selected.  Also, if Ctrl-A is pressed twice to unselect everything then the position of the cursor is hard to see in Icon View (although that is outside scope for this PR).

jeremypw:
Implemented focus only when moving cursor to first item for new view/columns.  This is similar to Nautilus's behaviour.

Also fixed issue where if a cursor key was pressed on a focussed but unselected item the cursor moved and selected.  Now, the focussed item is selected (again, similar to Nautilus).

Any issue of the visual distinguishability of focussed but unselected items from unfocussed, unselected items is not within the scope of this PR.

PR #325, opened 2018-02-10
"Search" icon in location bar for better discoverability (fixes #170, #396)

### LINKED ISSUE #170: No indication you can search files
There is nothing in Files' user interface that indicates you can actually search files. This can be a source of confusion for users, who will believe there is no such feature implemented.

Bounty: https://www.bountysource.com/issues/19860024-no-indication-you-can-search-files

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/50211479-no-indication-you-can-search-files?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jlnr:
This branch implements [the functionality](https://github.com/elementary/files/issues/170) proposed by @ypoz in #170:

* Replaces the reload icon in the location bar with a search icon. It isn't visible while searching, so there's never an awkward situation where the same icon appears on both sides of the bar.
* Moves the reload action to a menu button on the right end of the header bar.
* Also moves the two application-wide settings "Show Hidden Files" and "Show Remote Thumbnails" to the same menu and out of the directory-specific popup menu.

bildschirmfoto von 2018-02-11 00 26 01

What's not in this branch:

* The GActions corresponding to "Show Hidden Files" and "Show Remote Thumbnails" still live in the window, not the application, even though they affect the whole application.
* Before this branch, the "Show Remote Thumbnails" entry would only appear in the popup menu inside of network locations. Now it's *always* part of the gear menu, because it's an application-wide setting. It could be argued that this is still a violation of the elementary HIG because the app is showing UI that's not relevant to the user right now, and toggling this setting has no effect. I'm not sure how to best transport this information from AbstractDirectoryView to TopMenu, though.
* Updates to the localization files for the formerly invisible "Reload This Folder" action.

Anyway, I hope it's good enough to evaluate whether this UI makes sense or not. If not, I can reduce the scope to only replace the location bar icon, or we can discard this PR altogether (but then I'm not sure how #170 should be fixed). My biggest gripe is that the Reload action exists at all; shouldn't it be a no-op except maybe for network locations?


jeremypw:
@jlnr Thanks very much for starting to contribute to FIles! I hope you continue to do so - there is much that can be done ;-)   As this PR makes significant changes to the UX it will have to be referred to the UX team for an opinion.  I seem to remember the issue of an AppMenu for Files has arisen before and proven controversial (it is recommended to submit a blueprint for the proposed changes first for discussion).  I agree that "Show Hidden Files" has global effect and is rather out of place on the context menu (unless it be made contextual by only showing if there are hidden files and only applying to the current folder).  "Reload folder" and "Show Remote Thumbnails" are not truly global in that they affect only one folder or one type of folder so are not really candidates for an AppMenu.  Ideally "Show Remote Thumbnails" should be settable for particular connections as it is intended mainly for no overburdening slow connections.

I quite like the idea of the search icon in the pathbar instead of reload as it is mode often used and applies in all situation.  As you say, "reload" is mainly useful for remote folders. 

I do hope you continue to contribute, having braved Files code once.  Small PRs that fix one issue without significantly changing existing UI are likely to be quickly accepted. As are those that improve code quality or style.

jlnr:
@jeremypw Thank you for your feedback. I should have been more clear about this, but this branch was intended to add to the discussion rather than to be merged right away (if ever). It might have been wiser to only post my screenshot in #170 as a kind of blueprint, but I had already written the code as a learning experience. So I figured why not create a PR, it makes it easier to see the design in action.

I'm not attached to the branch and will happily axe it when decisions have been made. To be honest, I am not even happy with the app menu after using it for a bit.

If "Show Hidden Files" was a per-folder setting, and if "Show Remote Thumbnails" was a per-server setting (as you suggested), then it would only make sense that they stay in the right-click menu, and there would be no need for the app menu. Reload/refresh could also go in the right-click menu, and maybe only appear in remote folders. I might be biased towards simplicity because I rarely use these features, but that sounds like a much better and intuitive design to me.

Looking forward to future UX feedback on this issue, and I'll try to squash some minor bugs in the meantime :)

jlnr:
Since this came up again in #396, I've decided to clean up this PR by removing the app menu again. Now only two things have changed:

1. The reload icon in the navigation bar is now a search icon.
2. The reload functionality has moved into the background context menu as an entry "Reload", below "Show Hidden".

screenshot from 2018-03-27 23 54 12 2x

I hope this'll make it easier to discuss the search UX without being distracted by the app menu (which I wasn't a big fan of :/).

jeremypw:
After discussion with the UX team, it has been decided to fix this problem in a different way that does not involve an extra header bar icon (i.e. make the secondary entry icon a spinner while searching).  Therefore I am afraid it has been decided to reject this PR.  Thank you for your interest and input though.

jlnr:
@jeremypw No problem, but just to be sure - the current state of this branch did not introduce a new *header bar* icon (see most recent screenshot), it only replaced the refresh icon in the *search bar* with a search loupe, and that icon disappears as soon as the user starts searching, so no clash with a spinner.

If a clickable icon in the search bar is not great UX-wise, shouldn't the refresh icon also move out of there?

jlnr:
No problem. 👍 Just noticed the discussion in #419 and other tickets. Looking forward to further UX work on the navigation here!

jeremypw:
Yes, sorry - there was another PR with an added icon I was thinking of.
Anyway, the whole search UX is being reconsidered and Dan confirmed this PR
should not be merged now.
I vaguely recall some discussion on re the refresh icon placing but Its
dropped off Slack history now.

PR #333, opened 2018-02-13
Fix icon view keyboard select

### LINKED ISSUE #331: Focus bugs when renaming after keyboard selection
See #330  parts 2 & 3.

* Open a view containing at least two items
* Select an item by only pressing keyboard arrows
* Press F2 - nothing happens (expect start renaming)
* Press arrow to select the adjacent item.
* Press F2 - the previous item starts renaming instead of that currently selected

### LINKED ISSUE #332: Cursor move does not deselect properly in Icon View
* Open an Icon View containing several items
* Select multiple items 
* Use the keyboard arrows to move the cursor without holding Shift or Ctrl.
* Some of the items not under the cursor remain selected (expect behaviour to match List View and deselect all except under cursor)

jeremypw:
Fixes #331
Fixes #332 

Fixes the linked bugs by ensuring the list of selected files is updated after each key press.  (A previous commit had caused this to be delayed for performance reasons when large numbers of separate selection changes occur rapidly causing this regression).

In addition this adds another keyboard shortcut - Ctrl-Alt-plus (or Ctrl-Alt - equal) - to select arbitrary files in addition to the existing selection - this matches the existing shortcut of Ctrl-Alt-minus to deselect.  This permits moving the cursor using the arrow keys with the Ctrl key held down and selecting/deselecting single files without releasing the Ctrl key.   At present there is no keyboard only method to do this.

Many of the extra lines are format changes as per elementary code style.


jlnr:
I've tried this branch earlier today and it seems to work fine, thanks! :tada:

Issue #330, opened 2018-02-13
Focus bugs when selecting with the keyboard in Icon View


jlnr:
1. Enter a folder with several icons (four in my case) in Icon View
2. Select all of them using the mouse (I do so from right to left)
3. Press the left arrow key. Now the right-most icon is selected (bug 1?)
4. Press F2 to rename. Nothing happens (bug 2)
5. Press left to select the next icon.
6. Press F2 to rename the right-most (wrong) icon (bug 3)

Possibly related to #87 which was also about icon order and renaming.

Let me know if I should upload a GIF of this.

jlnr:
Sometimes step 3 results in three icons being selected (shrinking the selection), instead of one.

jlnr:
Reproduced both on stable Loki (0.4.1) and master.

jeremypw:
Thanks for pointing this out.  Regarding Bug 1 I guess part of the problem is that in Icon View, the cursor position is not distinguishable when there is multiple selection (unlike List View) so it is not obvious what will happen when an arrow key is pressed.  But you are right that what does happen is not always correct anyway.  I guess it should be the same as in List View, that is the cursor should move from its current position and other selected items be deselected?  This bug only seems to affect Icon View

The rename problems are related to keyboard selection and are confirmed. They affect all views so is likely to be a separate issue - I'll open another report.

jeremypw:
Actually the cursor position in Icon View multiple selections can be discerned - the color is a slightly brighter blue, but it is not obvious.

jlnr:
#333 fixes most of this, except part 3:

> Press the left arrow key. Now the right-most icon is selected (bug 1?)

This is still "buggy" from my initial perspective. If I open a fresh window containing three icons, select the one in the middle *using the mouse*, then press the left arrow key, I would expect the icon on the left to be selected. What happens is that Files selects the one on the right.

However, I just realized that this works as designed, so I'll spell it out: Selecting items with the mouse does not change the position of the *keyboard cursor*. Pressing the left arrow when there is no cursor moves the cursor to the last item on the list, and then selects this item. The F2 shortcut to rename icons does not operate on the keyboard *cursor* like the arrow keys, but instead renames the *selected* item.

Thinking about it, I think my expectation was that selecting icons using the mouse *also* moved the keyboard cursor to the (end of the) selection, in the same way that selecting text moves the keyboard cursor as well, but maybe I was misled by the styling, for which I've opened (repurposed) #332.

[...]

PR #340, opened 2018-02-18
Ensure cursor follows helper selection without losing existing selection

jeremypw:
This makes IconView and TreeView behave the same way when the cursor is moved with the keyboard after making a selection using the helpers and mouse.

In master, the treeview behaves differently to icon view - the former does not move the cursor to the last selected item whereas iconview does (this is due to differences in Gtk implementation).

In this PR, the cursor moves to last selected without losing the existing selection in all views (in treeview the selection is remembered and reestablished after moving the cursor).

This fixes part of #330

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -3247,7 +3247,7 @@ namespace FM {
                                     unselect_path (path);
                                 } else {
                                     should_deselect = false;
-                                    select_path (path);  /* Cursor follows */
+                                    select_path (path, true);  /* Cursor follow and selection preserved */
                                 }
                             }
 


jlnr:
Ah, I see now. This works and fixes my issue when using the helper badge, but the keyboard cursor is not updated when using a selection rectangle (which I often do even for single files). Is the difference between both interactions intentional?

jeremypw:
I found dealing with rubberband selections difficult so I left it for another day.  In IconView it not so clear which will be the last item selected when doing a rectangular selection.

jeremypw:
Whereas helper selection is handled by Files code, rubberband selection is handled by the Gtk Widget which complicates matters.

jlnr:
Makes sense, thanks.

PR #343, opened 2018-02-20
Fix cannot turn off external disk. (#185)

### LINKED ISSUE #185: Files can't turn off external hard disk
https://bugs.launchpad.net/pantheon-files/+bug/1611022

jeremypw:
Fixes #185 

Sidebar unmount/eject code rewritten so that if a drive can be stopped it is stopped when the last volume on it is unmounted.

TO TEST:

* Attach an external mechanical hard disk preferably with more than one mountable partition and with an LED showing whether it is running.
* Mount two volumes on it.
* Unmount one volume - the LED stays on.
* Unmount the last volume - the LED turn off.

* Check mounting and unmounting of other devices unaffected.


jeremypw:
Regression found following recent merges - fixing.

Issue #326, opened 2018-02-11
Re-ordering a folder in one window does not update other open windows


jlnr:
This is an edge case where the refresh button/shortcut has a visible effect:

1. Open first window showing the Home directory
2. Open second window showing the Home directory
3. Reverse the order of icons in one window via right-click->Sort By->Reverse
4. The other window retains its current order until the view is refreshed

Expected behavior: The refresh should happen automatically, just like "Show Hidden Files" automatically affects all windows.

jlnr:
Quoting you/@jeremypw in #358:

> I am uncertain what the desired behaviour is though.  There may be a use case for viewing two different orders simultaneously.

Good point, that sounds like a plausible use case. I personally expected the sort order to apply to all windows immediately because it's conceptually a property of the directory, not of the window, and I trust Files to refresh directory information reliably. I also found it unintuitive that pressing refresh in one window also updates the sort order in other windows.

Not sure if this is really medium priority, I doubt many people will run into this :)

jeremypw:
Reduced priority and flagged for UX input. I think conceptually the sort order is a property of the view (model) since the files are not re-ordered on disk.   However, Files placed some meta-data on disk which is immediately updated whenever a view showing that directory is re-ordered.  Perhaps this metadata should only be written when the last view of the directory is closed.

jeremypw:
In the early Gtk4 version, the sort order of the same folder in different windows persists during a refresh (unlike current version).  I think I'll close this as it will not be fixed in the Gtk3 version.  A new issue can be raised if the Gtk4 version is problematic.

PR #358, opened 2018-03-11
Thumbnail duplicate tab

### LINKED ISSUE #353: Opening a folder on two separate tabs prevents updates in the view
Steps to reproduce:

- Open a folder with photos, or elements that have a thumbnail (ie: ~/Photos)
- Open the same folder on a second tab
- Close the app
- Re-open the app, one tab will have all the thumbnails as normal, the second one will not be updated. In fact, adding/removing a file on the tab with the thumbnails will not be reflected on the second tab

It can also be reproduced like so:

- Open the app, and go to a folder
- Open a second tab, and go to the same folder
- Try to create a file or a folder. You won't see it appear, but clicking Refresh updates the view

Note: No warnings or errors shown when running from the terminal after doing said actions

jeremypw:
Fixes #353 
Steps are taken to ensure that if two views are opened of the same directory then they receive the same Async object.  Since initialisation of the async directories was made asynchronous and they were not added to the directory-cache until initialisation completed (since the location could change if created with a non-folder), a cache entry is made while waiting for initialisation to complete using an immutable creation key.  Two cache entries are made where the location changes during initialisation.

To avoid manual ref() and unref() of async objects, the cache now contains unowned objects.  The file and location members are made publicly read-only properties.


jlnr:
Will check later, but maybe this also fixes #326

jeremypw:
Unfortunately not.  Sorting is not handled by the async directory object.
and each view has its own model  To get synchronization a signal would have
to be emitted by the directory view when the sort order changes.  I am
uncertain what the desired behaviour is though.  There may be a use case
for viewing two different orders simultaneously.

jlnr:
Whoops, sorry for deleting my comment that you're now replying to -- I was about to report back that I've just looked at #326 on this branch and you are of course correct, it's unrelated.

PR #359, opened 2018-03-11
ViewContainer destruct

jeremypw:
This makes two changes that ensure that the ViewContainer object associated with each notebook tab is destroyed (along with its child slot(s), browser and overlaybar) when that tab is closed.

* Do not reference the ViewContainer in a nested closure.  Refactor the function to use the current tab page.
* Close the view later in the tab close process, and call destroy explicitly to break circular reference to overlaybar.

Some repeated code was made into a function.

To test:

Run with --debug flag and monitor for "ViewContainer destruct" message.

### DIFF
--- a/src/View/Browser.vala
+++ b/src/View/Browser.vala
@@ -35,6 +35,9 @@ namespace Marlin.View {
             forward_stack = new Stack<string> ();
         }
 
+        ~Browser () {
+            debug ("Browser destruct");
+        }
         /**
          * Use this method to track an uri location in
          * the back/forward stacks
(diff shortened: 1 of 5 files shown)


jeremypw:
I'll check whether ViewContainer is destructing properly in master now. I think there was more than one reason it wasn't.

jeremypw:
I have updated the PR to take advantage of the new Granite.Widgets.OverlayBar api.  Still needed to remove the nested closure reference so this PR is still valid.

jeremypw:
However, this PR will not pass Travis CI based on Loki, which does not use the latest version of Granite so this is blocked until that is resolved.

jeremypw:
Now passing TravisCI, based on juno-unstable.

Issue #363, opened 2018-03-14
Use admin:/// for administrator operations


donadigo:
It turns out that you can do admin operations on files or any other process if you use the `admin:///` URI scheme instead of the `file:///` one. In this case we should make all the context menu entries in root protected directories enabled and use this URI to allow the user to perform wanted operations.

mdh34:
at the moment, putting `admin://` into the address bar of files and authenticating allows most standard procedures like renaming, deleting etc. A few things i noticed were that permissions aren't available, and copying files creates a file owned by the current user rather than root.

danrabbit:
Closing as a duplicate of #165

PR #364, opened 2018-03-14
Remove 'new window as administrator' option

### LINKED ISSUE #224: Remove Open as Admin
This desktop action should go away to impeach users without knowledge to fuck their system

mdh34:
Now GVFS supports the `admin://` option opening a new window as an administrator shouldn't be needed
fixes #224


danrabbit:
I think we probably need to solve #363 in a smooth way before we should remove this option. Otherwise, from a user's perspective, it will look like we lost a feature

jeremypw:
Blocked pending implementation of alternative.

mdh34:
Closing this as it doesn't really address the issue

PR #373, opened 2018-03-19
Port ListModel to Vala

tintou:


tintou:
@jeremypw This is a huge port, and I don't really get why it's crashing, but the `FileEntry` structure needs to be changed because Vala don't like this…

jeremypw:
Is it not possible to use a (customised) GtkTreeStore for this?

There are all sorts of referencing counting/memory management problems manipulating GLists of TreeIters etc.  At least one crash is due to this.

tintou:
@jeremypw I update this branch to compile and work with master.
This is just a straight port, I propose to redesign it at a later stage.

jeremypw:
@tintou:  This works pretty well! The main issue I have found so far is that in List View there are no expanders and the keyboard shortcut for expanding folders does not work either.

Were you envisaging merging this or the gof-file port first?

tintou:
Closing until it get properly fixed

PR #376, opened 2018-03-19
Port IconInfo to Vala

tintou:


libcore/IconInfo.vala
```diff
@@ -0,0 +1,339 @@
...
+    private struct LoadableIconKey {
+        GLib.Icon icon;
+        int size;
+    }
+
+    private struct ThemedIconKey {
+        string filename;
+        int size;
+    }
+
```
jeremypw: Could make the keys simple classes to avoid having to box them.

libcore/IconInfo.vala
```diff
@@ -0,0 +1,339 @@
...
+    private static uint loadable_icon_key_hash (LoadableIconKey? key) {
+        return (key.icon.hash () ^ key.size);
+    }
+
+    private static bool loadable_icon_key_equal (LoadableIconKey? a, LoadableIconKey? b) {
+        return (a.size == b.size && a.icon.equal (b.icon));
+    }
+
+    private static uint themed_icon_key_hash (ThemedIconKey? key) {
+        return (key.filename.hash () ^ key.size);
```
jeremypw: Appears to dereference a potential null

libcore/IconInfo.vala
```diff
@@ -0,0 +1,339 @@
...
+        if (filename != null) {
+            filename = GLib.Path.get_basename (filename);
+            var last = filename.last_index_of_char ('.');
+            icon_name = filename.substring (0, last);
+        }
+    }
+
+    construct {
+        last_use_time = GLib.get_monotonic_time ();
+        schedule_reap_cache ();
```
jeremypw: Not sure why?

libcore/IconInfo.vala
```diff
@@ -0,0 +1,339 @@
...
+                    critical (e.message);
+                }
+            }
+
+            if (pixbuf != null) {
+                icon_info = new IconInfo.for_pixbuf (pixbuf);
+                loadable_icon_cache.insert (loadable_key, icon_info);
+            }
+
+            return null;
```
jeremypw: null here differs from .c file, which returns icon_info.

libcore/IconInfo.vala
jeremypw: On reflection, this replaces the reaping that in the c version was triggered by pixbuf_toggle_notify (i.e when a pixbuf became unused for a certain time).

jeremypw:
The new iconinfo test would have picked up the problem with the null return from iconinfo.lookup ().

jeremypw:
Not sure why Travis CI is failing - running ctest locally passes OK.  Will investigate

tintou:
@jeremypw Alright, seems like this is good this time, looks like it's ready to merge IMO

jeremypw:
OK lets go for it.

PR #377, opened 2018-03-19
Bump Gtk dependency to 3.22 and remove EELGtkExtension

tintou:


CMakeLists.txt
```diff
@@ -40,7 +40,7 @@ set (COMMON_DEPS
     gio-2.0
     gio-unix-2.0
     pango>=1.1.2
-    gtk+-3.0>=3.14
+    gtk+-3.0>=3.22
```
jeremypw: This breaks Travis CI while that install version 3.18.

src/View/Sidebar.vala
```diff
@@ -1496,10 +1496,7 @@ namespace Marlin.Places {
                 check_popup_sensitivity ();
             }
 
-            Eel.pop_up_context_menu (popupmenu,
-                                     Marlin.DEFAULT_POPUP_MENU_DISPLACEMENT,
-                                     Marlin.DEFAULT_POPUP_MENU_DISPLACEMENT,
-                                     event);
+            popupmenu.popup_at_pointer (event);
```
jeremypw: This loses the 2 px displacement of the menu but I cannot see any adverse implications.

danrabbit:
unblocking because CI passes

PR #393, opened 2018-03-25
FileConflictDialog: port to Vala

tintou:


danrabbit:
I proposed a branch to merge into this one that fixes some issues with style classes, spacing, and some other misc stuff https://github.com/elementary/files/pull/394

jeremypw:
There is no direct way to close/cancel this dialog - is that deliberate?  To cancel the user has to select "apply to all" and then press skip.

danrabbit:
@jeremypw nice catch. Fixed!

PR #400, opened 2018-03-28
Add check for stale thumbnail

### LINKED ISSUE #398: Renaming image file does not invalidate thumbnail for new path.
TO REPRODUCE:

* Create a folder containing two different image files which have been thumbnailed.
* Move one file out of the folder
* Rename the other file to the same name as the first file.
* The thumbnail changes to that of the first file (and therefore does not match the contents of the file)

airlanggacahya:
Fix  #398

Add function to check thumbnail change time and compare it with file change time. If file is newer than thumbnails, unlink thumbnail to force recreation of thumbnail file.

Several documents such as https://specifications.freedesktop.org/thumbnail-spec/thumbnail-spec-latest.html recommend to use modified time to check for stale thumbnail. However renaming doesn't update modified time at all, so this method is not viable in our case.

According to https://stackoverflow.com/questions/3385203/regarding-access-time-unix, file metadata will update change time value. Therefore file modification can be tracked using changed time too.

Possible effect on directory with many file, since this will add at least 1 file stats for each item displayed. But I think it will negligible.

### DIFF
--- a/libcore/gof-file.c
+++ b/libcore/gof-file.c
@@ -854,7 +854,42 @@ gof_file_query_thumbnail_update (GOFFile *file)
         g_free (md5_hash);
     }
 
+    gof_file_check_stale_thumbnail(file);
+
     gof_file_update_icon_internal (file, file->pix_size);
+}
+
+void gof_file_check_stale_thumbnail (GOFFile *file)
+{
+    GFile       *thumbs;
+    GFileInfo   *info;
+    GError      *err = NULL;
+    guint64     file_changed;
+    guint64     thumb_changed;
+    const gchar *thumb_path;
+
+    thumb_path = gof_file_get_thumbnail_path (file);
+    thumbs = g_file_new_for_path (thumb_path);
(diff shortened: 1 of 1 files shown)


jeremypw:
Thanks for the fix. I was wondering whether it would be more efficient for the thumbnail to invalidated when Files renames/deletes/moves the file - but I guess that would not work if the changes are made externally.  I'll see if this affects performance noticeably.

jeremypw:
This does not materially affect the display time for large folder of images (the rate limiting step is construction of the model - thumbnails are only updated on an as needed basis).

jeremypw:
This does not fix the related issue #402 . Would it be possible to incorporate a fix for that?

airlanggacahya:
Are you sure? Because moved file should have newer change time than the
thumbnails.

But if that is the case, let me check it up later.

jeremypw:
I agree - I was surporised, but I put a terminal message into the check stale thumbnail function and it was was being called but the thumb was still wrong.

danrabbit:
Closing because this branch has been superseded by #492

PR #407, opened 2018-04-09
Port ProgressInfoManager to Vala

tintou:
fixed some memory leaks on the go


tintou:
I prefer doing a 1:1 port when possible, it will be then easier to change that in Vala

tintou:
oh well, there is already only vala files using it, I'll do it with gee then

PR #408, opened 2018-04-09
Remove Exec and use GLib features

tintou:
Change some functions in GOFFile to use Gio instead of custom ones


tintou:
@jeremypw You're right, we don't need that, I just made the changes

jeremypw:
I suspect "drop_target" becomes null.  This may be exposing an existing bug as in master the script runs and there is no error message.

PR #413, opened 2018-04-12
Replace fcts by UserUtils

tintou:


libcore/UserUtils.vala
```diff
@@ -0,0 +1,138 @@
...
+
+        return true;
+    }
+
+    public string get_real_user_home () {
+        unowned string? real_uid_s = GLib.Environ.get_variable (GLib.Environ.get (), "PKEXEC_UID");
+        Posix.uid_t? scanned_id = null;
+
+        if (real_uid_s != null && get_id_from_digit_string (real_uid_s, out scanned_id)) {
+            return get_user_home_from_user_uid (scanned_id);
```
jeremypw: scanned_id may be null.

libcore/UserUtils.vala
```diff
@@ -0,0 +1,138 @@
...
+            uid = null;
+            return false;
+        }
+
+        uid = password_info.pw_uid;
+        return true;
+    }
+
+    public unowned string? get_user_home_from_user_uid (Posix.uid_t uid) {
+        unowned Posix.Passwd? password_info = Posix.getpwuid (uid);
```
jeremypw: may be passed a null uid.

PR #414, opened 2018-04-14
Add HiDPI support, restore color tagging support

tintou:
:tada: :tada: :tada: :tada: :tada:


danrabbit:
With this branch, icon renders are half sized:

screenshot from 2018-04-15 11 02 53 2x

jeremypw:
I can confirm that the icons are rendered at the same size regardless of the window scaling factor chosen (with interactive debugging), so relative to the text size the become smaller.

jeremypw:
The above comment only applies to thumbnail icons, not themed icons.  Seeking fix.

PR #435, opened 2018-04-23
Do not stop search when unfocusing the window

donadigo:
This allows the user to freely unfocus the window and let Files search for a query instead of just stopping the search operation.

### DIFF
--- a/libcore/ProgressInfo.vala
+++ b/libcore/ProgressInfo.vala
@@ -46,7 +46,7 @@ public class PF.Progress.Info : GLib.Object {
     private bool progress_at_idle;
 
     public Info () {
-    
+
     }
 
     construct {
@@ -120,7 +120,7 @@ public class PF.Progress.Info : GLib.Object {
         if (activity_mode) {
             return -1;
         } else {
-            return progress;
+            return total;
         }
     }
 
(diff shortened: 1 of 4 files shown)


donadigo:
After some more testing, this is ready to review.

jeremypw:
A side effect of this change is that if another window is focussed during a search and that window overlays the main Files window then the search window remains on top, if already showing, or pops up over the other app when search completes with matches.  Clicking on a match under these conditions navigates Files as expected but does not focus the Files window, which remains hidden.

Is this the desired behaviour?   I can see that some notification of search completion is desirable, but if you click on a match I think Files should be focussed.

jeremypw:
@donadigo :  Yes, please fix in this branch unless it requires large changes (I don't imagine it will).

donadigo:
@jeremypw done.

jeremypw:
Reverted by #481 due to unforeseen side-effects including possible crash.

Was the main reason for this to keep a lengthy search going while doing something else?  Maybe a different solution can be found.

donadigo:
@jeremypw yes, that was the main reason I did this. Perhaps it's time to rethink search entirely?

jeremypw:
Yes.  As mentioned in Slack maybe splitting it into separate widgets rather than having one multi-mode widget would be simpler.

PR #461, opened 2018-05-07
Fix filename appearing in pathbar in some circumstances

### LINKED ISSUE #239: Open in folder (firefox) opens Files successfully but with the filename in the pathbar
I don't know whether this is problem with Files or Firefox, but when I hit show in folder in downloads, this happens:
image

jeremypw:
Fixes #239 

Also fixes same issue when opening a file (not a folder) from the command line without any option.
Updates slot info display after loading in case location changed during slot initialization.

TO TEST
At commandline type `io.elementary.files ~/Test_Folder/test_file.txt`  where Test_Folder and test_file.txt exist.

In master Files opens showing Test_Folder with test_file.txt selected, but the pathbar has an extra breadcrumb to `test_file.txt` and the tab name is `test_file.txt`.

In PR there is no extra breadcrumb and the tab name is `Test_Folder`

### DIFF
--- a/src/View/ViewContainer.vala
+++ b/src/View/ViewContainer.vala
@@ -446,6 +446,7 @@ namespace Marlin.View {
                 browser.record_uri (null);
             }
 
+            refresh_slot_info (slot.location);
             loading (false); /* Will cause topmenu to update */
             overlay_statusbar.update_hovered (null); /* Prevent empty statusbar showing */
         }


tintou:
Works when the folder exists but when it doesn't exist it still shows the complete filename in the pathbar

jeremypw:
When the file does not exist it not possible to tell whether or not it is a folder so Files behaves as if it is a non-existing folder, - I do not think this is an issue.  I suppose we could assume that if the name has an extension it is not meant to be folder but whether it is worth the extra code is questionable.

PR #476, opened 2018-05-10
Fix crash when searching twice.

### LINKED ISSUE #466: Crash when searching twice
Open files
Search for something (that least to at least one directory result)
Go into a search result directory
Type anything to crash Files

jeremypw:
Fixes #466

This issue was due to an unnoticed side-effect of 6114ea723318d486f7be77d27c7b7e69ab5ea177

Make sure breadcrumbentry focus out handler loses race with window.has_top_level_focus () by increasing length of delay.

This is just a quick fix.  The pathbar widgets need refactoring to make less fragile.

Some trailing spaces have been automatically removed.


jeremypw:
Unable to find a solution that fixes this and maintains search mode while other apps focussed.

PR #477, opened 2018-05-10
Fix freeze when clicking on folder that is already open in Miller View

### LINKED ISSUE #467: clicking directory twice in column view locks out the pathbar
If I click a directory to open it and then click it again, the pathbar goes insensitive and I get a spinner on the tab. This doesn't happen with a fast double-click which seems to set the directory as the top level in the column

screenshot from 2018-05-08 11 25 54 2x

jeremypw:
Fixes #467

Reuse opened slot in Miller.  Fixes signal race whereby the ViewContainer receives the "done-loading" signal before the "path changed" signal when clicking on a folder that is already opened resulting in the view staying in the "working" and "frozen" state.

### DIFF
--- a/src/View/Miller.vala
+++ b/src/View/Miller.vala
@@ -106,7 +106,6 @@ namespace Marlin.View {
             }
 
             new_slot.active (scroll, animate); /* This will set the new slot to be current_slot. Must do this before loading */
-
         }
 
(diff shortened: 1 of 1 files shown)


jeremypw:
Closing - there is already a (shorter) PR that fixes this!

PR #493, opened 2018-05-18
Simplify trash/delete. (#315)

### LINKED ISSUE #315: Remapped ALT + BackSpace to behave like DEL key not working
I don't have a <kbd>DEL</kbd> key on my laptop so I remapped <kbd>ALT</kbd>+<kbd>BackSpace</kbd> to do this. It is working correctly in Scratch or any other text editors like ViM but doesn't delete a file/folder in Pantheon File Manager.

Switching map to <kbd>Shift</kbd>+<kbd>BackSpace</kbd> works correctly (deletes text or file) on both text editor and Pantheon Files.

I installed Nautilus File Manager and <kbd>ALT</kbd>+<kbd>BackSpace</kbd> works as intended.

So the issue is only with <kbd>ALT</kbd>+<kbd>BackSpace</kbd> combination on Pantheon Files.

Also posted this on [unix stackexchange](https://unix.stackexchange.com/questions/421124/how-to-remap-alt-backspace-to-work-like-del-key)

jeremypw:
Fixes #315 

* Allow modifiers on delete key for trashing to allow more remapping possibilities.
* Simplify Del key handling code.
* Lose duplicated code for deleting files.
* Do not confirm individual file deletion from trash, even if confirm empty trash.
* Clarify "confirm trash" setting description.


danrabbit:
Can you resolve conflicts here? :)

PR #500, opened 2018-05-19
Do not activate on long button press, but select. (#227)

### LINKED ISSUE #227: Press and hold primary mouse button on a file or folder to select it
Problem: A lot of users including me face problems with the single-click to open paradigm. Especially the pain of selecting files or folder without opening it. elementary OS provides two ways to do that by holding CTRL on keyboard and clicking to select files, but it requires an extra input device for a work that could have been done using only the mouse. Plus a lot of people might not be aware of it or find it uneasy to handle even if they discover the way. And the other way is clicking the tick on the corner of the file. But many people didn't expect that feature. Even I came to know if it recently. Plus it is not handy when it comes to clicking on the small tick icon with some touchpads as opposed to clicking the bigger file icon.

Solution: There had always been a way to select files and folder on touch input devices, running iOS (iPhone) or Android. One has to simply touch on the icon and hold it either to select it or open context menu. In elementary OS, we do have secondary-click to open context menu. So, we could just click and hold on a file or folder using the primary mouse button to select it. On selecting a file, a smaller cotext menu would appear showing common controls like "cut" "copy", etc.. A horizontal context menu with icons only should get in to way less. On clicking else where the context menu would disappear, but the file would remain selected. In this moment, clicking on other files would select or deselect them. The context menu would reappear after  a certain amount of delay when no more clicks are done. This would make it easy for people who are familiar with usual touch interfaces and would also make it easier for new people to get used to it. It would work the same way with mouse, touch-pads, track-balls or any other pointing device. This would also open up opportunities for touch screen devices like tablets or touch-enabled laptops

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/51617718-press-and-hold-primary-mouse-button-on-a-file-or-folder-to-select-it?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #227


danrabbit:
I think I'm on board with this as an idea, but there's currently no way to know if the select has succeeded. It seems like files are always shown as selected after clicking whether or not releasing the mouse button will launch or select

jeremypw:
Items are selected on button down at the moment, but activated on button release.  I'll see if I can fix.

jeremypw:
Last commit does some refactoring to try and achieve desired behaviour.  Particularly impacts Column View. It was decided to considerably simplify matters by using a middle-click instead of a double click to open a folder in the root position.   Further cleanup and testing required.

jeremypw:
Current version stops the immediate select but the select status does not show until the button is release so it is still unclear what will happen.  This because there is no timeout running during button down.  The length of press is only determined upon release.  Will try to fix.

danrabbit:
Hey right on, this sure seems like it works as intended to me. Nice work :) Just gotta resolve that conflict ;p

jeremypw:
I think some simplification to the button release handler is possible now there is a timeout.

jeremypw:
Because of various related merges before this, I think it safer to start again than try to fix conflicts.

PR #506, opened 2018-05-20
Use 'folder-open' icon when folder has been opened or expanded. (#474 #148 #153)

### LINKED ISSUE #148: Crash when refreshing multiple windows rapidly

### LINKED ISSUE #153: Objects not destroyed when tab closed
If a tab containing the only open view of a folder is close, the corresponding GOFDirectoryAsync, Slot and ViewContainer should be, but are not destroyed (the DirectoryAsync and Slot are destroyed as expected on view mode change)

### LINKED ISSUE #474: Use "folder-open" for open folders in list and column view
For folders that are expanded/open, we should use the icon `folder-open`:

folder-drag-accept

jeremypw:
Fixes #474 Fixes #148 Fixes #153 

Note that this PR causes the "folder-open" icon to be used if the folder is open *any* how.  i.e if it is open in another tab or window or another Miller column or as an expanded row in list view.   It relies on the GOF.Directory.Async object being removed from the cache when it is not being displayed anywhere.

In order for this to work properly it was necessary to incorporate a fix for ViewContainer destruction and this in turn required fixing a regression in dragging a tab to a new window (which was only working because the ViewContainer was not being destroyed when it should).  Reluctantly a timeout had to be used to overcome a crash being caused by some undiagnosed race condition when dragging a duplicate tab out of the window.

If the "folder-open" state is to be confined to individual views, then it will be necessary to implement a more complicated solution involving adding another column to the fm-list-model which does not derive its data from a GOF file (GOF file objects are shared between views).


jeremypw:
Regression not fully resolved.

jeremypw:
Regression fixed

danrabbit:
Looks good UX wise :)

jeremypw:
The need for a long timeout no longer needed (race condition was found and fixed). Reverted to original Idle loop, which is still needed.

PR #513, opened 2018-05-25
Add "Keep Newest" option in case of file conflict (#405).

### LINKED ISSUE #405: Add a "keep newest" option to FileConflictDialog
Currently in FileConflictDialog you can either "Skip", "Replace", or "Rename", where "Rename" requires manual entry for each file. When copying near-identical folders over each other, I frequently find myself preferring a non-existent "keep newest" option that I can quickly apply to all future conflicts.

Related to #164

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/56741182-add-a-keep-newest-option-to-fileconflictdialog?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #405 

It is not clear what should happen if the files are exactly the same modification date (very unlikely if the files are different).  In this PR, then the operation is skipped but this would be easy to change.  Ideally Files should check whether the file contents are identical - there is an open issue regarding this. It did not seem worth adding another dialog.


lewisgoddard:
I'm not sure I'm qualified to just review and merge this myself, so I'm waiting on a second opinion.

jeremypw:
Thanks for the review @lewisgoddard

PR #516, opened 2018-05-28
Use actionbar in trash view and reduce minimum width (#147).

### LINKED ISSUE #147: Sidebar (left panel) shrinks on selecting 'Rubbish Bin' with small window
Hi, I just noticed this small issue when I accidentally left-clicked on the Rubbish Bin in the left column instead of my Workspace folder shortcut (I usually right click and delete from there).

This only happens when there is something in the bin to delete, and it appears to be caused by the text prompt at the top of the main window pane which says 'These items may be restored or deleted' pushing the left column to fit the text instead of wrapping the text to fit the current window size. When there is nothing inside of the bin to delete, it behaves normally. Sorry about the long-winded explanation - it would have been easier with screenshots.

Hope this helps, and thank you for your time.

### LINKED ISSUE #557: Entering Trash Folder Causes the Window to Resize Eratically
Opening the Trash folder when it’s not empty causes the window to resize in a weird way.

Files window before entering trash:

Files window before entering trash

And after:

Files window after entering trash

I'm on Juno Beta 1, all updates installed. Maybe this is caused by the Info-bar at the top and might not be apparent in a different language where the text is shorter.

jeremypw:
Fixes #147 
Fixes #557 

As well as resolving the linked bug, the PR makes some UI changes and other minor bug fixes

* Destructive action styling for permanent delete button
* Can act on selections using buttons (button text changes accordingly)
* Dragging trash tab to a new window maintains correct actionbar visibility
* Actionbar visibility correct when restoring to a trash view
* Actionbar visibility correct when empty trash opened in second window and then files trashed in first window.

Some information text has been removed but this did not seem to add much and was causing the bug because it was not ellipsizable.  The button text and actionbar visibility gives equivalent information.

Translation templates will need updating once additional strings finalized.


tintou:
If you use a `const string` better use `MY_STRING = N_("MyTranslatedString")` and then do `_(MY_STRING)` everywhere

danrabbit:
Is it possible to put the actionbar below the view instead of above it? It looks a little funky with that styling up top

jeremypw:
This has been suggested in the past, but the problem is that it gets obscured by the overlay since the actionbar is added inside the slot whereas the overlay is added by the view container.  One solution would be to inhibit the overlay when in trash (properties would still be accessible via the context menu).

PR #526, opened 2018-06-11
Remove unexpected multiple spaces

jeremypw:


libwidgets/Chrome/ButtonWithMenu.vala
```diff
@@ -121,8 +121,8 @@ namespace Marlin.View.Chrome {
 
             mnemonic_activate.connect (on_mnemonic_activate);
 
-            events |= Gdk.EventMask.BUTTON_PRESS_MASK
-                   |  Gdk.EventMask.BUTTON_RELEASE_MASK;
+            events |= Gdk.EventMask.BUTTON_PRESS_MASK |
```
danrabbit: Hm is this a bug in the linter? I think the way it was before arguably looks nicer. Might be worth discussing?

jeremypw: I follow the convention of having the operator at the end of the line.   If you use spaces other than at the start of the line to align things it makes it difficult for the linter to detect what is unexpected.

PR #546, opened 2018-07-14
Fixed templates label

Djaler:
This PR set fixed templates submenu label, regardless of directory name. This makes it possible to change the location of the template directory and also additional opportunities for l10n


Djaler:
Travis build failed not because of me 👆

Djaler:
@jeremypw check it now, please

PR #550, opened 2018-07-15
Activate in idle

### LINKED ISSUE #549: Crash with activate directory after click on bookmark
This does not always happen but this is how I reproduced the problem:

* Use List View
* Start away from Home
* Click Home bookmark
* Click on a subfolder (on a white part of the name field, after the name) _as soon as possible._

When carried out rapidly Files crashes maybe one in five times.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/60902556-crash-with-activate-directory-after-click-on-bookmark?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #549 

Delays activating items when clicked on until button_release handler ends (both primary and middle clicks).  
Stops sporadic segmentation fault.


naaando:
I tested files for 2 minutes clicking Home > end of label on some folder and crashed files 6 times, couldn't reproduced with this PR, so I believe it fixes the issue.

Method: 
Ran **io.elementary.files** for 2 minutes, **6 crashes**
```
set counter 0
while true
    io.elementary.files
    set counter (math "$counter+1")
end
```

Ran **this PR** for 2 minutes, **0 crashes**
```
set counter 0
while true
    ./src/io.elementary.files
    set counter (math "$counter+1")
end
```

jeremypw:
@naaando Thanks for the testing! That is very helpful.

PR #555, opened 2018-07-21
Refactor PF.AppChooserDialog

### LINKED ISSUE #554: Open in --> Other Application... has wrong headerbar
## Actual behavior

screenshot from 2018-07-21 18-59-54

## Expected behavior

screenshot from 2018-07-21 18-59-54-2

## Steps to reproduce the behavior

1. Open Files
2. Right-click and select "Open in" --> "Other Application..."
3. The dialog has wrong headerbar

I'm not sure this issue is whether a bug or intentional feature.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/61158303-open-in-other-application-has-wrong-headerbar?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
PF.AppChooserDialog now derived from Gtk.Dialog, fixing the linked issue and allowing more flexibility in design.

When showing unrelated applications the set as default box is unchecked to avoid accidentally setting an unsuitable app as default for that type.

Required content type is derived from pre-existing GOF.File rather than querying again.

### DIFF
--- a/src/Dialogs/ChooseAppDialog.vala
+++ b/src/Dialogs/ChooseAppDialog.vala
@@ -19,51 +19,100 @@
 * Authored by: Jeremy Wootten <jeremy@elementaryos.org>
 */
 
-class PF.ChooseAppDialog : Object {
-    Gtk.AppChooserDialog dialog;
-    Gtk.CheckButton check_default;
+class PF.ChooseAppDialog : Gtk.Dialog {
+    private Gtk.AppChooserWidget app_chooser;
+    private Gtk.CheckButton check_default;
+    private string content_type;
 
     public GLib.File file_to_open { get; construct; }
-    public Gtk.Window parent { get; construct; }
 
     public ChooseAppDialog (Gtk.Window? parent, GLib.File file_to_open) {
-        Object (parent: parent, file_to_open: file_to_open);
+        Object (transient_for: parent,
+                resizable: false,
+                deletable: false,
+                window_position: Gtk.WindowPosition.CENTER_ON_PARENT,
+                destroy_with_parent: true,
+                file_to_open: file_to_open
+        );
     }
 
     construct {
-        dialog = new Gtk.AppChooserDialog (parent, Gtk.DialogFlags.MODAL | Gtk.DialogFlags.DESTROY_WITH_PARENT, file_to_open);
-        dialog.deletable = false;
+        /* Called from view so corresponding GOF.File must exist and already have content type */
+        content_type = (GOF.File.@get (file_to_open)).get_ftype ();
 
-        var app_chooser = dialog.get_widget () as Gtk.AppChooserWidget;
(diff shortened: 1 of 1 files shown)


src/Dialogs/ChooseAppDialog.vala
```diff
@@ -19,51 +19,100 @@
...
+        show_switch.active = false;
+        show_switch.state = false;
+
+        show_grid.add (show_label);
+        show_grid.add (show_switch);
+
+        grid.add (show_grid);
+        content_area.add (grid);
+
+        add_button (_("Select"), Gtk.ResponseType.OK);
```
ryonakano: I wonder why this is not `add_button (_("_Select"), Gtk.ResponseType.OK);`.

src/Dialogs/ChooseAppDialog.vala
```diff
@@ -19,51 +19,100 @@
...
+        add_button (_("Cancel"), Gtk.ResponseType.CANCEL);
```
ryonakano: Same as above, `add_button (_("_Cancel"), Gtk.ResponseType.CANCEL);`

jeremypw:
Elementary apps tend not to have button mnemonics - not sure why - I'll check with the UX team.

jeremypw:
In this case the buttons are keyboard accessible anyway - use the arrow and Enter keys. There may be an argument for making the checkbox and switch keyboard accessible.

danrabbit:
For reference:

screenshot from 2018-07-22 10 30 44 2x

jeremypw:
Thanks for the feedback.  

What do you suggest instead of the switch?  I did not like the large button much.

I must admit I did not realise the search bar existed!  Is this what causes the horizontal line that was the subject of the issue?

danrabbit:
I'm not sure. I think I kind of prefer the big button to the switch but I'm not sure what would be even better than that

Yeah the search bar is where the horizontal line comes from. I wonder if it might be easier to try to get rid of this from the stylesheet side and we can consider the overall design of this dialog some other time as part of a Granite widget or something.

jeremypw:
@ryonakano UX team say button mnemonics are not desirable at the moment because of translation implications. They may be considered in the future.

jeremypw:
It seems that Gtk.AppChooserDialog uses a function of the widget to link in the search entry which is kindly not exposed publicly.  Therefore there is no simple way to reproduce this functionality at the moment.  It would be necessary to write a new AppChooser widget which is not really worthwhile.  I'll leave this open for a while in case someone knows different.

Issue #556, opened 2018-07-22
[Feature Request] Append Slash When Autocompleting in Path Bar


amxmln:
I’m not sure whether this is a bug or intended behaviour, but if I switch directories by using Ctrl+L to enter the path-bar and then type the names of folders, I always have to manually enter the slash after autocompleting with Tab. This gets very frustrating very quickly because it’s not consistent with the way autocomplete works in the Terminal for example—is there any reason behind it behaving like that?

jeremypw:
I guess that if the foldername is unambiguously a completed folder then a slash should be entered automatically.  Care would need to be taken to deal names that may or may not be a folder or may not be complete.

jeremypw:
This is addressed by PR #516

amxmln:
Maybe take a look at how it’s implemented in GNOME Files—there it behaves like in the Terminal.

jeremypw:
Its not quite the same situation as in Terminal.  In Terminal, you can only cd to a folder.  However, in FIles it is possible to navigate to a regular file - in this case Files automatically navigates to the parent folder and highlights the requested file.

jeremypw:
Try creating e.g. a folder named "Documents" and a file named "documents". In Terminal, typing `cd doc` and pressing `Tab` completes to `cd Documents/`  In Files it completes to "..../documents" if you type "doc" + Tab and "Documents" if you type "Do".  However, this has revealed a bug in that under these circumstances (using the home directory), as it should not propose a completion at this point since there is also a folder "Downloads".  Moreover typing "Doc" causes the proposed completion to disappear (?!)

jeremypw:
I'll create a new issue about this.

cassidyjames:
I'm not sure that navigating to a specific file from the pathbar is a useful behavior, especially when the path bar is used primarily for navigating between folders. I think I'd like the tab appending behavior as well.

jeremypw:
Its mainly for consistency of behaviour with bookmarks and commandline options I think. I have never used it.  I am about to submit a PR that fixes the issues mentioned here without losing that behaviour, but it would be simple enough just to ignore files that are not folders when completing.

amxmln:
I agree with Cassidy, I had no idea selecting files through the pathbar was even a thing—it seems somewhat counter-intuitive to begin with.

Thanks for looking at this and solving it in any case! You're all doing a great job.

jeremypw:
[...]

PR #561, opened 2018-07-23
Only complete folders

jeremypw:
See comments to #556 

With this PR, only completions of directory names are offered in the pathbar, similar to Terminal completion.  Typing a regular file path will still open that files parent folder

This branch is based on #559


danrabbit:
Requesting a review from @cassidyjames for maybe some more clarity on expecting vs experienced behavior. Does having the ability to complete files cause some interactions problems? It seems from this PR like we don't save any code complexity by disallowing it

cassidyjames:
@danrabbit I think it's a couple of things:

1. The case where it affects interaction is when there is a file and a directory with the same (case-insensitive) name. I would expect it to always prefer the directory, since you're typing in the path bar.
2. It feels a little weird to type in/complete a file in the first place, when this context is navigating to a folder. But I'm less concerned about this if it doesn't get in the way (which it does now), and just prefers the directory instead.

But I'm also not super concerned about this interaction to start with :stuck_out_tongue: 

I can give it a try when I get a chance, though.

jeremypw:
@cassidyjames It is only the completion function that is case-insensitive and in that case it ignores the regular files (in the PR).  If you type the full path in the pathbar then it is case-sensitive and it will load exactly what you typed.  So there could only be a problem if, in the situation you propose, the folder was in upper case, the regular file in lower case and the user typed the full name in lower case rather than using completion intending to navigate to the folder. Fairly unlikely.

PR #567, opened 2018-07-28
Meson build

jeremypw:
Keeping cmake compatability for now for safety.  Considerably more code lines will be saved when that is dropped.

Significant changes made to the connect server functionality in order to simplify porting to meson.  Those changes are also the subject of a separate PR, which should be merged first. Update: now merged.


jeremypw:
@leonardo-lemos :  Which /data files are you referring to?  There some files that are used by cmake and not meson (and vice versa).

jeremypw:
:-<  Seems to have been some bit rot since this was first pushed.  Just compiling and installing is not enough! I check it more thoroughly for regressions.

jeremypw:
@leonardo-lemos :  I have fixed the problems with the .ui file and pkexec.  It wasn't bit rot - I just got it wrong :-(

PR #568, opened 2018-07-29
Move ConnectServer Dialog and Operation to libcore

jeremypw:
* Use tintou port of Dialog to vala (modified)
* Lose network core plugin
* Sidebar inserts and handles ConnectServer item itself
* Avoid external function
* Connect server dialog always handled by Window (from hotkey or sidebar).

This is in preparation for a meson build


libcore/ConnectServerDialog.vala
```diff
@@ -0,0 +1,530 @@
...
+        connect_button.no_show_all = true;
+        connect_button.show ();
+        connect_button.clicked.connect (on_connect_clicked);
+
+        continue_button = new Gtk.Button.with_label (_("Continue"));
+        continue_button.no_show_all = true;
+        continue_button.hide ();
+        continue_button.clicked.connect (on_continue_clicked);
+
+        ((Gtk.Container)(get_action_area ())).add (cancel_button);
```
danrabbit: To avoid deprecation warnings about `get_action_area ()` you can do something like `continue_button = add_button (_("Continue"));`

libcore/ConnectServerDialog.vala
```diff
@@ -0,0 +1,530 @@
...
+    public string server_uri {get; private set; default = "";}
+
+    public ConnectServerDialog (Gtk.Window window) {
+        Object (
+            transient_for: window
+        );
+
+        var t = new Gtk.Label (_("Connect to Server"));
+        /*FIXME Appropriate style class */
+        t.get_style_context ().add_class (Granite.STYLE_CLASS_H2_LABEL);
```
danrabbit: [`Granite.STYLE_CLASS_PRIMARY_LABEL`](https://valadoc.org/granite/Granite.STYLE_CLASS_PRIMARY_LABEL.html) :)

libcore/ConnectServerDialog.vala
```diff
@@ -0,0 +1,530 @@
...
+*
+* You should have received a copy of the GNU General Public
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1335 USA.
+*
+* Authored by: Corentin Noël <corentin@elementary.io>
+*/
+
+public class PF.ConnectServerDialog : Gtk.Dialog {
```
danrabbit: Could avoid some layout work with [`Granite.MessageDialog`](https://valadoc.org/granite/Granite.MessageDialog.html)

libcore/ConnectServerDialog.vala
```diff
@@ -0,0 +1,530 @@
...
+        deletable = false;
+        resizable = false;
+
+        info_bar = new Gtk.InfoBar ();
+        info_bar.message_type = Gtk.MessageType.INFO;
+        info_bar.no_show_all = true;
+        info_bar.hide ();
+
+        var info_grid = new Gtk.Grid ();
+        info_grid.orientation = Gtk.Orientation.HORIZONTAL;
```
danrabbit: Grid is horizontal by default, so you don't need to specify this

libcore/ConnectServerDialog.vala
```diff
@@ -0,0 +1,530 @@
...
+        continue_button = new Gtk.Button.with_label (_("Continue"));
+        continue_button.no_show_all = true;
+        continue_button.hide ();
+        continue_button.clicked.connect (on_continue_clicked);
+
+        ((Gtk.Container)(get_action_area ())).add (cancel_button);
+        ((Gtk.Container)(get_action_area ())).add (connect_button);
+        ((Gtk.Container)(get_action_area ())).add (continue_button);
+
+        key_press_event.connect ((e) => {
```
danrabbit: Instead of doing this, we should use the Gtk.Entry property [`activates_default`](https://valadoc.org/gtk+-3.0/Gtk.Entry.activates_default.html) and set the connect button as default. I think this is just with [`has_default`](https://valadoc.org/gtk+-3.0/Gtk.Widget.has_default.html) but I'm not sure what the difference is between that and `can_default` or `receives_default` or if it needs a combination of those. Either way, there's a native way to do this in Gtk+ without listening to key events :)

libcore/ConnectServerDialog.vala
```diff
@@ -0,0 +1,530 @@
...
+        });
+
+        key_release_event.connect ((e) => {
+            return true;
+        });
+
+        var grid = new Gtk.Grid ();
+        grid.margin_start = grid.margin_end = 12;
+        grid.margin_bottom = 6;
+        grid.orientation = Gtk.Orientation.VERTICAL;
```
danrabbit: If you're just using attach, probably don't need to set orienation

libcore/ConnectServerDialog.vala
```diff
@@ -0,0 +1,530 @@
...
+        grid.attach (user_header_label, 0, 5, 4, 1);
+        grid.attach (domain_label, 0, 6, 1, 1);
+        grid.attach (domain_entry, 1, 6, 3, 1);
+        grid.attach (user_label, 0, 7, 1, 1);
+        grid.attach (user_entry, 1, 7, 3, 1);
+        grid.attach (password_label, 0, 8, 1, 1);
+        grid.attach (password_entry, 1, 8, 3, 1);
+        grid.attach (remember_checkbutton, 1, 9, 3, 1);
+
+        var label_sizegroup = new Gtk.SizeGroup (Gtk.SizeGroupMode.HORIZONTAL);
```
danrabbit: What's this size group used for? It seems like all of these labels are in the Grid above?

libcore/ConnectServerDialog.vala
jeremypw: That was the in the first version but there are problems with that - using that function associates the button with a response signal and I do not want the buttons always to emit a response because that causes the dialog to close prematurely in some cases.  Although this is a subclass of Gtk.Dialog it continues to be active after the "Connect button" is pressed so that it can deal with requests from the mount operation to enter further details.  Also during the "connect phase" pressing "Cancel" should end the mount operation but not necessarily close the dialog.

Maybe I can put the buttons in the content area instead.

libcore/ConnectServerDialog.vala
jeremypw: I'll try it, thanks.

libcore/ConnectServerDialog.vala
jeremypw: OK, I had a feeling there was a better way.

libcore/ConnectServerDialog.vala
jeremypw: Good point.

danrabbit:
- [ ] You have duplicated icons here in the infobar. A symbolic icon is already automatically added in Gtk.CSS so you don't need to also add one yourself :)
- [ ] I would also recommend adding the style class [`Gtk.STYLE_CLASS_FRAME`](https://valadoc.org/gtk+-3.0/Gtk.STYLE_CLASS_FRAME.html) since the infobar is kind of floating and not really being revealed from under a headerbar

Before:

screenshot from 2018-07-30 08 14 26 2x

After:

screenshot from 2018-07-30 08 17 36 2x

danrabbit:
- [ ] Another thing I noticed is that the "Connect" button is still sensitive while the connection is loading. This should probably be insensitive since presumably clicking it again doesn't do anything while we're already loading

PR #577, opened 2018-08-03
Render correct icon for open special directories in view

### LINKED ISSUE #573: Take XDG dirs into account for open folder icons
We just merged a branch to add icons for `folder-pictures-open`, `folder-music-open`, etc. Now we just need Files to use them :)

screenshot from 2018-07-30 11 27 06 2x

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/61519218-take-xdg-dirs-into-account-for-open-folder-icons?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #573 

Note: sidebar icons do not show open-closed status.  To achieve this would require substantial refactoring.

This solution also assumes a particular icon naming scheme (which is true for the elementary icon scheme but not necessarily other schemes).  This is for simplicity.  A flexible solution would be significantly more difficult.


danrabbit:
Is it possible/easy to enable icon name fallbacks here? I just encountered an issue where if an icon name is incorrect/missing Files shows no icon when it would be nice if it would at least fallback to the next available icon (in this case probably `folder-foo` or at least `folder`)

danrabbit:
This could probably be another branch, but can we make this apply to drag accept also?

jeremypw:
I have implemented drag-accept for special directories as well.

The issue of fallingback (or not) seems to be associated with the Marlin.IconInfo  lookup and more particularly using `theme.choose_icon_for_scale ()` which does not accept the generic fallback flag for some reason.  I'll see if there is a workround but if you can give an example for testing it would help.

PR #579, opened 2018-08-04
Do not highlight unless hovered - show cursor position differently 

### LINKED ISSUE #535: Items stay in the "illuminated" state after deselecting
Files version: 0.3.6
elementary OS version: Juno beta

Steps to reproduce:

1. Select one or more items using the check mark or the keyboard:

screenshot from 2018-07-05 17-30-20

Notice that the last selected item appears illuminated (same as the hover state)

2. Clear the selection either by clicking elsewhere or by pressing <kbd>ESC</kbd>.

screenshot from 2018-07-05 17-31-38

The last selected item still stays illuminated.

This does not happen when the items are selected by dragging a rectangular area; only when the items are selected one by one.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/60524318-items-stay-in-the-illuminated-state-after-deselecting?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #552: First item is illuminated without selection.
Files version: 0.3.6
elementary OS Version: Juno Beta 1

By default when I open files or a directory in files, the first item displayed is always "illuminated" despite not ever being selected or hovered over. This is similar to what's reported in #535 but different in that no selection has to be made.

screenshot from 2018-07-17 09 56 18
screenshot from 2018-07-17 09 59 10

jeremypw:
Fixes #552 
Fixes #535 

Instead of prelighting the icon when focused, a small "selection-add" emblem is shown. This indicates where the cursor and also that the item can be selected using a hotkey (currently `<Ctrl><Alt>Plus/Equal`).  However, another way of indicating cursor position using CSS may be preferred. The cursor position on a selected item is shown by colorizing as before.

The opportunity is taken to fix a discrepancy between IconView and ListView that becomes apparent with these changes - moving the cursor in IconView using `<Ctrl><ArrowKey>`  now just moves the cursor, and does not select the item.  This facilitates non-contiguous selecting using the keyboard in IconView and matches ListView behaviour.


src/IconRenderer.vala
```diff
@@ -197,16 +197,21 @@ namespace Marlin {
             style_context.restore ();
 
             /* Do not show selection helpers or emblems for very small icons */
-            if ((selected || prelit) && file != drop_file) {
-                helper_size = Marlin.IconSize.LARGE_EMBLEM > (int.max (pixbuf.get_width (), pixbuf.get_height ()) / icon_scale) / 2 ?
-                              Marlin.IconSize.EMBLEM : Marlin.IconSize.LARGE_EMBLEM;
+            if ((focused || selected || prelit) && file != drop_file) {
+                if (focused &&!(selected || prelit)) {
+                    /* Always use small helper to show focus (use smaller?) to minimise obscuration of icon*/
```
danrabbit: This kind of looks like a bug imo since the icon size changes as soon as you move the cursor or focus

danrabbit:
Yeah, I would ideally like to do something like this for focus to match the list and column views:
```
iconview.view.cell:focused {
    background: @colorAccent;
}
```
But you'll see that if we do that every cell gets colored since the whole view is technically where Gtk is detecting focus.

For reference, here's what I had mocked up when we talked about this previously:

bitmap

src/IconRenderer.vala
jeremypw: Yes, I was not sure about that. The reason I used a small helper to indicate focus was because the linked issue objected to the appearance of the icon changing when focused but not selected.  The small helper is less likely to partially obscure the icon.  However, it would be easy to have the helper stay the same size.  I would consider this a temporary fix pending a CSS fix (if possible).

I'll see if I can get a similar effect programmatically.

jeremypw:
There is no obvious way (to me) of applying CSS to cellrenderers as they are not widgets.  It is possible to paint the background "manually" for each renderer but this is ugly and I have not yet found a way to reproduce the desired effect.

jeremypw:
Closing as no progress made.

PR #584, opened 2018-08-07
Only lose final "\r\n" when pasting text.

### LINKED ISSUE #583: can not copy many file/folder
gifout

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/61771233-can-not-copy-many-file-folder?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #583 

Fixes regression caused by  a1475450098eb65d4cd073acf6222453fb88b146
It is an open question whether, when pasting multiple file paths into a text application, the individual paths should be separated by newlines or, perhaps, spaces.  But for now I have retained the newlines except at the end.  So pasting a single path into terminal for example will not end the command (but pasting more than one will).  Any further change can be made in a future PR, it required.

### DIFF
--- a/src/ClipboardManager.vala
+++ b/src/ClipboardManager.vala
@@ -244,7 +244,6 @@ namespace Marlin {
 
             switch (target_info) {
                 case ClipboardTarget.GNOME_COPIED_FILES: /* Pasting into a file handler */
-warning ("gnome copied");
                     string prefix = manager.files_cutted ? "cut" : (manager.files_linked ? "link" : "copy");
                     DndHandler.set_selection_data_from_file_list (sd,
                                                                   manager.files,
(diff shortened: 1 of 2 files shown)


src/ClipboardManager.vala
```diff
@@ -244,13 +244,15 @@ namespace Marlin {
 
             switch (target_info) {
                 case ClipboardTarget.GNOME_COPIED_FILES: /* Pasting into a file handler */
+warning ("gnome copied");
```
danrabbit: Where these intended to be removed after testing? If not maybe should be a `debug` since it's a little noisy

jeremypw: Sorry, yes should have been removed.

PR #597, opened 2018-08-12
StockDialogs: Use Terminal style for error details

### LINKED ISSUE #596: Error while copying dialog has no wrapping in error details
screenshot from 2018-08-12 11 12 25 2x

This should probably use the nice little fake Terminal pattern from AppCenter and wrap properly

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/62048196-error-while-copying-dialog-has-no-wrapping-in-error-details?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #596 

Makes the dialog resizable because scrolling around inside a little messagebox isn't fun for long messages

## BEFORE
screenshot from 2018-08-12 11 12 25 2x

## AFTER
screenshot from 2018-08-12 11 30 34 2x


danrabbit:
In a HIG-compliant dialog nothing should go under the icon except the action area. See https://github.com/elementary/granite/issues/203

:heavy_check_mark: 

I'm really trying to avoid any "cleverness" because it always gets us into weird situations like dialogs expanding offscreen or some other unpredictable result. I don't want the dialog to be akwardly large by default, but I would like to make sure that the message can be displayed in a window that's a comfortable size, whatever that means for the user. I think especially with Terminal output where we have hard line breaks that being able to resize the window really helps with legibility.

Now that the bug I was using to test this is fixed in master, do you know of an easy way to throw this dialog so I can test new changes?

jeremypw:
OK - it just looks unbalanced to me as it is. 

In this specific case the detailed message is unlikely to be more than a few lines.  I agree that trying to calculate the window size is "too clever".

I triggered this by copying a file to the clipboard, trashing it, and then pasting it back into the folder.

jeremypw:
I seem to be getting a large space below the expander. Cannot reduce the window height to reduce it but reducing window width causes expanded scrolled window to consume the space.  As if minimum window height set?

screenshot from 2018-08-17 17 39 12
screenshot from 2018-08-17 17 39 42
screenshot from 2018-08-17 17 40 03

PR #649, opened 2018-10-10
Sidebar plugin support

naaando:
Adds and reactivate sidebar hooks that trigger when a sidebar is initiated and updated


jeremypw:
Need to make sure that any old sidebar plugins remaining installed on the user's system cannot get loaded into the sidebar, if the cloud provider functionality needs to be implemented as a plugin.

naaando:
The cloud providers functionality adds have a library(that isn't stable) as a dependency, so would be better to have it as a plugin.
Rename the method to avoid the old plugins could be a quick fix but would be better to have some kind of version checking while the plugins are being loaded

jeremypw:
I'm not too sure how much of a problem it is in practice anyway - it was a problem at the time the old plugin was deprecated if you were compiling from source but I think anyone updating from the stable repository should be OK.

jeremypw:
It may be possible to get cmake to empty the Files core plugin folder before installing?

naaando:
I couldn't find on google or on CMake documentation how to delete a folder/files before installing

PR #650, opened 2018-10-10
Transform eject icon into action icon

naaando:
Make eject icon column support any icon
Changes column name from EJECT_ICON to ACTION_ICON

### DIFF
--- a/src/View/Sidebar.vala
+++ b/src/View/Sidebar.vala
@@ -369,7 +369,8 @@ namespace Marlin.Places {
                                                    Volume? volume,
                                                    Mount? mount,
                                                    uint index,
-                                                   string? tooltip = null) {
+                                                   string? tooltip = null,
+                                                   Icon? action_icon = null) {
 
             bool show_eject, show_unmount, can_stop;
             check_unmount_and_eject (mount, volume, drive,
@@ -386,11 +387,8 @@ namespace Marlin.Places {
                 show_eject_button = (show_unmount || show_eject);
             }
 
-            GLib.Icon eject;
             if (show_eject_button) {
-                eject = this.eject_icon;
-            } else {
-                eject = null;
+                action_icon = this.eject_icon;
             }
 
             GLib.Error error = null;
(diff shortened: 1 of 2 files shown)


src/View/Sidebar.vala
```diff
@@ -223,15 +223,15 @@ namespace Marlin.Places {
                                 "active", Column.SHOW_SPINNER,
                                 "pulse", Column.SPINNER_PULSE);
 
-            crpb = new Gtk.CellRendererPixbuf (); /* Icon for eject button (hidden while ejecting or unmounted) */
+            crpb = new Gtk.CellRendererPixbuf (); /* Icon for eject button  (hidden while ejecting or unmounted) and another signs */
```
jeremypw: Double space before bracket in comment. "and another signs" should be "or other icons". Overlong line.  Could probably lose the words in brackets.

src/View/Sidebar.vala
```diff
@@ -223,15 +223,15 @@ namespace Marlin.Places {
...
             this.eject_spinner_cell_renderer = crpb;
             crpb.stock_size = Gtk.IconSize.MENU;
-            crpb.gicon = new ThemedIcon.with_default_fallbacks ("media-eject-symbolic");
             crpb.xpad = ICON_XPAD;
             crpb.ypad = BOOKMARK_YPAD;
 
             col.pack_start (crpb, false);
-            col.set_attributes (crpb, "visible", Column.SHOW_EJECT);
+            col.set_attributes (crpb,
+                                "gicon", Column.ACTION_ICON);
```
jeremypw: Not necessary to split line here.

src/View/Sidebar.vala
```diff
@@ -2110,6 +2109,7 @@ namespace Marlin.Places {
             var rowref = new Gtk.TreeRowReference (store, path);
             store.@set (iter, Column.SHOW_SPINNER, true);
             store.@set (iter, Column.SHOW_EJECT, false);
+            store.@set (iter, Column.ACTION_ICON, null);
```
jeremypw: Not sure why this is necessary.  For this iter, an eject action icon should be appropriate (even if hidden).  It gets set back to an eject icon in `finish_eject_or_unmount ()` anyway.

naaando: It avoids the eject icon to be displayed aside the spinner, when the eject is finished it sets it back if eject has failed.

jeremypw: Ah, OK.  I was just looking at the code - I'll test function asap.

PR #656, opened 2018-10-13
Plugin cloudproviders

naaando:
Almost done, I still having issues trying to "merge" two menus without using a submenu
I'll try to split to more PRs and clean up some things first

[Fork of libcloudproviders with Vala bindings and fixes](https://gitlab.gnome.org/naaando/libcloudproviders/tree/elementary-files-cloud)


jeremypw:
If this is to be split up, best label it as "In Progress".

naaando:
But WIP already means Work In Progress, didn't get the point but I'll rename it.

jeremypw:
Sorry, I meant using GitHub labels (see menu on the right) - I've added one for you now.

naaando:
Oh ofc, but normal users can't set status

jeremypw:
Oh, OK I did not realize that.

libcore/AbstractSidebar.vala
```diff
@@ -127,4 +143,34 @@ namespace Marlin {
                                                   string? tooltip = null,
                                                   Icon? action_icon = null) ;
     }
+
+    public class PluginItem : Object {
```
jeremypw: Public classes should be put in their own file.

plugins/pantheon-files-cloud/meson.build
```diff
@@ -0,0 +1,20 @@
+project('pantheon-files-cloud', 'vala', 'c',
```
jeremypw: I would leave meson build to a separate PR (until Files is migrated to meson).

naaando:
Done and the passing tests, now I just need to check why the example server is looking disabled and if possible test it with a real implementation of libcloudproviders server(is there any?).

peteruithoven:
@naaando I'm afraid Files has moved from CMAKE to Meson in the mean time, resulting in a conflict. 

Have you tried using the Nextcloud client as a libcloudproviders server?

> Available with this release is seamless integration of synced folders into the GNOME file manager sidebar with libcloudproviders.

From: https://nextcloud.com/blog/nextcloud-desktop-client-2.5-is-out-with-end-to-end-encryption-new-login-flow-and-much-more/
More info: https://blog.juliushaertl.de/index.php/2017/08/28/gsoc-final-report/

naaando:
Thanks @peteruithoven, I'd tried but no success before because I didn't know that it had to be compiled from source and to write the libcloudproviders config file `/usr/share/cloud-providers/com.nextcloudgmbh.Nextcloud` myself to enable this functionality (at least this was my way to get this working). Cool thing is that anything now is available from docker, so I managed to setup a nextcloud server within minutes to test this.

image

peteruithoven:
That looks great! Even with the nextcloud icon. 

> it had to be compiled from source

The Nextcloud desktop client? But why?

naaando:
Looks like their debian packaged version(v2.5.2 beta) from PPA don't ship the config file for cloudprovider nor expose the DBus interface, I saw an [open PR to make it work on appimage](https://github.com/nextcloud/desktop/pull/479) so I believe it doesn't ship it also.

naaando:
What about using providers names before account names?

| image | image |
| ------- | ------ |

danrabbit:
> What about using providers names before account names?
> 
> image	image

Kind of seems like this obscures the name of the share. I'd imagine that the service would be recognizable from the icon since this is something the user sets up. Thoughts @elementary/ux?

cassidyjames:
@danrabbit @naaando yeah I think I'd keep it clean with the icon and account name, then if you want, you could put the provider name in the tooltip.

plugins/pantheon-files-cloud/src/plugin.vala
```diff
@@ -0,0 +1,123 @@
...
+
+    You should have received a copy of the GNU General Public License along
+    with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+    Author(s):  Fernando da Silva Sousa <wild.nando@gmail.com>
+***/
+
+public class Marlin.Plugins.Cloud.Plugin : Marlin.Plugins.Base {
+    Marlin.AbstractSidebar? sidebar;
+    CloudProviders.Collector collector = CloudProviders.Collector.dup_singleton ();
```
jeremypw: I would assign this in a construct clause.

plugins/pantheon-files-cloud/src/plugin.vala
```diff
@@ -0,0 +1,123 @@
...
+    CloudProviders.Provider[] providers_connected = {};
+
+    public Plugin () {
+        collector.providers_changed.connect (on_providers_changes);
```
jeremypw: I would assign this in a construct clause.

tintou:
@naaando elementary is now shipping cloudprovider.vapi directly with libcloudproviders-dev, there is no need to include it here anymore :)

naaando:
Awesome, I'll update it tomorow.

Em qua, 3 de abr de 2019 06:44, Corentin Noël <notifications@github.com>
escreveu:

cassidyjames:
@naaando this is looking really good. Have you had a chance to look at @jeremypw's review? Do you know if that's an issue with this branch or with NextCloud itself?

naaando:
> However, selecting "Log out" did not appear to do anything - the menu option was unchanged - but the Nextcloud client indicated that the server had been logged out.

Yeah,  looks like the server is passing status flags syncing and disconnected in dbus on log out, but they continue to be visible, should 'em be hidden from menu?

naaando:
> With this setup, Files did not display the NextCloud set up after just starting the server with `docker run -d -p 8080:80 nextcloud` when Files was already running. After signing into the server using the Nextcloud sync client installed from the above source, Files still did not show the shortcut. After closing Files and reopening, the shortcut was shown and mostly worked as expected.

Couldn't reproduce here, it shows up whenever I open nextcloud client. I'll take a look

plugins/pantheon-files-cloud/plugin.vala
```diff
@@ -0,0 +1,127 @@
+/***
+    Copyright (c) 2018 elementary LLC <https://elementary.io>
```
donadigo: Update the year here.

plugins/pantheon-files-cloud/plugin.vala
```diff
@@ -0,0 +1,127 @@
...
+    You should have received a copy of the GNU General Public License along
+    with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+    Author(s):  Fernando da Silva Sousa <wild.nando@gmail.com>
+***/
+
+public class Marlin.Plugins.Cloud.Plugin : Marlin.Plugins.Base {
+    Marlin.AbstractSidebar? sidebar;
+    CloudProviders.Collector collector;
+    CloudProviders.Provider[] providers_connected;
```
donadigo: Since this is an array made out of non-simple types, I would advise to use `Gee.ArrayList` or `Gee.LinkedList` here. It's not necessary in current state of this branch though.

plugins/pantheon-files-cloud/plugin.vala
```diff
@@ -0,0 +1,127 @@
...
+        account.notify.connect (() => {
+            return_if_fail (sidebar != null);
+            sidebar.update_plugin_item (adapt_plugin_item (provider, account), reference);
+        });
+    }
+
+    /**
+     * Generate a SidebarPluginItem from provider and account informations
+     */
+    Marlin.SidebarPluginItem adapt_plugin_item (CloudProviders.Provider provider, CloudProviders.Account account) {
```
donadigo: Can be static.

plugins/pantheon-files-cloud/plugin.vala
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+
+    void on_providers_changes () {
+        //  Listen to new accounts
+        unowned GLib.List<CloudProviders.Provider> providers = collector.get_providers ();
+        foreach (var provider in providers) {
+            //  Avoid listening to same provider again
+            if (!(provider in providers_connected)) {
+                providers_connected += provider;
```
donadigo: What happens if a provider is removed? From what I see, current code doesn't account for that at all. Please correct me if I'm wrong.

plugins/pantheon-files-cloud/plugin.vala
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /**
+     * Get icon for current account status
+     *
+     * @param a status {@link CloudProviders.AccountStatus} of a {@link CloudProviders.Account}
+     *
+     * @return a error icon if status is error else returns null
+     */
+    Icon? get_icon (CloudProviders.AccountStatus status) {
```
donadigo: Can be static as well.

src/View/Sidebar.vala
```diff
@@ -1611,9 +1613,27 @@ namespace Marlin.Places {
...
+                if (model == null) {
+                    menu.build ().popup_at_pointer (event);
+                } else {
+                    menu
+                    .add_separator ()
+                    .add_open_tab (open_shortcut_in_new_tab_cb)
+                    .add_open_window (open_shortcut_in_new_window_cb)
+                    .add_separator ()
+                    .build_from_model (model, action_group_namespace, action_group)
+                    .popup_at_pointer (event);
```
donadigo: We usually don't use this syntax, but it's fine here. I would indent the lines after menu to be clear where the call chain ends.

lainsce:
By request of @naaando since his PC is currently out of commission, I fixed some of the diff comments for him.

jeremypw:
It would help if there could be a specification of what this PR is intended to achieve (and not achieve) or a pointer to any existing one.  

I guess we can assume for now that the nextcloud server and client will be running before Files is run.
The main issues I have found are:

* If the NextCloud client is closed then the bookmark disappears as expected but it does not reappear when the client is restarted unless Files is restarted as well

* If Files is used to log out of the server then the `Log out` menu option does not change or disappear.  It would be nice if Files could be used to log in as well.

* There is no visible change in the bookmark whether or not the user is logged into NextCloud.

These are not critical and could be fixed later.

lainsce:
> It would help if there could be a specification of what this PR is intended to achieve (and not achieve) or a pointer to any existing one.

From what I understood, it is to make the base on which cloud providers are accessible with Files, with further PRs(fixing those NextCloud bugs, for example) welcome after merging this one.

jeremypw:
@lainsce Thanks.  I guess "accessible in Files" in this context means to control the client through Files since the synced folders are accessible anyway.  This PR provides the following functions (when a working NextCloud server and a logged in client are present) :

* Open the website (access to the uploaded files) - works
* Recently changed files info list - works
* Pause/resume synchronization - works
* Link to NextCloud help webpage - works
* Access NextCloud client settings - launches client UI (but not at settings dialog)
* Log out of client - works (with issue mentioned above)
* Quit sync client - works

I think this is a very reasonable list of functions for a first PR so I am minded to merge this now if no-one has any further comments or objections.

plugins/pantheon-files-cloud/plugin.vala
jeremypw: Ending the client process causes the bookmark to be removed from Sidebar without problems.

donadigo: In which way and where in the code the bookmark is removed from the sidebar? I don't see it. `providers_connected` will always grow in size, because nothing is ever removed from it.

plugins/pantheon-files-cloud/plugin.vala
jeremypw: Changed to ArrayList.

plugins/pantheon-files-cloud/plugin.vala
jeremypw: Added code to remove lost providers from list.

jeremypw:
I tried this with the lates docker image of nextcloud and with version 2.5 of NextCloud client from the ppa (this version is supposed to include libcloudprovider support).  However it does not seem to be working at present. Not sure why - I did get it working on another computer with docker version 18.09.4, build d14af54266 and the associated Nextcloud server image and nextcloud/desktop#1185 for the client/libcloudproviders library.

naaando:
Damn, so it's done? Sorry for not finishing this myself, as lains pointed out my PC has broke and I couldn't fix it yet.

jeremypw:
I think I will put my recent failure to get a NextCloud setup working this down to my faulty/incompatible NextCloud setup as I did get it working previously and nothing has changed that would break this - and it does not cause regressions in the absence of cloudprovider clients.  I'll merge it.

f-ricchiuto:
Hello, I've read different threads and all seems to point here to this... Sorry but I cannot understand if support for Google Drive has been implemented, and if yes, how to have it working. 
Thanks :-)

jeremypw:
This PR makes Files aware of any cloudprovider that have installed on the system and uses the mechanism provided by libcloudproviders.  Neither Files nor libcloudproviders provide any clients themselves. Nor does elementary ship any by default at the moment. 

More info can be found here: https://github.com/juliushaertl/libcloudproviders

The only provider I know provides libcloudprovider support at the moment is NextCloud but it needs a ppa to get the correct version.  Hopefully someone will develop a suitable client for Google Drive in due course.

PR #660, opened 2018-10-13
Plugin friendly api

naaando:
Depends on #659 for using start_spinner method


libcore/AbstractSidebar.vala
```diff
@@ -128,5 +130,47 @@ namespace Marlin {
...
+         * Adds plugin item to TreeStore or update if a iter is present
+         *
+         * @param item is a {@link Marlin.AbstractSidebar.PluginItem} that resumes
+         * every column a plugin would need of the TreeStore
+         *
+         * @param iter receives the iterator that points to TreeModel to be updated
+         *
+         * @return TreeIter to be used later to update the item
+         */
+        public abstract Gtk.TreeIter add_or_update_plugin_item (PluginItem item, Gtk.TreeIter? iter = null);
```
jeremypw: Can we be sure the iter remains valid as long as needed?  Would a RowReference be better?

libcore/AbstractSidebar.vala
```diff
@@ -128,5 +130,47 @@ namespace Marlin {
...
+        public Mount? mount { get;set; }
+        public Icon? icon { get;set; }
+        public uint index { get;set; }
+        public bool can_eject { get;set; }
+        public string? tooltip { get;set; }
+        public Icon? action_icon { get; set; }
+        public bool show_spinner { get;set; default = false; }
+        public uint64 free_space { get;set; default = 0; }
+        public uint64 disk_size { get;set; default = 0; }
+        public MenuModel? menu_model { get;set; }
```
jeremypw: Code style x 13:  There should be spaces after all the semi-colons

libcore/AbstractSidebar.vala
```diff
@@ -128,5 +130,47 @@ namespace Marlin {
...
+
+        public bool is_bookmark () {
+            return place_type == Marlin.PlaceType.BOOKMARK;
+        }
+
+        public bool is_category () {
+            return (place_type == Marlin.PlaceType.BOOKMARKS_CATEGORY) ||
+                (place_type == Marlin.PlaceType.PERSONAL_CATEGORY)  ||
+                (place_type == Marlin.PlaceType.STORAGE_CATEGORY)   ||
+                (place_type == Marlin.PlaceType.NETWORK_CATEGORY);
```
jeremypw: Parameters look misaligned on web browser.

libcore/PluginItem.vala
```diff
@@ -0,0 +1,48 @@
...
+    public string? tooltip { get; set; }
+    public Icon? action_icon { get; set; }
+    public bool show_spinner { get; set; default = false; }
+    public uint64 free_space { get; set; default = 0; }
+    public uint64 disk_size { get; set; default = 0; }
+    public MenuModel? menu_model { get; set; }
+    public PluginCallbackFunc? cb;
+
+    public bool is_bookmark () {
+        return place_type == Marlin.PlaceType.BOOKMARK;
```
jeremypw: Since place_type has a const value PlaceType.PLUGIN_ITEM this function will always return false; so comparison is not needed.

libcore/PluginItem.vala
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    public bool is_category () {
+        return (place_type == Marlin.PlaceType.BOOKMARKS_CATEGORY) ||
+               (place_type == Marlin.PlaceType.PERSONAL_CATEGORY)  ||
+               (place_type == Marlin.PlaceType.STORAGE_CATEGORY)   ||
+               (place_type == Marlin.PlaceType.NETWORK_CATEGORY);
+    }
```
jeremypw: Since place_type has a const value PlaceType.PLUGIN_ITEM this function will always return false; so comparison is not needed.

src/View/Sidebar.vala
```diff
@@ -441,6 +441,46 @@ namespace Marlin.Places {
...
+        public override Gtk.TreeIter add_or_update_plugin_item (PluginItem item, Gtk.TreeIter? iter = null) {
+            Gtk.TreeIter parent;
+            store.get_iter (out parent, network_category_reference.get_path ());
+
+            if (iter == null) {
+                store.append (out iter, parent);
+            } else if (!store.iter_is_valid (iter)) {
+                return iter;
+            }
+
```
jeremypw: If this is only supposed to be used by plugin items, shouldn't we check that iter points to a plugin item?

src/View/Sidebar.vala
```diff
@@ -441,6 +441,46 @@ namespace Marlin.Places {
             return iter;
         }
 
+        public override Gtk.TreeIter add_or_update_plugin_item (PluginItem item, Gtk.TreeIter? iter = null) {
+            Gtk.TreeIter parent;
+            store.get_iter (out parent, network_category_reference.get_path ());
```
jeremypw: Restricts future plugin items to be in the network category?

naaando: Sure, so I'll split add and update functions then so the params don't get too long.

PR #670, opened 2018-10-19
Record iconrender helper rect only for prelit icons (fixes #668).

### LINKED ISSUE #668: Select with helper can fail when Control pressed.
To reproduce:

* Use Icon View
* Select three files using the mouse with the Control held down.
* Keeping the Control held down, deselect the middle file
* Keeping the Control held down, try to reselect the middle file

The middle file cannot be reselected with the helper under these conditions. (It can be reselected with the icon).

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/64975744-select-with-helper-can-fail-when-control-pressed?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #668 

The iconrender helper_rect property can sometimes not correspond to the one the mouse is over.  This property is changed so that it only records the hovered helper.


src/IconRenderer.vala
```diff
@@ -277,12 +279,12 @@ namespace Marlin {
         }
 
         public override void get_preferred_width (Gtk.Widget widget, out int minimum_size, out int natural_size) {
-            minimum_size = (int)icon_size + helper_rect.width;
+            minimum_size = (int)icon_size + hover_helper_rect.width;
```
leonardo-lemos: Just a minor code style issue here.

src/IconRenderer.vala
```diff
@@ -277,12 +279,12 @@ namespace Marlin {
...
             natural_size = minimum_size;
         }
 
         public override void get_preferred_height (Gtk.Widget widget, out int minimum_size, out int natural_size) {
-            minimum_size = (int)icon_size + helper_rect.height / 2;
+            minimum_size = (int)icon_size + hover_helper_rect.height / 2;
```
leonardo-lemos: And here.

Issue #712, opened 2018-10-27
White / Light File-Thumbnails are Invisible


amxmln:
Images that are mostly white, especially icons, aren't visible unless selected. This creates a very broken-looking, inconvenient user experience:
![bildschirmfoto von 2018-10-27 15-04-50](https://user-images.githubusercontent.com/15271679/47604402-ae59b980-d9f9-11e8-90db-d471267904e0.png)
![bildschirmfoto von 2018-10-27 15-06-00](https://user-images.githubusercontent.com/15271679/47604412-d6491d00-d9f9-11e8-8d93-506db5672fd0.png)

Perhaps in cases like this there should be some sort of border around the thumbnail? Or a darkened background color? It's a tricky thing to fix, but I believe it would be greatly beneficial if a solution could be found.

Keep up the great work!

jeremypw:
I guess an off-white background might be acceptable (perhaps as an option like in Code) but needs the UX-team to alter the style sheet.  Icons that are mostly the same color as the background (whatever it is) are always going to be hard to see. 

I doubt a border would be acceptable.

amxmln:
If a border is truly not an option (why not?), then an off-white background is probably going to be the best solution, I think there's much more pure-white out there than, say, a light grey.

jeremypw:
You are welcome to submit a mockup of how a border would look in the case shown above.

Akryum:
In Windows and MacOS, they chose to use a border shadow:

![windows](https://mspoweruser.com/wp-content/uploads/2016/03/Windows-File-Explorer.jpg)

![macos](https://support.apple.com/library/content/dam/edam/applecare/images/en_US/social/macos-mojave-macbook-finder-social-card.jpg)
(Note the additional inner padding)

Gnome went with a border and inner padding:

![gnome](https://www.gnome.org/wp-content/uploads/2012/09/files-recent.png)

IMO the best looking solution is border-shadow + inner padding

jeremypw:
@elementary/ux  Any opinion on this?

cassidyjames:
I would not be terribly opposed to looking a subtle border + shadow or something on thumbnails (but not other icons, obviously).

Akryum:
The border makes it clear it's a thumbnail with a transparent or white image, and also where you can click with the mouse to open it.

Akryum:
Maybe some smart logic could be applied, if the thumbnail is mostly transparent or white, add a checkerboard. 😸

jeremypw:
[...]

PR #715, opened 2018-10-28
Scale up icon on hover if much smaller than requested size

jeremypw:
See #712.

An experimental branch which magnifies very small icons (less than half the requested size) when hovered.  This does not affect the great majority of icons at normal zoom levels but may ameliorate the linked issue without changing the style-sheet.

TO TEST.

Browse e.g. /usr/share/icons/elementary/actions/16/ in Icon View.

### DIFF
--- a/src/IconRenderer.vala
+++ b/src/IconRenderer.vala
@@ -178,6 +178,30 @@ namespace Marlin {
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


jeremypw:
UX team solution is to embed small icon on suitably side card rather than scale up.

PR #720, opened 2018-10-30
Spelling: URIs, each in their own tab

comradekingu:
Possibly "own seperate tab", as opposed to (hypothetically) something that might be a pre-existing tab belonging to the URI itself.


jeremypw:
"each in a new tab" already means each URI gets its own new tab, so I see little reason to make it even more verbose.  I agree URI should be capitalized though.

Need to run `make po` and `make pot` when string finalized (in separate commits)

PR #749, opened 2018-11-12
Start porting the simple methods of GOF.File to Vala

tintou:


jeremypw:
Is it a good idea in terms of discoverability to put these methods in the .vapi file or is that a temporary measure pending complete porting to vala?

tintou:
@jeremypw It's a temporary measure as the GOF.File is really big I think that it's better to not have a big merge request but split it :)

jeremypw:
@tintou: OK, that's fine.  I would consider splitting out the GOFFile binding to its own gof-file.vapi as the amount of code is quite large compared to the 'normal' bindings. This file would then be replaced by gof-file.vala eventually.

tintou:
@jeremypw okay, I just did what you suggested. (the CMake diff are bigger than expected but it won't last much longer thanks to Meson)

jeremypw:
@tintou :  Can you confirm that you have ported the code exactly (or as closely as possible for it to compile and run) regardless of any possible future improvements?  I have not noticed any differences or regressions yet but will continue to test a bit longer.

tintou:
@jeremypw Yes, I kept the same logic while using the most Vala-ish syntax

PR #752, opened 2018-11-14
Use GLib.Source.REMOVE and  GLib.Source.CONTINUE

jeremypw:
To improve clarity, `return false` and ` return true` are replaced by `return GLib.Source.REMOVE` and  `return GLib.Source.CONTINUE` in Source closures.

A few drive-by code-style improvements are made.


tintou:
Nice thing :)

PR #754, opened 2018-11-14
Fix unequal and incorrect disk usage bars

### LINKED ISSUE #734: File System disk usage bar showing wrong value
screenshot from 2018-11-03 12-00-53

Hovering over the progress bar shows the following tooltip: `/ - ext3/ext4 (437.0 GB free of 501.2 GB`. The progress bar should be at about 8%, but it seems to be stuck in the beginning.

  * Files 4.0
  * elementary OS 5.0 Juno (64-bit)
  * Ubuntu 18.04 LTS
  * Linux 4.15.0-38-generic
  * GTK+ 3.22.30
  * ThinkPad T450
  * SAMSUNG MZ7LN512HCHP-000L1 (512 GB SSD)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65398801-file-system-disk-usage-bar-showing-wrong-value?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #740: Disk Usage bars have not the same size!
The size of Disk Usage bar for "File system" is not equal to other bars of the same type.

OS: Juno up to date

screenshot from 2018-11-06 23 59 55
 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65498446-disk-usage-bars-have-not-the-same-size?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #740 
Fixes #734 

image

The renderer for the storage bars in the sidebar is refactored to fix some issues with the total and used lengths.

Also, some hard-coded numbers are replaced with constants. 

The bar was made slightly thicker to improve visibility (requires UX team opinion).


jeremypw:
Although the class is called "DiskRenderer" it is actually used to render all the sidebar text cells so could do with renaming.

danrabbit:
It looks like the trough and the actual progressbar disagree about their radius :)

I'm wondering, can we use the styles from `levelbar.continuous block.filled` and `levelbar.continuous block.empty`? That would probably give us the most native looking design and then we could add a rule for like `.siderbar levelbar block { min-height: 3px; }` or something

jeremypw:
@danrabbit : You are right, although its not noticeable on my screen at normal size.  I'll try using the suggested styles.

jeremypw:
@danrabbit : I think I'll need your help with this.  Using the suggested styles directly did not work. Did you intend to create a custom CSS provider?

jeremypw:
I have achieved this using Granite.StorageBar style classes and using Cairo to draw the frame (which wasnt actually visible before), but I suspect a more elegant solution is possible.

screenshot from 2018-11-16 12 12 51


At 3X:

screenshot from 2018-11-16 12 12 16

jeremypw:
The latest commit uses storagebar classes and produces:

screenshot from 2018-11-16 16 12 43

danrabbit:
Ah so the rendering here isn't quite right because what you've done matches `.storage-bar.trough`, where one widget has two classes. But how it works for the actual widget is that there is an outer widget `.storage-bar` that contains another widget `.trough` inside of it.

jeremypw:
Latest PR gives the following:

screenshot from 2019-01-11 19 28 03

jeremypw:
@danrabbit :  granite-widgets.css does contain some CSS for .source-list.level-bar and .source-list.fill-block but I had to use a custom CssProvider to get the colors similar to those in master.

jeremypw:
Correction: It does in fact work OK without the custom CSS.

jeremypw:
@danrabbit Is this OK now?

PR #769, opened 2018-11-17
Bookmark: fix memory leak and simplify some functions

tintou:


libcore/Bookmark.vala
```diff
@@ -24,13 +24,13 @@ namespace Marlin {
         public signal void contents_changed ();
         public signal void deleted ();
 
-        private string custom_name = "";
+        private string? custom_name = null;
```
jeremypw: Is there an advantage to using `null` here?  There is a school of thought that using `null` unnecessarily is not a good idea.

tintou: null is handled anyway so I prefer null over "" here, the (tiny) advantage is that there is no string allocation when using null

PR #771, opened 2018-11-18
Fix open with default for folders

Mohelm97:
Open in Terminal is not working when you're not selecting any folder, but if you choose to Open in -> Terminal it works fine and opens Terminal in the (Not selected but current) Folder.


jeremypw:
So, I think what this fixes is:

1) Set Terminal as the default app for folders using the "Other Application ..." dialog.
2) This produces an extra option "Open in Terminal" in the root context menu.
3) This option does not work for the background folder only for folder items in the view.

Mohelm97:
Yes, The option in the root context menu appears in last updates.

PR #775, opened 2018-11-19
Several fixes to callback-based operations

tintou:


src/View/AbstractDirectoryView.vala
```diff
@@ -975,34 +971,25 @@ namespace FM {
...
-            }
             /* Start to rename the file once we get signal that it has been added to model */
-            view.slot.directory.file_added.connect_after (view.after_new_file_added);
-            view.unblock_directory_monitor ();
+            slot.directory.file_added.connect_after (after_new_file_added);
+            unblock_directory_monitor ();
         }
 
-        /** Must pass a pointer to an instance of FM.AbstractDirectoryView as 3rd parameter when
+        /** Must pass a pointer to an instance of FM.AbstractDirectoryView as 2nd parameter when
```
jeremypw: Not relevant now?

tintou: You're right, the annotation is here to inform any C user anyway, I removed the comment.

PR #776, opened 2018-11-20
Fix ownership mismatch using GLib.List, spotted with latest Vala

tintou:


libcore/gof-callwhenready.vala
```diff
@@ -28,11 +28,11 @@ public class GOF.CallWhenReady : Object {
 
 
     public CallWhenReady (GLib.List<GOF.File> _files, call_when_ready_func? _f = null) {
-        files = _files.copy ();
+        files = _files.copy_deep ((GLib.CopyFunc) GOF.File.@ref);
```
jeremypw: This does not compile with the new gof-file.vapi.  Can we use Object.ref?   Or add back ref() and unref() to .vapi?

src/View/AbstractDirectoryView.vala
```diff
@@ -750,18 +750,9 @@ namespace FM {
...
-            selection.@foreach ((file) => {
-                list.prepend (file);
-            });
-
-            list.reverse ();
-
-            return list;
+        protected GLib.List<GOF.File>
+        get_selected_files_for_transfer (GLib.List<GOF.File> selection = get_selected_files ()) {
+            return selection.copy_deep ((GLib.CopyFunc) GOF.File.@ref);
```
jeremypw: Does not compile - see other comment

src/View/AbstractDirectoryView.vala
```diff
@@ -2652,15 +2643,13 @@ namespace FM {
...
                     var goffile = GOF.File.get_by_uri (file.get_display_target_uri ());
                     goffile.query_update ();
-                    action_files.prepend (goffile);
+                    action_files.append (goffile);
                 });
-
-                action_files.reverse ();
             } else {
-                action_files = selected_files;
+                action_files = selected_files.copy_deep ((GLib.CopyFunc) GOF.File.@ref);
```
jeremypw: Does not compile - see other comment

libcore/gof-callwhenready.vala
tintou: I'm going to use Object.ref here as GOF.File.ref is just a wrapper (that will disappear when the Vala port will be completed)

tintou:
@jeremypw So after looking at the color plugin I fixed the issue with async access to the files in the color plugin that was triggering the crash.
This plugin needs some structural changes to avoid threading and async issues in the future but that's not for this branch (and the issue is already in master anyway)

tintou:
I had to change the plugin interface to correctly handle the ownership of the given objects.

jeremypw:
The ctags plugin always seem to cause problems ... we should see if there is a lighter weight way of providing this facility.

jeremypw:
Problem with plugin install - the filenames of the ctag plugins installed by meson and cmake are different!

jeremypw:
Meson build produces .so file with different name to the details in the .plug file.

jeremypw:
I'll fix this in master ...

PR #785, opened 2018-11-28
Finish the GOF.File port to Vala

tintou:
Currently no regression found.


jeremypw:
I am not sure about the non-specific name "File.vala"

libcore/File.vala
```diff
@@ -0,0 +1,1177 @@
...
+
+    public signal void changed ();
+    public signal void icon_changed ();
+    public signal void destroy ();
+
+    public bool is_gone;
+    public GLib.File location = null;
+    public GLib.File target_location = null;
+    public GOF.File target_gof = null;
+    public GLib.File directory = null; /* parent directory location */
```
naaando: Can we use a property here returning [GLib.File's get_parent () method](https://valadoc.org/gio-2.0/GLib.File.get_parent.htm)?

libcore/File.vala
```diff
@@ -0,0 +1,1177 @@
+public class GOF.File : GLib.Object {
```
naaando: @jeremypw  Isn't this file expected to have a GPL header?

jeremypw: Yes.

SolarAquarion:
I need this patch on arch, simply adding a .patch to the end of the url doesn't seem to work, can you upload the patch as a text file for me?

libcore/File.vala
jeremypw: Yes, but I think the idea of GOF.File is to speed things up by caching a lot of file info.  GOF.FIle.directory is assigned once on creation.

jeremypw:
@SolarAquarion :  Until this is merged into master it is not approved by elementaryos and is not stable.  You could always fork it and create a patch file yourself.

jeremypw:
After merging master and deleting old plugins before installing, most of the plugins are now working.  Exception is the dropbox plugin which presumably needs recompiling against the new library.

jeremypw:
I can confirm that the dropbox plugin will work with this branch provided that it is amended to use the pantheon-files-core.h rather than gof-file.h and recompiled/reinstalled.

jeremypw:
@tintou: Rather than breaking and fixing non-core plugins (i.e. dropbox) would it be better to generate a separate gof.h file (if that is possible)?

PR #786, opened 2018-11-28
Fix deprecations

jeremypw:
Silences most of the "deprecated" warnings on compiling with the exception of:

* Gtk.Dialog.get_action_area has been deprecated since 3.12 - will be addressed separately
*  BreadcrumbsEntry context menu - will be addressed separately
* Warnings originating from outside Files

### DIFF
--- a/libcore/CMakeLists.txt
+++ b/libcore/CMakeLists.txt
@@ -91,7 +91,7 @@ pkg_check_modules(DEPS REQUIRED
     gthread-2.0
     gio-2.0
     pango>=1.1.2
-    gtk+-3.0>=3.4.0
+    gtk+-3.0>=3.22
     gail-3.0
     gee-0.8
     gmodule-2.0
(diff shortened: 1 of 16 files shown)


cassidyjames:
I think for anything relating to apps, we want to keep the icons in the context menu. See the changes in Code for how we did that in the past, I think this is the relevant PR: https://github.com/elementary/code/pull/369

cassidyjames:
Would it be possible to use a Granite.Dialog to get around the get_action_area deprecations while keeping a consistent look with the rest of the system?

jeremypw:
@cassidyjames : Thanks for the review. So, I should add icons to the other context menus rather than remove them from the BreadcrumbEntry menu? 

I think I'll take the dialog fixes out of this PR and deal with separately if they are to be replaced with Granite Dialogs.

tintou:
can you update the Meson project to require GTK 3.22 (required by `Gtk.Menu.popup_at_pointer ()`)?

jeremypw:
All the Gtk dependencies appeared to be already at >=3.22 except for one cmake file.  Am I missing something?

tintou:
I didn't double-check, so everything is fine :+1:

PR #788, opened 2018-11-29
Icon in "Open with" menus

jeremypw:
Icons are added to the "Open with" submenus in both the pathbar and view.

The simplest (smallest diff) method is used for each.  In the pathbar a grid is added to a GtkMenuItem.  In the view, a GLib.MenuItem is used, linked to the appropriate pre-existing action.

If necessary, GLibMenuItems and actions can be used in the pathbar too.

The unnecessary use of null instead of an empty list in a related function is removed.

### DIFF
--- a/src/Utils/MimeActions.vala
+++ b/src/Utils/MimeActions.vala
@@ -101,10 +101,6 @@ public class Marlin.MimeActions {
             }
         }
 
-        if (result == null) {
-            return null;
-        }
-
         if (!file_has_local_path (file)) {
             filter_non_uri_apps (result);
         }
@@ -138,7 +134,7 @@ public class Marlin.MimeActions {
         return result;
     }
 
-    public static List<AppInfo>? get_applications_for_files (GLib.List<GOF.File> files) {
+    public static List<AppInfo> get_applications_for_files (GLib.List<GOF.File> files) {
         assert (files != null);
         /* Need to make a new list to avoid corrupting the selection */
         GLib.List<unowned GOF.File> sorted_files = null;
@@ -171,21 +167,17 @@ public class Marlin.MimeActions {
             GLib.List<AppInfo> one_result = get_applications_for_file (file);
             one_result.sort (application_compare_by_id);
 
-            if (result != null) {
+            if (result.length () > 0) {
                 result = intersect_application_lists (result, one_result);
             } else {
                 result = (owned) one_result;
             }
 
-            if (result == null) {
-                return null;
+            if (result.length () == 0) {
+                return result;
             }
 
             previous_file = file;
-        }
-
-        if (result == null) {
-            return null;
         }
 
         result.sort (application_compare_by_name);
--- a/src/Utils/MimeActions.vala
+++ b/src/Utils/MimeActions.vala
@@ -167,13 +167,13 @@ public class Marlin.MimeActions {
             GLib.List<AppInfo> one_result = get_applications_for_file (file);
             one_result.sort (application_compare_by_id);
 
-            if (result.length () > 0) {
+            if (result.data != null) {
                 result = intersect_application_lists (result, one_result);
             } else {
                 result = (owned) one_result;
             }
 
-            if (result.length () == 0) {
+            if (result.data == null) {
                 return result;
             }
 
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -2115,7 +2115,7 @@ namespace FM {
                         unowned string label = app_info.get_display_name ();
                         unowned string exec = app_info.get_executable ().split (" ")[0];
                         if (label != last_label || exec != last_exec) {
-                             var menu_item = new GLib.MenuItem (app_info.get_name (), "selection.open_with_app::" + index.to_string ());
+                             var menu_item = new GLib.MenuItem (app_info.get_name (), GLib.Action.print_detailed_name ("selection.open_with_app", new GLib.Variant.int32 (index)));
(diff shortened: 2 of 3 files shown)


src/View/AbstractDirectoryView.vala
```diff
@@ -2103,26 +2102,27 @@ namespace FM {
...
-                                count++;
-                            }
-
-                            last_label = label.dup ();
-                            last_exec = exec.dup ();
+                        /* Ensure no duplicate items */
+                        unowned string label = app_info.get_display_name ();
+                        unowned string exec = app_info.get_executable ().split (" ")[0];
+                        if (label != last_label || exec != last_exec) {
+                             var menu_item = new GLib.MenuItem (app_info.get_name (), "selection.open_with_app::" + index.to_string ());
```
tintou: Just a nice thing that is also possible :
```suggestion
                             var menu_item = new GLib.MenuItem (app_info.get_name (), GLib.Action.print_detailed_name ("selection.open_with_app", new GLib.Variant.int32 (index)));
```

jeremypw:
Thanks for the suggestion!

src/View/AbstractDirectoryView.vala
```diff
@@ -2103,26 +2102,27 @@ namespace FM {
...
+                            menu_item.set_icon (app_info.get_icon ());
+                            apps_section.append_item (menu_item);
+                            count++;
                         }
+
+                        last_label = label.dup ();
```
tintou: Vala will automatically duplicate the string if it is `unowned string` to `string`
```suggestion
                        last_label = label;
```

src/View/AbstractDirectoryView.vala
```diff
@@ -2103,26 +2102,27 @@ namespace FM {
...
+                        last_exec = exec.dup ();
```
tintou:
```suggestion
                        last_exec = exec;
```

src/Utils/MimeActions.vala
```diff
@@ -171,23 +167,19 @@ public class Marlin.MimeActions {
             GLib.List<AppInfo> one_result = get_applications_for_file (file);
             one_result.sort (application_compare_by_id);
 
-            if (result != null) {
+            if (result.length () > 0) {
```
tintou: length iterates the whole list when you can do:
```suggestion
            if (result.data != null) {
```

src/Utils/MimeActions.vala
```diff
@@ -171,23 +167,19 @@ public class Marlin.MimeActions {
...
                 result = intersect_application_lists (result, one_result);
             } else {
                 result = (owned) one_result;
             }
 
-            if (result == null) {
-                return null;
+            if (result.length () == 0) {
```
tintou: length iterates the whole list when you can do:
```suggestion
            if (result.data == null) {
```

src/View/AbstractDirectoryView.vala
```diff
@@ -2103,26 +2102,27 @@ namespace FM {
...
-                if (open_with_apps != null) {
+                if (open_with_apps.length () > 0) {
                     var apps_section = new GLib.Menu ();
                     int index = -1;
                     int count = 0;
                     string last_label = "";
                     string last_exec = "";
 
-                    foreach (var app in open_with_apps) {
+                    foreach (var app_info in open_with_apps) {
```
tintou: GLib.List allows unowned iteration, better use it to not add reference count without any reason ;)
```suggestion
                    foreach (unowned GLib.AppInfo app_info in open_with_apps) {
```

src/View/AbstractDirectoryView.vala
```diff
@@ -2103,26 +2102,27 @@ namespace FM {
 
                 filter_this_app_from_open_with_apps ();
 
-                if (open_with_apps != null) {
+                if (open_with_apps.length () > 0) {
```
tintou:
```suggestion
                if (open_with_apps.data != null) {
```

src/View/Widgets/BreadcrumbsEntry.vala
```diff
@@ -464,22 +464,19 @@ namespace Marlin.View.Chrome {
             foreach (AppInfo app_info in app_info_list) {
                 if (app_info != null && app_info.get_executable () != Environment.get_application_name ()) {
                     at_least_one = true;
-                    var menu_item = new Gtk.ImageMenuItem.with_label (app_info.get_name ());
+                     var item_grid = new Gtk.Grid ();
+                    item_grid.add (new Gtk.Image.from_gicon (app_info.get_icon (), Gtk.IconSize.MENU));
```
tintou: I have the feeling that you'll need to also specify `pixel_size` to not allow the image to be bigger than Gtk.IconSize.MENU (happened in the past with Firefox and LibreOffice, to verify)

src/View/AbstractDirectoryView.vala
jeremypw: OK, thanks. Maybe an earlier Vala version didn't do that?

src/View/AbstractDirectoryView.vala
jeremypw: I'll commit all these changes in one PR

src/Utils/MimeActions.vala
jeremypw: In this case the length is only a handful of apps so I wasn't concerned but I'll make these changes.

tintou:
> @tintou: Regarding pixel_size property of Image, do you need to consider the display scale-factor when setting this?

No it will scale it accordingly

PR #789, opened 2018-11-29
Fix crash when scrolling on mode switch

### LINKED ISSUE #783: Crash when scrolling over view switcher
Not sure if this is what was seen in #6 since interestingly I also discovered in in RTL, but it occurs in LTR as well. Scrolling over the view switcher (i.e. rapidly switching view modes) crashes Files very reliably for me.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/66743291-crash-when-scrolling-over-view-switcher?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #783

* Simplify
* Throttle rate of mode change
* Make ViewContainer and Slot more resilient to rapid mode change.

Attempt to make Window/ViewContainer/Slot/AbstractDirectoryView sufficiently resilient to the extremely rapid mode change signals generated by scrolling were not successful, so the rate of mode change was throttled with a TImeout.

Needs UX opinion on best balance between responsivemess and controllability when clicking/scrolling on mode switch.


tintou:
Just thinking out loud: Instead of a timeout, don't you think that a GLib.Idle would be safer?

jeremypw:
I'll try it.

jeremypw:
@tintou:   An Idle does work provided it is given a low priority.  However, the downside is that because scroll produces very rapid events, the view changes rapidly so it is jerky and difficult to control which view you end up on.  The timeout seems more controllable.

There was an error in the PR anyway - the timeout_id variable was not being assigned.

jeremypw:
I think the timeout is safer in terms of stability as it is much longer than the time to create a new slot (loading the files happens asynchronously and is cancelled when the slot mode changes).

jeremypw:
@tintou  Is this OK now?

PR #792, opened 2018-12-02
Fix open with app context menu

### LINKED ISSUE #791: Segmentation fault on context menu open
When I try to open context menu of any file, I faced with `segmentation fault`.
Also, when I try to do `Open in -> Terminal` on any directory, nothing happen.

Looks like the problem in this commit https://github.com/elementary/files/commit/64a8bbe5810c4c734ce4106908f08c83c8e2c5ea

jeremypw:
Fixes #791 

* Construct detailed name properly; 
* Use int as target for open_with_app action

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -54,7 +54,7 @@ namespace FM {
         /* Menu Handling */
         const GLib.ActionEntry [] selection_entries = {
             {"open", on_selection_action_open_executable},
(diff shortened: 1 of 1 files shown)


tintou:
Oops didn't saw that, I also have #798 that fixes other names and uses uint instead of int for the open-with-app (as it seems that it's what is used afterwards)

jeremypw:
Closed in favour of #798 which is more far-reaching.

PR #794, opened 2018-12-02
Search action bar with button to increase search scope

### LINKED ISSUE #741: Search is limited to 10 results
As a proof of concept, I created 11 subdirectories under `$HOME/some/subdir` and touched a .foo and .bar file in each of them:
screenshot from 2018-11-07 16 07 47 2x

After enabling hidden files I then pressed CTRL+F to search for .foo files, but files just shows the first 10 results:

cropped screenshot from 2018-11-07 15 58 16 2x

My real world use case was to search for files with a certain pattern in my 11GB `src` directory, and even if that might be something that doesn't apply to most users I still find it probable that many users at least have more than 10 files that matches a certain pattern.

To sum it up I think that files needs a better search function that can list all results, preferably also with filters for type, size, modification/access time, etc.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65498445-search-is-limited-to-10-results?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #741 

Adds an actionbar and button to the search result popup window.
Clicking the button increases the maximum results and maximum search depth incrementally and restarts the search with the current search term.
The scope is reset when the search term is changed.

The actionbar could be used to add further search functionality in future.

#804 gives the same function through clicking on an ellipsis in the search view and could be merged instead of or in addition to this.

### DIFF
--- a/libwidgets/View/SearchResults.vala
+++ b/libwidgets/View/SearchResults.vala
@@ -114,6 +114,8 @@ namespace Marlin.View.Chrome {
 
         int current_count;
         int deep_count;
+        int max_results = MAX_RESULTS;
+        int max_depth = MAX_DEPTH;
 
         bool local_search_finished = false;
         bool global_search_finished = false;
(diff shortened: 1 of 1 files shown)


jeremypw:
Closing for now as I think the simpler solutions I pushed are more likely to find favor.

PR #797, opened 2018-12-02
Fix open with menu

Mohelm97:
#788 Uses GLib.Action.print_detailed_name instead of string, but GLib.Action.print_detailed_name doesn't work for action groups and returns null, so the whole "open with" menu isn't working.


tintou:
I proposed a fix to the cause: https://github.com/elementary/files/pull/798
`print_detailed_name` works fine with action groups, it's just that the name of the action wasn't valid.

PR #798, opened 2018-12-03
Fix GLib.Action naming

tintou:
It seems that action_name is valid if it consists only of alphanumeric characters, plus '-' and '.'. (according to https://developer.gnome.org/gio/stable/GAction.html#g-action-name-is-valid )
This commit makes all action name valid. Some change has been done to simplify `open-with-app`.
Supersedes #797

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -54,11 +54,11 @@ namespace FM {
         /* Menu Handling */
         const GLib.ActionEntry [] selection_entries = {
             {"open", on_selection_action_open_executable},
-            {"open_with_app", on_selection_action_open_with_app, "s"},
-            {"open_with_default", on_selection_action_open_with_default},
-            {"open_with_other_app", on_selection_action_open_with_other_app},
+            {"open-with-app", on_selection_action_open_with_app, "u"},
+            {"open-with-default", on_selection_action_open_with_default},
+            {"open-with-other-app", on_selection_action_open_with_other_app},
             {"rename", on_selection_action_rename},
-            {"view_in_location", on_selection_action_view_in_location},
+            {"view-in-location", on_selection_action_view_in_location},
             {"forget", on_selection_action_forget},
             {"cut", on_selection_action_cut},
             {"trash", on_selection_action_trash},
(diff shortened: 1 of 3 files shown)


src/View/AbstractDirectoryView.vala
```diff
@@ -2104,35 +2103,33 @@ namespace FM {
...
                         unowned string exec = app_info.get_executable ().split (" ")[0];
                         if (label != last_label || exec != last_exec) {
-                             var menu_item = new GLib.MenuItem (app_info.get_name (), GLib.Action.print_detailed_name ("selection.open_with_app", new GLib.Variant.int32 (index)));
+                             var menu_item = new GLib.MenuItem (label, GLib.Action.print_detailed_name ("selection.open-with-app", new GLib.Variant.uint32 (count)));
                             menu_item.set_icon (app_info.get_icon ());
                             apps_section.append_item (menu_item);
-                            count++;
                         }
 
                         last_label = label;
```
jeremypw: If `last_label` and `last_exec` are unowned (not copied) then wont they just update when `label` and `exec` change so will never differ?

jeremypw:
Withdrawn in favour of #798 which is more far-reaching.

jeremypw:
Oops - wrong one!

src/View/AbstractDirectoryView.vala
jeremypw: OK, I though last_label would just point to label and therefore return whatever label was pointing at (which changes).

Mohelm97: I think @jeremypw is right, I don't understand that much from it but here's my note:
by adding 
debug(exec+":"+last_exec+":"+(exec != last_exec).to_string()); before the if 
I got this?
** (io.elementary.files:30799): DEBUG: 16:55:29.369: AbstractDirectoryView.vala:2115: \x80\xe7eJ\xdeU:Files:true
(\x80\xe7eJ\xdeU) is not right, in the next print of last_exec showes that it should be something else.

Mohelm97: Ohhh, I got it XD, The problem is with the split it, if you put it in unowned string you will lose the value the next line, so at least exec and last_exec should be strings

jeremypw: On reflection I think you are right.  Sorry I misunderstood at first. As this is now merged a separate PR can fix this (if @tintou agrees).

tintou: Yeah a separate PR can fix it, but I wonder why it is necessary here at all when the filtering could have been done ahead of time when filling the appinfo list.
Notice that the current deduplication method here didn't even ensure that there were no duplicates (if 2 applications are not following each others, they won't be deduplicated)

jeremypw: I agree - a separate PR should revisit the deduplication problem (although there never has been a reported instance of duplicate names not following each other because MimeActions already sorts the returned list).  I am not sure under what conditions you could get duplicate execs with different names (maybe different versions co-installed?) - these are intended to be included in the menu ( is this correct behaviour?) If they have the same name they would be filtered out anyway.

@tintou Would you like to propose a better solution?

PR #806, opened 2018-12-11
Sync pointer and keyboard focus

### LINKED ISSUE #801: Keyboard focus and pointer focus can differ
Both keyboard focus and pointer focus are indicated by highlighting (see #322). When the keyboard focus is on an unselected item there can apparently be two different focussed items.

One way to ameliorate this is for the two focuses to be synchronised.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/67176917-keyboard-focus-and-pointer-focus-can-differ?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #801 

Subject to UX approval.


danrabbit:
Mentioning @elementary/ux-developers 

I'm not sure this is a good idea. I don't think any other app does this. Usually we end up defying users' expectations by messing with how Gtk+ handles focus by default

jeremypw:
OK, I'll abandon this then.

PR #811, opened 2018-12-13
Add tests to limited length path

naaando:


jeremypw:
I am working on this.

jeremypw:
Should now pass tests and some additional tests added.

Returned string from limited_length_string function should not exceed specified length by more than 1.

PR #816, opened 2018-12-17
Always show search icon in pathbar

### LINKED ISSUE #813: No search button
I know we closed #170 with https://github.com/elementary/files/pull/763 but I think there is still room for improvement. Like finding a subtle way to include a search button. 

I personnaly like the idea of adding another action icon. The tooltip for this button could be a a place to show the Ctrl + F keyboard shortcut. It could even be a place to explain that you can simply start typing to search. 
search indicator

It's a UI approach you can also see in Firefox for example:
screenshot from 2018-12-16 14 10 18

@jeremypw had a good point on this though:
> The native Gtk.Entry widget that the pathbar is based on can only show one action icon at the right so it would take some refactoring.

jeremypw:
Fixes #813 

A possible solution to requests to make search capability more discoverable.  Breadcrumbs are drawn so as not to obscure the primary icon and the primary is always shown.  Also the primary icon is made activatable and given a tooltip.

screenshot from 2018-12-17 19 30 26


cassidyjames:
I'm :-1: for this design. I'm not sure how to articulate exactly why, but it feels off. Perhaps it is because the search then looks like it is only part of the first breadcrumb?

jeremypw:
Closing as this solution does not gain the approval of the UX team.

PR #823, opened 2018-12-21
Honor global history setting when restoring and saving tabs

### LINKED ISSUE #822: Restore tabs should honor global privacy setting
See https://github.com/elementary/code/issues/473

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/67842799-restore-tabs-should-honor-global-privacy-setting?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
#Fixes #822

* Do not restore tabs when privacy/history setting off


danrabbit:
I don’t think we want to clear any existing history when history is turned off. Just not record any new data.

There is a dialog for clearing any existing history that is separate from just turning off history for now

jeremypw:
I thought that turning off history *did* clear history as well - that seems to be the case for the recent:// folder at least.  I need to check the source, but this button should be insensitive when there is no data to clear (but isnt always).  It would be difficult for the plug to clear history from various application settings ....

jeremypw:
I cannot see anything in the code that deletes history when it is switched off so perhaps it is some thing that happens at system level when the gnome privacy is turned on?

I'll revert the clearing of settings for now. It can be added later if desired.

danrabbit:
@jeremypw Yeah that is a massive and devastating bug :( It didn't used to do that, I don't think.

Is it possible to use Zeitgeist (or some other central history authority) for remembering tabs so that we can have an easy history API to clear?

jeremypw:
Possibly, but that is for another PR - all elementary would have to use that API.

jeremypw:
@danrabbit Is this OK now?

4jNsY6fCVqZv:
I am very happy that you are working on an integration! Hope the PR can be merged soon.

jeremypw:
@4jNsY6fCVqZv  Patience!

4jNsY6fCVqZv:
@jeremypw Oh, excuse me! I try with patience. I don't know your workflows well either. I only saw that your PR is already more than nine months old and despite improvements since then without further public review. That's why I thought something was missing to merge it.

src/View/Window.vala
```diff
@@ -892,31 +891,32 @@ namespace Marlin.View {
 
         private void save_tabs () {
             VariantBuilder vb = new VariantBuilder (new VariantType ("a(uss)"));
```
jlnr: Micro-optimization: this variable can go into the `if` block.

Since the whole method is a no-op without the `remember_history` setting, an early-exit `return` would keep the indentation flat, as in the `restore_tabs` method.

jeremypw:
@4jNsY6fCVqZv There are currently 150+ outstanding elementaryos PRs for review and only a handful of reviewers so  it can take an undefined amount of time before someone has time/interest to look at a particular one, if it is not critical.  Files code is relatively hard to review as well.

jeremypw:
I have decided to revert dropping the restore-tabs setting as it is not strictly essential to fix the stated issue.  It is possible that a user would want Files to never restore tabs even if the global history is on.  So I think removing this requires a separate issue and design decision.

4jNsY6fCVqZv:
@jeremypw Is it somehow possible to filter elementary GitHub project for open PRs?

4jNsY6fCVqZv:
@jeremypw @jlnr Thank you for your work & time!
Asked out of curiosity: Is this the same PR code that will allow privacy in Code as well?
https://github.com/elementary/code/pull/612

jeremypw:
@4jNsY6fCVqZv The purpose is the same but the implementation is different; Code has its own way of saving and restoring tabs.  Ideally of course there should be shared code, maybe in Granite, but the apps were developed separately.

4jNsY6fCVqZv:
@jeremypw Do you think a system-wide solution would be useful and feasible?

jeremypw:
@4jNsY6fCVqZv On further thought, only a limited amount of code could be shared as different apps save different information about the tabs.  e.g. Files saves what kind of view each tab contains which would be meaningless to Code.  Code might want to save the cursor position which would be meaningless to Files.

4jNsY6fCVqZv:
I understand that very well, thank you! 
I also find it difficult to imagine how the different approaches can be meaningfully integrated into a superordinate schema, which makes it easier for all apps system-wide at the same time.
Then I hope that there will be people who can continue to work on https://github.com/elementary/code/pull/612 and its review.

PR #824, opened 2018-12-27
Use Gtk.TreeStore as base model for views.

jeremypw:
This builds on the work of tintou in porting fm_list_model to Vala.  This seeks to further reduce the code burden by using Gtk.TreeStore as a basis for ListModel  (renamed to DirectoryModel).

As this does not re-implement the reverse lookup maps in the custom model, there may be a performance hit for large folders for some operations, but this has not yet been quantified.  Some steps have been taken to minimise the occasions when a lookup is required.

The model itself now has a single column holding a GOF.File object. Cell data functions are used to extract the appropriate property to display when not using custom renderers.

Display of color tags is simplified.

A 'null' GOF.File is introduced to allow empty rows to be displayed.

The model class is thereby reduced to under 300 lines.


jeremypw:
I can confirm that, once you get into the 1000's of files in the model it takes considerably longer to generate the view because adding a new file to the model takes longer and longer.  I'll see if I can get round this.

jeremypw:
Last commit reduced difference in initial display time to about 10% longer  which is not really significant in practice.

libcore/ListModel.vala
```diff
@@ -0,0 +1,323 @@
+/* Copyright (c) 2018 elementary LLC (https://elementary.io)
```
tintou:
```suggestion
/* Copyright 2018-2019 elementary, Inc (https://elementary.io)
```

libcore/ListModel.vala
```diff
@@ -0,0 +1,323 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301, USA.
+ */
```
tintou: Don't forget to add you as an author ;)

libcore/ListModel.vala
```diff
@@ -0,0 +1,323 @@
...
+
+/* Uses Gtk.TreeStore instead of custom model to reduce code base to support */
+/* Assumptions:
+ *      User will not add duplicate files
+ *      Traversing model to find/remove file sets will be fast enough for general use
+ * If these assumptions prove false, then will have to add mechanisms for mapping files to rows and detecting duplicates
+ */
+namespace FM {
+public interface DirectoryViewInterface : Object {
```
tintou:
```suggestion
public interface FM.DirectoryViewInterface : Object {
```
It's as clean as before :)

libcore/ListModel.vala
```diff
@@ -0,0 +1,323 @@
...
+
+    /* 'reversed' indicates whether the sort should be the natural order for that property
+     * (defined by the sort function in gof.file) or not */
+    public abstract bool get_order (out FM.ColumnID sort_file_property, out bool reversed);
+    public abstract void set_order (FM.ColumnID sort_file_property, bool? reversed = null);
+}
+
+//    public signal void subdirectory_unloaded (GOF.Directory.Async directory);
+
+public class DirectoryModel : Gtk.TreeStore, DirectoryViewInterface {
```
tintou:
```suggestion
public class FM.DirectoryModel : Gtk.TreeStore, FM.DirectoryViewInterface {
```

libcore/ListModel.vala
```diff
@@ -0,0 +1,323 @@
...
+
+    private int file_match_func (GOF.File? a, GOF.File? b) {
+        if (a == null || b == null) {
+            return 0;
+        }
+
+        return a.compare_for_sort (b, (int)sort_file_property, sort_directories_first, reversed);
+    }
+
+    public new bool get_sort_column_id (out int sort_col, out Gtk.SortType sort_type) {
```
tintou: I advocate to never use the `new` keyword, either `override` the previous method or use a different name to be sure that the right function will be called.

libcore/ListModel.vala
```diff
@@ -0,0 +1,323 @@
...
+        /* We do not want the normal method to be called externally */
+        /* Externally, 'get_order ()' should be called */
+        sort_col = 0;
+        sort_type = 0;
+        assert (false);
+        return true;
+    }
+
+    public new void set_sort_column_id (int sort_col, Gtk.SortType sort_type) {
```
tintou: As above

libcore/ListModel.vala
```diff
@@ -0,0 +1,323 @@
...
+
+    public void unset_order () {
+        unsorted = true;
+    }
+
+    public GOF.File? file_for_path (Gtk.TreePath path) {
+        GOF.File? file = null;
+        Gtk.TreeIter? iter;
+        if (get_iter (out iter, path) && iter != null) {
+            @get (iter, ColumnID.FILE_COLUMN, out file);
```
tintou:
```suggestion
            return file_for_iter (iter);
```

libcore/ListModel.vala
```diff
@@ -0,0 +1,323 @@
...
+                }
+
+                return false;
+            });
+
+            seq_iter = seq_iter.next (); /* file was not in model */
+        }
+
+        bool valid = false;
+        foreach (Gtk.TreeRowReference row_ref in rows_to_remove) {
```
tintou: I think that this might be better
```suggestion
        foreach (unowned Gtk.TreeRowReference row_ref in rows_to_remove) {
```

libcore/ListModel.vala
```diff
@@ -0,0 +1,323 @@
...
+                return false;
+            });
+
+            seq_iter = seq_iter.next (); /* file was not in model */
+        }
+
+        return (owned)rows_found;
+    }
+
+    public new void get_value (Gtk.TreeIter iter, int column, out Value return_value) {
```
tintou: Avoid the `new` keyword

libcore/gof-file.c
```diff
@@ -1354,7 +1384,11 @@ gof_file_get (GFile *location)
     GOFFile *file = NULL;
     GOFDirectoryAsync *dir = NULL;
 
-    g_return_val_if_fail (location != NULL && G_IS_FILE (location), NULL);
+    //g_return_val_if_fail (location != NULL && G_IS_FILE (location), NULL);
+
+    if (location == NULL) {
+
```
tintou: something is missing here I guess

src/View/ListView.vala
```diff
@@ -18,64 +18,119 @@
...
+
+            col.clickable = true;
+            col.clicked.connect (on_column_clicked);
+            col.sort_indicator = false;
+            col.set_resizable (false);
+            col.set_expand (false);
+            col.min_width = 24;
+
+            if (id == FM.ColumnID.SIZE || id == FM.ColumnID.MODIFIED) {
+                renderer.@set ("xalign", 1.0f);
```
tintou:
```suggestion
                renderer.xalign = 1.0f;
```

src/View/ListView.vala
```diff
@@ -18,64 +18,119 @@
...
+            } else {
+                renderer.@set ("xalign", 0.0f);
```
tintou:
```suggestion
                renderer.xalign = 0.0f;
```

libcore/ListModel.vala
jeremypw: I am not sure it is necessary to override this method anyway as the model now only has one column.

libcore/gof-file.c
jeremypw: Not sure what I was going to do here ... I think return a GOF.File.get_null () rather than NULL.

src/View/ListView.vala
jeremypw: Done

src/View/ListView.vala
jeremypw: Done

libcore/ListModel.vala
jeremypw: Done

libcore/ListModel.vala
jeremypw: Done

libcore/ListModel.vala
jeremypw: Done

libcore/ListModel.vala
jeremypw: Done

libcore/ListModel.vala
jeremypw: Done

jeremypw:
Updated to compile and run with Vala port of GOF.File.  Due to large number of conflicts fixed may be some regressions.

jeremypw:
Note: Vala model needed to faciltate implementation of new, efficient IconView.

jeremypw:
Closing as the chance of this being reviewed is minimal.

PR #838, opened 2018-12-31
Show correct default app for e.g. text files

### LINKED ISSUE #836: Files does not open files with a default application
## Expected Behavior

Files remember default applications and open some kind of files with that application.
In the following example, text files should be opened with VS Code, which is selected as a default application to open text files.

## Actual Behavior

Files does not remember default applications or open some kind of files with that application.
In the following example, text files are opened with Code, which is not selected as a default application to open text files.

## Steps to Reproduce the Behavior

1. Install Visual Studio Code (VS Code)

virtualbox_elementary os_30_12_2018_15_07_04

2. Set VS Code as a default text editor in System Settings
3. Create a text file in a directory and click it to open

virtualbox_elementary os_30_12_2018_15_07_23

4. The file is opened with Code, although you set VS Code as a default text editor.
You may see Code stays at the top of the context menu, and VS Code is in the "Open with" menu

virtualbox_elementary os_30_12_2018_15_07_36

5. Select "Open with" in the context menu→"Other Application…" and try to set VS Code as a default application to open text files in the dialog above. Even after I restart Files or the system, text files is not opened with VS Code, and instead opened with Code

## Note

* This issue happens with Files 4.1.x. I didn't face this issue with Files 4.0
* This issue also happens with other kind of files e.g. xlsx files

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/68112851-files-does-not-open-files-with-a-default-application?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #836 

A slight change when porting  GOF.File.get_default_handler () to Vala resulted in default apps being required to support uris unnecessarily thereby excluding some that should have been shown.


naaando:
This lets us open a file with the default application set on Switchboard, but once you set another option as default inside the 'Open with' menu it doesn't update on Switchboard even after restarting it, this is still broken here or might be an issue in Switchboard?

jeremypw:
OK, so what is happening is that the Switchboard default app plug, when first opened, shows the default for a specific type of file e.g. "image/jpeg" for photos.  So if you use Files to change the default for "image/png" files the plug will not show this, even after closing and reopening.  However, if you use the plug the change the default for image files then it changes it for a whole range of image types so Files reflects that.  If you want to see Files change the Switchboard default plug then you have to use Files to change the default for "image/jpeg" files for example. Note that you have to close the whole Switchboard app not just the plug for changes to be reflected because the plugs are cached and there is no monitoring of external changes to the defaults.

This raises the question of whether Files is too specific when it sets the default app for a type? Should it copy Switchboard and change the default for all image files if it is changed for one?

jeremypw:
Or maybe the appchooser dialog should show an option to set default for the whole class of content types when appropriate.

jeremypw:
I would suggest this PR is merged for now (if there are no other problems) so at least Files is internally consistent.  Then we can decide whether we want to change the intended behaviour or interface.

PR #857, opened 2019-01-10
Fix crash when opening multiple video files

### LINKED ISSUE #856: Activating multiple (video) files causes crash
See https://elementaryos.stackexchange.com/questions/17868/closes-files-after-selecting-multiple-videos

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/68788840-activating-multiple-video-files-causes-crash?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #856 

Should not use an unowned list of files inside an Idle loop as it may not remain valid.


src/View/AbstractDirectoryView.vala
```diff
@@ -558,8 +558,10 @@ namespace FM {
                         }
                     }
                 } else if (default_app != null) {
+                    var files_to_open = get_files_for_action ();
```
tintou:
```suggestion
                    /* Because this is in another thread we need to copy the selection to ensure it remains valid */
                    var files_to_open = selection.copy_full (GLib.Object.ref);
```
It is probably better like this so we are logically using the exact same list

jeremypw: Do you mean this?
```
var files_to_open = selection.copy_deep ((GLib.CopyFunc)(GLib.Object.ref));
```

jeremypw: I have followed your suggestion (using copy_deep)

PR #874, opened 2019-01-28
Remove soversion argument for core and widget libraries

### LINKED ISSUE #837: meson: shared libraries installed twice, with different versions (since 4.1.0)
Since commit https://github.com/elementary/files/commit/b2a92948a56e61276cfc43d7fcd1fb3c4874edc3 (the 4.1.0 Release commit, which also fixed the meson project version to match the actual version), there are now *two* shared libraries with different versions installed simultaneously:

- `/usr/lib64/libpantheon-files-core.so.0` and `/usr/lib64/libpantheon-files-widgets.so.0`, and
- `/usr/lib64/libpantheon-files-core.so.4.1.1` and `/usr/lib64/libpantheon-files-widgets.so.4.1.1`

It looks like those files are installed (or even compiled) twice somewhere in the meson definitions, with different versioning rules attached to them :interrobang: 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/68117318-meson-shared-libraries-installed-twice-with-different-versions-since-4-1-0?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #837 
following  tintou 's advice.

### DIFF
--- a/libcore/meson.build
+++ b/libcore/meson.build
@@ -70,7 +70,6 @@ pantheon_files_core_library = shared_library(
     install: true,
     install_dir: [true, join_paths(get_option('prefix'), get_option('includedir'), pantheon_files_core_name), true],
     version: meson.project_version(),
-    soversion: '0',
 )
 
 install_headers(
--- a/libwidgets/meson.build
+++ b/libwidgets/meson.build
@@ -30,7 +30,6 @@ pantheon_files_widgets_library = shared_library(
     install: true,
     install_dir: [true, true, true],
     version: meson.project_version(),
-    soversion: '0',
 )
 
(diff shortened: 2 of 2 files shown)


tintou:
Giving that we don't have any out-of-tree plugins right now that should be alright, for people using the FIles libraries, please ensure that you're linking against the right one…

PR #889, opened 2019-02-14
Add libwidgetsgrid to source tree and compile

jeremypw:
Start trying to use libwidgetgrid to replace Gtk.IconView in io.elementary.files


jeremypw:
wrong target

Issue #885, opened 2019-02-13
Make better use of the address bar space


sprite-1:
![image](https://user-images.githubusercontent.com/33685433/52705209-51403e80-2fbd-11e9-8020-4a24d1ccc726.png)

As can be seen on the above screenshot, the address bar is wasting quite a lot of white space while truncating the rest of the path. I think it should stretch all the way to the end or just leave a small gap at the end so the user can click that to edit the path just like before.

jeremypw:
There does not seem to be any reason not to reduce the space at the right.

jeremypw:
Fixed by #896

PR #896, opened 2019-02-25
Reduces gap at the right of the Breadcrumbs element

MelD06:
Fixes issue #885 
The space has shrunk quite a bit, should this be an UX discussion ?
screenshot from 2019-02-26 00 34 41

### DIFF
--- a/libwidgets/Chrome/BasicBreadcrumbsEntry.vala
+++ b/libwidgets/Chrome/BasicBreadcrumbsEntry.vala
@@ -40,10 +40,10 @@ namespace Marlin.View.Chrome {
         }
 
         public bool hide_breadcrumbs { get; set; default = false; }
-        public const double MINIMUM_LOCATION_BAR_ENTRY_WIDTH = 36;
+        public const double MINIMUM_LOCATION_BAR_ENTRY_WIDTH = 10;
         public const double MINIMUM_BREADCRUMB_WIDTH = 12;
         public const double COMPLETION_ALPHA = 0.5;
-        public const int ICON_WIDTH = 48;
+        public const int ICON_WIDTH = 6;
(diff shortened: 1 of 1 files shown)


libwidgets/Chrome/BasicBreadcrumbsEntry.vala
```diff
@@ -40,10 +40,10 @@ namespace Marlin.View.Chrome {
         }
 
         public bool hide_breadcrumbs { get; set; default = false; }
-        public const double MINIMUM_LOCATION_BAR_ENTRY_WIDTH = 36;
+        public const double MINIMUM_LOCATION_BAR_ENTRY_WIDTH = 10;
         public const double MINIMUM_BREADCRUMB_WIDTH = 12;
         public const double COMPLETION_ALPHA = 0.5;
-        public const int ICON_WIDTH = 48;
+        public const int ICON_WIDTH = 6;
```
jeremypw: I am not sure the ICON_WIDTH constant is needed (any more) but while it is there it would be better to set it to a valid IconSize - the smallest available is 16px.  You can reduce the minimum entry width to keep the total to say 24px.  The breadcrumbs entry is in need of refactoring anyway and I'll look into whether there are any regressions if ICON_WIDTH is removed.

MelD06: After setting ICON_WIDTH to 0, I noticed that it needs to be set to a proper value to ensure that this doesn't happen while resizing the window :
screenshot from 2019-02-26 19 33 59
After fiddling around with both values, breadcrumbs bar behaves like this on last commit and access to path on the tightest resize is maintained. I may have overlooked this on previous commit...
screenshot from 2019-02-26 19 38 32
screenshot from 2019-02-26 19 38 44

PR #914, opened 2019-03-28
Put Files CSD on a diet

quequotion:
This allows Files CSD to match thickness with Terminal and Epiphany (from which I borrowed those [margin specs](https://github.com/elementary/os-patches/blob/dbc305d506a3fae94395513bf0b038a3517ea50b/src/ephy-header-bar.c#L733#L734)).

CSD margins should be handled by the css anyway, as discussed in [Terminal Issue #270](https://github.com/elementary/terminal/issues/270).

[Matching CSD](https://postimg.cc/1n58LTtC)

### DIFF
--- a/libwidgets/Chrome/BasicLocationBar.vala
+++ b/libwidgets/Chrome/BasicLocationBar.vala
@@ -58,9 +58,8 @@ namespace Marlin.View.Chrome {
         }
 
         construct {
-            margin_top = 4;
-            margin_bottom = 4;
-            margin_start = 3;
+            margin_start = 24;
+            margin_end = 6;
         }
 
         public BasicLocationBar (Navigatable? _bread = null) {


quequotion:
Speaking of Epiphany, looks like upstream has a [completely new way of doing CSD margins coming up](https://github.com/GNOME/epiphany/commit/81b06827ff8908d0be72cc56162e412d47cbde46#diff-b97eb8d334dc9100ce1b3e3c985027fb).

In any case, it's nice having the CSD the same size in as many applications as possible; matching chrome in different applications improves general desktop usability (less disorientation when switching from one application to another).

quequotion:
Not all that different! I like how the proposed margins line up the [back]/[forward], [close] and [(un)maximize] buttons; putting these controls in exactly the same place across the two apps.

The [close] and [(un)maximize] buttons in Terminal are spaced out a few pixels wider. Is that center alignment? If that's the standard spacing in elementary apps, then we should set it for epiphany as well (and yes, definitely nix the top and bottom margins in Files).

danrabbit:
I'm going to close this since #915 was merged

quequotion:
Yeah, probably good; just hadn't tested yet.

PR #918, opened 2019-04-04
Shift home end selection

### LINKED ISSUE #917: Selecting files with SHIFT + ( Pos1 / END) broken
Massively using file manager for handling and moving around large batches of folders, I experienced odd behaviours when selecting multiple files using keyboard and SHIFT.

What does happen:

Step: Select an arbitrary file and press SHIFT+Pos1
Expectation: Select this file any everything "from there" to the first item in the current folder view.
Outcome: Select all items in the first column of the current window

Step: Select an arbitrary file and press SHIFT+END
Expectation: Select this file and everything "from there" to the last item in the current view.
Outcome: Jumps to the last item and unselects everything that has been selected before. 

That doesn't feel particularly concise or meaningful... ;)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/72348818-selecting-files-with-shift-pos1-end-broken?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #917


jeremypw:
@donadigo I think it behaves reasonably under most sequential selection circumstances.  I am not sure how many hoops it is worth jumping through to deal with unusual sequences of actions.  I have had to introduce another variable to remember the last selected file as `get_selected_files ()` in IconView is not necessarily in temporal order.

PR #920, opened 2019-04-14
Implement org.freedesktop.portal.FileChooser

donadigo:
Implements the standard DBus interface `org.freedesktop.portal.FileChooser` by re-targeting the Files code to be a shared library that can be used by both the main app and the file chooser.

You can test this with building this branch. Then launching `io.elementary.files-portal` and in another terminal `GTK_USE_PORTAL=1 app`.

There are currently some problems with Gtk not properly subscribing to the response signal so showing the dialog will make the app freeze.

TODO:
- [x] Implement the skeleton of  `org.freedesktop.portal.FileChooser`
- [x] Make Files code a shared library
- [ ] Support all / most variant options for the dialog
- [ ] Strip the dialog out of Files specific features (open in tab, open in app etc.)

filechooser

### DIFF
--- a/filechooser-portal/FileChooser.vala
+++ b/filechooser-portal/FileChooser.vala
@@ -1,0 +1,145 @@
+/*-
+ * Copyright (c) 2019 elementary, Inc. (https://elementary.io)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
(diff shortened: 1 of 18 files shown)


donadigo:
Looks like I misunderstood the way portal implementation happens. I tried to implement the portal handler instead of the `impl` interface. I will do another PR to https://github.com/elementary/xdg-desktop-portal-pantheon. Closing.

donadigo:
Moved to https://github.com/elementary/xdg-desktop-portal-pantheon/pull/2.

PR #922, opened 2019-04-16
Make Files a shared library

donadigo:
This is a draft meant to adapt Files codebase to be a shared library much like plank's. However, this is not in anyway a final "design" of this branch and is only meant to succesfully compile https://github.com/elementary/xdg-desktop-portal-pantheon/pull/2.


jeremypw:
I guess this would require rewriting all other PRs once merged ?

donadigo:
@jeremypw I guess for some PR's that would be true, I'm also not sure if this is really the way we want to go here.

jeremypw:
Its not necessarily a problem, but ideally the library should be as up to date as possible i.e. derived from a master with few outstanding PRs.  Is there any other way to implement NativeFileChooser?

donadigo:
@jeremypw the other way is that we put all the widgets we need in the libwidgets library but I believe that this will essentially mean, that almost all widgets will have to be put there because of their inheritance and access to other widgets.

jeremypw:
That may be more straightforward.  Is there any other future use for a complete Files library?

donadigo:
@jeremypw I don't know, I think @Alecaddd said that it would be probably nice to use list view in some apps?

danrabbit:
I'm gonna close this since this filechooser portal is now part of Files and it's not clear that this has other useful applications

PR #925, opened 2019-04-18
Get placeholder text color from style context if possible.

### LINKED ISSUE #923: Placeholder/completion text always black, regardless of stylesheet
This is a hardcoded magic color:

https://github.com/elementary/files/blob/7888a730827596d70115732d555d52e79918813b/libwidgets/Chrome/BasicBreadcrumbsEntry.vala#L718

This causes issues in other stylesheets plus if we were to theoretically support a dark style ever. Instead, it should probably do something like what the search popover does:

https://github.com/elementary/files/blob/36c122b652e44f9d3a2aa128c5fab5bc593a94d6/libwidgets/View/SearchResults.vala#L1109-L1125

…which appears to work correctly with various stylesheets and when using a dark style.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/73009138-placeholder-completion-text-always-black-regardless-of-stylesheet?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #923

### DIFF
--- a/libwidgets/Chrome/BasicBreadcrumbsEntry.vala
+++ b/libwidgets/Chrome/BasicBreadcrumbsEntry.vala
@@ -42,7 +42,6 @@ namespace Marlin.View.Chrome {
         public bool hide_breadcrumbs { get; set; default = false; }
         public const double MINIMUM_LOCATION_BAR_ENTRY_WIDTH = 16;
         public const double MINIMUM_BREADCRUMB_WIDTH = 12;
-        public const double COMPLETION_ALPHA = 0.5;
         public const int ICON_WIDTH = 32;
         protected string placeholder = ""; /*Note: This is not the same as the Gtk.Entry placeholder_text */
         protected BreadcrumbElement? clicked_element = null;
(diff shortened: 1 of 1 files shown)


cassidyjames:
@jeremypw this works in the elementary stylesheet, but for some reason is still black in Adwaita. I'm not extremely concerned about it, but any idea why that'd be?

jeremypw:
Maybe "placeholder_text_color" style is not defined in Adwaita?  The code would then fallback to the existing behaviour. Is there a general way of determining whether a dark theme is in use?

cassidyjames:
@jeremypw I don't think reliably, especially because some light themes still have dark headerbars. Would it be possible to just inherit the normal entry text color for that context and reduce opacity? That might not be ideal, but it could work. I'm still happy with this commit because it does help fix things for elementary OS, but I'm not sure what would be a more general fix.

jeremypw:
@cassidyjames OK, that sounds like an idea.  I'll give it a go.

PR #933, opened 2019-04-29
Simplify generation of tab name, ensuring unescaped.

### LINKED ISSUE #930: Spaces in Samba share folder names show as "%20" in tabs
image

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/73388681-spaces-in-samba-share-folder-names-show-as-20-in-tabs?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #930

### DIFF
--- a/src/View/ViewContainer.vala
+++ b/src/View/ViewContainer.vala
@@ -365,33 +365,21 @@ namespace Marlin.View {
 
        private void update_tab_name () {
             string? slot_path = Uri.unescape_string (this.uri);
-            string? tab_name = null;
+            string tab_name = Marlin.INVALID_TAB_NAME;
 
             if (slot_path != null) {
-                if (this.location.get_path () == null) {
-                    tab_name = Marlin.protocol_to_name (this.uri);
+                string protocol, path;
+                PF.FileUtils.split_protocol_from_path (slot_path, out protocol, out path);
+                if (path == "" || path == Path.DIR_SEPARATOR_S) {
+                    tab_name = Marlin.protocol_to_name (protocol);
+                } else if (protocol == "" && path == Environment.get_home_dir ()) {
+                    tab_name = _("Home");
                 } else {
-                    try {
-                        var fn = Filename.from_uri (slot_path);
-                        if (fn == Environment.get_home_dir ()) {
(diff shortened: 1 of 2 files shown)


src/View/ViewContainer.vala
```diff
@@ -365,33 +365,21 @@ namespace Marlin.View {
...
-                            tab_name = _("Home");
-                        } else if (fn == "/") {
-                            tab_name = _("File System");
-                        }
-                    } catch (ConvertError e) {}
-
-                    if (tab_name == null) {
-                        tab_name = location.get_parse_name ();
-                    }
+                    tab_name = Path.get_basename (path);
```
tintou: Shouldn't we use https://valadoc.org/gio-2.0/GLib.FileAttribute.STANDARD_NAME.html here ?

jeremypw: I used this because `path` is already in hand and known to be sanitized and unescaped at this point.  Also it is slightly simpler (involves fewer entities).

I tried `tab_name = slot.directory.file.info.get_attribute_string (GLib.FileAttribute.STANDARD_DISPLAY_NAME);` and it seems to work OK though.

Is there any advantage of using this form or is it just a style thing?

jeremypw: Also using the attribute gives some non-fatal critical errors during start up because the info is not ready.

tintou: theorically, the path isn't always UTF-8 encoded so the right way is to always use STANDARD_DISPLAY_NAME but yeah if it's not available it's difficult.

jeremypw: I have tested it with e.g. Japanese characters and it works OK.  Do you still want it changed?

tintou: I guess that we can still change it later if there is a real breakage at some point ;) let's go with that

PR #936, opened 2019-05-03
Clarify GOF.File can only have one thumbstate

jeremypw:
* Rename GOF.File member `flags` to `thumbstate`
* Reduce unneeded thumbnailing

It was noticed that the thumbnailer was being called unnecessarily on scrolling. This fixes that and makes the code clearer.

### DIFF
--- a/libcore/File.vala
+++ b/libcore/File.vala
@@ -25,10 +25,10 @@ public class GOF.File : GLib.Object {
     }
 
     public enum ThumbState {
-        UNKNOWN = 0x00,
-        NONE = 0x01,
-        READY = 0x02,
-        LOADING = 0x03
+        UNKNOWN,
+        NONE,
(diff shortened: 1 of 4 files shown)


libcore/File.vala
```diff
@@ -70,7 +70,7 @@ public class GOF.File : GLib.Object {
     public bool is_expanded = false;
     [CCode (cname = "can_unmount")]
     public bool _can_unmount;
-    public uint flags = GOF.File.ThumbState.UNKNOWN;
+    public uint thumbstate = GOF.File.ThumbState.UNKNOWN;
```
tintou: maybe even ?
```suggestion
    public GOF.File.ThumbState thumbstate = GOF.File.ThumbState.UNKNOWN;
```

PR #940, opened 2019-05-07
Add test to check icons haven't been translated

davidmhewitt:
This creates a meson/ninja test that fails when someone has erroneously translated the icon name in one of the `.po` files. I had meant to open this PR in my own fork of files to see if it would fail travis or not, but I accidentally opened it here, so it may as well stay here for visibility.


davidmhewitt:
Seems houston/debian build steps don't run meson tests anyway, so this won't flag a failure in CI.

PR #957, opened 2019-06-02
Open new tab not new window by default when opened from terminal

### LINKED ISSUE #99: Do not open another window when opened from terminal
Most time I have Files open and I want to open the tab directory from the terminal but instead, Files opens another window. I feel like the only time we want a new window is when Files isn't currently opened.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/47739102-do-not-open-another-window-when-opened-from-terminal?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #99

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -110,7 +110,11 @@ public class Marlin.Application : Gtk.Application {
     private int _command_line (ApplicationCommandLine cmd) {
         /* Setup the argument parser */
         bool version = false;
-        bool open_in_tab = false;
+        /* The -t option is redundant but is retained for backward compatability
+         * Locations are always opened in tabs unless -n option specified,
+         * in the active window, if present, or after opening a window if not.
+         */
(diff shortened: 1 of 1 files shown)


peteruithoven:
Just to clarify what this PR changed:
Locations are now always opened in tabs unless `-n` option specified. This is not limited to opening locations from terminal.

jeremypw:
@peteruithoven Yes, that's right.  So any scripts that open a location in Files without an explicit -n option, needing a new window rather than the currently active one, will have to be amended.

PR #960, opened 2019-06-06
Use custom WidgetGrid instead of Gtk.IconView

jeremypw:
Pushed for visibility and comment.

Uses libwidgetgrid from https://github.com/jeremypw/widget-grid.

Uses recycled pool of widgets for display and a Gtk.Layout. 

Replaces custom C model with a Gtk.TreeStore plus appropriate interfaces.

Advantages:  Performance does not deteriorate badly with very large folders.  File items are displayed by widgets not cell renderers and therefore have a stylecontext and more complicated behaviour.

Disadvantages: Maintaining a custom widget.


jeremypw:
Not yet compatible with searchbar.

jeremypw:
Search and sort now working

PR #965, opened 2019-06-11
Module: only use memory if a FileChooserDialog is in use

tintou:
By default, the module was always using memory, even if the application wasn't using any FileChooserDialog. It is now also freeing any used memory after all the dialogs are freed.

### DIFF
--- a/filechooser-module/Plugin.vala
+++ b/filechooser-module/Plugin.vala
@@ -20,44 +20,46 @@
  * Authored by: Corentin Noël <tintou@mailoo.org>
  */
 
-public class PantheonModule.FileChooserDialog : GLib.Object {
-    /* Catching dialogs section by: tintou (https://launchpad.net/~tintou) */
-    Gee.TreeSet<unowned Gtk.FileChooserDialog> tree_set;
-    public FileChooserDialog () {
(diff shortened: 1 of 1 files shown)


jeremypw:
The custom filechooser seems not to be working for me at the moment (always get standard gtk file chooser).  This module is supposed to be replaced by a NativeFileChooser interface at some point.

tintou:
@jeremypw for me,`echo $GTK_MODULES` outputs `:pantheon-filechooser-module:gail:atk-bridge`
is it your case ?

jeremypw:
@tintou: Yes, the same.  but e.g. opening a file with the Code headerbar button brings up the gtk file chooser.

tintou:
weird, it's working here

jeremypw:
It is not working with a 'clean' install of Juno stable either (for me) ?!

jeremypw:
Running Code with --gtk-module=pantheon-filechooser-module gives no error but still does not give the custom chooser dialog.

jeremypw:
Just installed the April unstable iso on a new laptop and still no custom FileChooser using Code :-(

jeremypw:
Hmm. Now the custom FIleChooser is working again on the unstable Juno install, but not on the stable install.  I have just installed some dev related software as well as the chromium browser.

PR #967, opened 2019-06-12
Refactor color tag menuitem drawing

### LINKED ISSUE #952: Color tag margin in context menu inconsistent
The margins above and below the color tags in the context menu seems way smaller than the margins above and below text items.
Screenshot from 2019-05-24 11 (copy) 55 58@2x

I'm assuming these tags are a Files specific widget which is why I chose to open an issue here. 

This could also somewhat help with: https://github.com/elementary/files/issues/618

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/74593650-color-tag-margin-in-context-menu-inconsistent?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
* Align start margin
* Implement RTL

Fixes #952

### DIFF
--- a/plugins/pantheon-files-ctags/plugin.vala
+++ b/plugins/pantheon-files-ctags/plugin.vala
@@ -330,12 +330,17 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
 
     private class ColorWidget : Gtk.MenuItem {
         private new bool has_focus;
-        private int height;
+        private const int BUTTON_WIDTH = 10;
+        private const int BUTTON_HEIGHT = 10;
+        /* Set start margin to match other menuitems. Is there a way to determine
+         * this programmatically?  */
+        private const int MARGIN_START = 27;
+        private const int SPACING = 15;
+
         public signal void color_changed (int ncolor);
 
         public ColorWidget () {
             set_size_request (150, 20);
-            height = 20;
 
             button_press_event.connect (button_pressed_cb);
             draw.connect (on_draw);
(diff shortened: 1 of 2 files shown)


plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -376,23 +376,43 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
+            /* Set start margin to match other menuitems. Is there a way to determine
+             * this programmatically?  */
+            int margin_start = 27;
+            int spacing = 15;
+            int x0 = btnw + spacing;
+
+            if (Gtk.StateFlags.DIR_RTL in get_style_context ().get_state ()) {
+                var width = get_allocated_width ();
+                int x = width - margin_start - btnw;
+                for (i = 0; i <= 9; i++) {
```
tintou: What about this ?
```suggestion
                for (int i = 0; i < TAGS_COLORS.length (); i++) {
```

tintou:
Looks good to me, just a tiny suggestion

plugins/pantheon-files-ctags/plugin.vala
jeremypw: Implemented in latest commits - thanks!

PR #973, opened 2019-06-17
File color chooser widget: GTK instead of Cairo

lainsce:
Instead of drawing with Cairo, use native GTK widgets for HiDPI clarity.

Looks like:
Screenshot from 2019-06-17 18 30 13

### DIFF
--- a/plugins/pantheon-files-ctags/plugin.vala
+++ b/plugins/pantheon-files-ctags/plugin.vala
@@ -329,29 +329,264 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
     }
 
     private class ColorWidget : Gtk.MenuItem {
-        private new bool has_focus;
-        private const int BUTTON_WIDTH = 10;
-        private const int BUTTON_HEIGHT = 10;
+        private const int BUTTON_WIDTH = 16;
+        private const int BUTTON_HEIGHT = 16;
         /* Set start margin to match other menuitems. Is there a way to determine
          * this programmatically?  */
         private const int MARGIN_START = 27;
-        private const int SPACING = 15;
+        private const int SPACING = 10;
 
         public signal void color_changed (int ncolor);
 
         public ColorWidget () {
             set_size_request (150, 20);
 
+            var css_provider = new Gtk.CssProvider();
+            string style = """
+            .color-button {
+                border-bottom-left-radius: 16px;
+                border-top-left-radius: 16px;
+                border-top-right-radius: 16px;
+                border-bottom-right-radius: 16px;
+                text-shadow: 1px 1px transparent;
+                padding: 3px;
+            }
+            .color-red {
+                background-color: @STRAWBERRY_300;
+                border: 1px solid @STRAWBERRY_500;
+                box-shadow:
+                inset 0 0 0 1px alpha (shade (@STRAWBERRY_700, 1.4), 0.05),
+                    inset 0 1px 0 0 alpha (@STRAWBERRY_700, 0.45),
+                    inset 0 -1px 0 0 alpha (#cccccc, 0.15);
+                    background-image:
+                    linear-gradient(
+                        to bottom,
+                        transparent,
+                        transparent 50%,
+                        alpha (
+                            #000,
+                            0.04
+                        )
+                    );
+            }
+            .color-red:active {
+                border-color: @STRAWBERRY_300;
+                box-shadow:
+                    inset 0 1px 0 0 alpha (@STRAWBERRY_700, 0.45),
+                    inset 0 -1px 0 0 alpha (@STRAWBERRY_700, 0.15),
+                    0 0 0 1px alpha (#eeeeee, 0.25);
+                transition: all 100ms ease-out;
+            }
+            .color-orange {
+                background-color: @ORANGE_300;
+                border: 1px solid @ORANGE_500;
+                box-shadow:
+                inset 0 0 0 1px alpha (shade (@ORANGE_700, 1.4), 0.05),
+                    inset 0 1px 0 0 alpha (@ORANGE_700, 0.45),
+                    inset 0 -1px 0 0 alpha (#cccccc, 0.15);
+                    background-image:
+                    linear-gradient(
+                        to bottom,
+                        transparent,
+                        transparent 50%,
+                        alpha (
+                            #000,
+                            0.04
+                        )
+                    );
+            }
+            .color-orange:active {
+                border-color: @ORANGE_300;
+                box-shadow:
+                    inset 0 1px 0 0 alpha (@ORANGE_700, 0.45),
+                    inset 0 -1px 0 0 alpha (@ORANGE_700, 0.15),
+                    0 0 0 1px alpha (#eeeeee, 0.25);
+                transition: all 100ms ease-out;
+            }
+            .color-yellow {
+                background-color: @BANANA_300;
+                border: 1px solid @BANANA_500;
+                box-shadow:
+                inset 0 0 0 1px alpha (shade (@BANANA_700, 1.4), 0.05),
+                    inset 0 1px 0 0 alpha (@BANANA_700, 0.45),
+                    inset 0 -1px 0 0 alpha (#cccccc, 0.15);
+                    background-image:
+                    linear-gradient(
+                        to bottom,
+                        transparent,
+                        transparent 50%,
+                        alpha (
+                            #000,
+                            0.04
+                        )
+                    );
+            }
+            .color-yellow:active {
+                border-color: @BANANA_300;
+                box-shadow:
+                    inset 0 1px 0 0 alpha (@BANANA_700, 0.45),
+                    inset 0 -1px 0 0 alpha (@BANANA_700, 0.15),
+                    0 0 0 1px alpha (#eeeeee, 0.25);
+                transition: all 100ms ease-out;
+            }
+            .color-green {
+                background-color: @LIME_300;
+                border: 1px solid @LIME_500;
+                box-shadow:
+                inset 0 0 0 1px alpha (shade (@LIME_700, 1.4), 0.05),
+                    inset 0 1px 0 0 alpha (@LIME_700, 0.45),
+                    inset 0 -1px 0 0 alpha (#cccccc, 0.15);
+                    background-image:
+                    linear-gradient(
+                        to bottom,
+                        transparent,
+                        transparent 50%,
+                        alpha (
+                            #000,
+                            0.04
+                        )
+                    );
+            }
+            .color-green:active {
+                border-color: @LIME_300;
+                box-shadow:
+                    inset 0 1px 0 0 alpha (@LIME_700, 0.45),
+                    inset 0 -1px 0 0 alpha (@LIME_700, 0.15),
+                    0 0 0 1px alpha (#eeeeee, 0.25);
+                transition: all 100ms ease-out;
+            }
+            .color-blue {
+                background-color: @BLUEBERRY_300;
+                border: 1px solid @BLUEBERRY_500;
+                box-shadow:
+                inset 0 0 0 1px alpha (shade (@BLUEBERRY_700, 1.4), 0.05),
+                    inset 0 1px 0 0 alpha (@BLUEBERRY_700, 0.45),
+                    inset 0 -1px 0 0 alpha (#cccccc, 0.15);
+                    background-image:
+                    linear-gradient(
+                        to bottom,
+                        transparent,
+                        transparent 50%,
+                        alpha (
+                            #000,
+                            0.04
+                        )
+                    );
+            }
+            .color-blue:active {
+                border-color: @BLUEBERRY_300;
+                box-shadow:
+                    inset 0 1px 0 0 alpha (@BLUEBERRY_700, 0.45),
+                    inset 0 -1px 0 0 alpha (@BLUEBERRY_700, 0.15),
+                    0 0 0 1px alpha (#eeeeee, 0.25);
+                transition: all 100ms ease-out;
+            }
+            .color-violet {
+                background-color: @GRAPE_300;
+                border: 1px solid @GRAPE_500;
+                box-shadow:
+                inset 0 0 0 1px alpha (shade (@GRAPE_700, 1.4), 0.05),
+                    inset 0 1px 0 0 alpha (@GRAPE_700, 0.45),
+                    inset 0 -1px 0 0 alpha (#cccccc, 0.15);
+                    background-image:
+                    linear-gradient(
+                        to bottom,
+                        transparent,
+                        transparent 50%,
+                        alpha (
+                            #000,
+                            0.04
+                        )
+                    );
+            }
+            .color-violet:active {
+                border-color: @GRAPE_300;
+                box-shadow:
+                    inset 0 1px 0 0 alpha (@GRAPE_700, 0.45),
+                    inset 0 -1px 0 0 alpha (@GRAPE_700, 0.15),
+                    0 0 0 1px alpha (#eeeeee, 0.25);
+                transition: all 100ms ease-out;
+            }
+            """;
+            try {
+                css_provider.load_from_data(style, -1);
+            } catch (GLib.Error e) {
+                warning ("Failed to parse css style : %s", e.message);
+            }
+            Gtk.StyleContext.add_provider_for_screen (
+                Gdk.Screen.get_default (),
+                css_provider,
+                Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION
+            );
+
+            var color_button_red = new Gtk.Button ();
+            color_button_red.halign = Gtk.Align.CENTER;
+            color_button_red.height_request = BUTTON_WIDTH;
+            color_button_red.width_request = BUTTON_HEIGHT;
+            var color_button_red_context = color_button_red.get_style_context ();
+            color_button_red_context.add_class ("color-button");
+            color_button_red_context.add_class ("color-red");
+
+            var color_button_orange = new Gtk.Button ();
+            color_button_orange.halign = Gtk.Align.CENTER;
+            color_button_orange.height_request = BUTTON_WIDTH;
+            color_button_orange.width_request = BUTTON_HEIGHT;
+            var color_button_orange_context = color_button_orange.get_style_context ();
+            color_button_orange_context.add_class ("color-button");
+            color_button_orange_context.add_class ("color-orange");
+
+            var color_button_yellow = new Gtk.Button ();
+            color_button_yellow.halign = Gtk.Align.CENTER;
+            color_button_yellow.height_request = BUTTON_WIDTH;
+            color_button_yellow.width_request = BUTTON_HEIGHT;
+            var color_button_yellow_context = color_button_yellow.get_style_context ();
+            color_button_yellow_context.add_class ("color-button");
+            color_button_yellow_context.add_class ("color-yellow");
+
+            var color_button_green = new Gtk.Button ();
+            color_button_green.halign = Gtk.Align.CENTER;
+            color_button_green.height_request = BUTTON_WIDTH;
+            color_button_green.width_request = BUTTON_HEIGHT;
+            var color_button_green_context = color_button_green.get_style_context ();
+            color_button_green_context.add_class ("color-button");
+            color_button_green_context.add_class ("color-green");
+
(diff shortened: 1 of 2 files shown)


jeremypw:
This makes RTL handling unnecessary so is a good idea.
Needs @elementary/ux to sign off on change of appearance.  In the mean time I have some comments:

* I suggest ColorButton be made into its own class which you can construct by just providing a color (or color name to substitute into the CSS).
* There does not seem to be any need to use a function to determine which button was clicked - each button can signal when it is clicked and be connected to the same handler.
* ColorWidget is probably large enough class to warrant its own file (although that can be left for another PR).  In general elementary prefers one class per file now.
* Use a construct clause where possible 

There are some elementary code style issues and one runtime warning which I have noted in the code.

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -330,20 +330,300 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
+        // SPACING = geometric center of each of the color buttons.
+        private const int SPACING = 3;
 
         public signal void color_changed (int ncolor);
 
         public ColorWidget () {
             set_size_request (150, 20);
 
+            var css_provider = new Gtk.CssProvider();
+            string style = """
```
jeremypw: This can be DRY'd by parameterizing the color and using a separate ColorButton class.

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -330,20 +330,300 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
+            }
+            .nohover:hover {
+                background: @bg_color;
+            }
+            """;
+            try {
+                css_provider.load_from_data(style, -1);
+            } catch (GLib.Error e) {
+                warning ("Failed to parse css style : %s", e.message);
+            }
```
jeremypw: Insert blank line after try/catch clause.

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -330,20 +330,300 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
+            Gtk.StyleContext.add_provider_for_screen (
+                Gdk.Screen.get_default (),
+                css_provider,
+                Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION
+            );
+
+            var color_button_red = new Gtk.Button ();
```
jeremypw: DRY the code by making ColorButton class.

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -330,20 +330,300 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
+            color_button_slate_context.add_class ("color-slate");
+
+            var color_button_remove = new Gtk.Button ();
+            color_button_remove.halign = Gtk.Align.CENTER;
+            color_button_remove.height_request = BUTTON_WIDTH;
+            color_button_remove.width_request = BUTTON_HEIGHT;
+            var color_button_remove_context = color_button_remove.get_style_context ();
+            color_button_remove_context.add_class ("color-button");
+
+            var colorbox = new Gtk.Box (Gtk.Orientation.HORIZONTAL, -1);
```
jeremypw: Cannot use `-1` as spacing - use >= 0.

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -330,20 +330,300 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
+            colorbox.pack_start (color_button_remove, true, true, 0);
+            colorbox.pack_start (color_button_red, true, true, 0);
+            colorbox.pack_start (color_button_orange, true, true, 0);
+            colorbox.pack_start (color_button_yellow, true, true, 0);
+            colorbox.pack_start (color_button_green, true, true, 0);
+            colorbox.pack_start (color_button_blue, true, true, 0);
+            colorbox.pack_start (color_button_violet, true, true, 0);
+            colorbox.pack_start (color_button_slate, true, true, 0);
+
             button_press_event.connect (button_pressed_cb);
```
jeremypw: Connect each button to same handler and query signal source to determine which color pressed.

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -352,6 +632,12 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
             deselect.connect (() => {
                 has_focus = false;
             });
+
+            this.add (colorbox);
+            // Remove pesky hover state coloring
+            var thiscontext = this.get_style_context ();
```
jeremypw: Unnecessary local variable.

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -330,20 +330,300 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
-        private const int SPACING = 15;
+        // SPACING = geometric center of each of the color buttons.
+        private const int SPACING = 3;
 
         public signal void color_changed (int ncolor);
 
         public ColorWidget () {
             set_size_request (150, 20);
 
+            var css_provider = new Gtk.CssProvider();
```
jeremypw: Insert space before opening bracket.

plugins/pantheon-files-ctags/plugin.vala
lainsce: How would that look like as an example?

lainsce:
> There does not seem to be any need to use a function to determine which button was clicked - each button can signal when it is clicked and be connected to the same handler.

There is. The Gtk.MenuItem widget slurps every pressed signal, so one cannot acess the buttons from within.

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -328,180 +328,151 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
 
-    private class ColorWidget : Gtk.MenuItem {
-        private new bool has_focus;
-        private const int BUTTON_WIDTH = 10;
-        private const int BUTTON_HEIGHT = 10;
-        /* Set start margin to match other menuitems. Is there a way to determine
-         * this programmatically?  */
-        private const int MARGIN_START = 27;
-        private const int SPACING = 15;
+    private class ColorButton : Gtk.Button {
```
jeremypw: You need to give each button an id property (just an int will do) so that the shared signal handler can determine the source button.

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -328,180 +328,151 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
+        private const int BUTTON_WIDTH = 16;
+        private const int BUTTON_HEIGHT = 16;
+        private string color_name = "";
+        private string palette_name = "";
```
jeremypw: Do you need these two members? They are only used in construction, I think.

jeremypw:
Thanks.  I'll have a play with it and see whether I can resolve the signal issue you mention.

plugins/pantheon-files-ctags/plugin.vala
lainsce: It's to assure existence before the button instance is constructed. Found the same fix while developing Notejot. Better safe than sorry.

plugins/pantheon-files-ctags/plugin.vala
lainsce: So in the main public method of the custom class I add in a `int id`, and assign one for each button?

jeremypw:
I would probably have a separate widget for removing the color rather than multi-purposing the color button.

lainsce:
I can do that. Using the cross icon from the old design even.

danrabbit:
The style doesn't seem to be applied here:

Screenshot from 2019-06-18 10 55 43@2x


Would it be possible to use Check menuitems as we did in Calendar? Making the current selection clear is nice:
Screenshot from 2019-06-18 10 57 56@2x

lainsce:
The height disparity is due to the icon'd color remove button, while the others don't have an image.

Check menuitems can be used, although I can't assure that the checkmark will stick or even show.

lainsce:
By using `Gtk.Image` instead of `Gtk.Button`, I could make everyone the same height.

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -328,180 +328,159 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
+
+            Gtk.StyleContext.add_provider_for_screen (
+                Gdk.Screen.get_default (),
+                css_provider,
+                Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION
+            );
+
+            this.height_request = BUTTON_WIDTH;
+            this.width_request = BUTTON_HEIGHT;
+            this.get_style_context ().add_class ("color-button");
```
danrabbit: Instead of getting the style context twice, please cache it in a variable

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -328,180 +328,159 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
-        private const int SPACING = 15;
+    private class ColorButton : Gtk.Image {
+        private const int BUTTON_WIDTH = 16;
+        private const int BUTTON_HEIGHT = 16;
+        private string color_name = "";
+        private string palette_name = "";
+
+        public ColorButton (int id, string color_name, string palette_name) {
+            this.palette_name = palette_name;
+            this.color_name = color_name;
```
danrabbit: Please use GObject-style construction here. The only thing in the constructor should be
```
Object (
    color_name: color_name,
    id: id,
    palette_name: palette_name
)
```

All of these variables should be construction properties, for example, `public int id { get; construct; }`

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -328,180 +328,159 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
+
+            var css_provider = new Gtk.CssProvider ();
+
+            string style = """
+            .color-button {
+                border-bottom-left-radius: 16px;
```
danrabbit: if the radius is the same for all corners, better to just do ` border-radius: 16px`

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -328,180 +328,159 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
+                background-color: @%s\_100;
+                transition: all 100ms ease-out;
+            }
+            .nohover:hover {
+                background: @bg_color;
+            }
+            """.printf(color_name, palette_name, palette_name, color_name, palette_name);
 
+            try {
+                css_provider.load_from_data(style, -1);
```
danrabbit: Please make sure to use spaces before `()`. There are a few instances of this in this PR
```suggestion
                css_provider.load_from_data (style, -1);
```

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -328,180 +328,159 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
+            } catch (GLib.Error e) {
+                warning ("Failed to parse css style : %s", e.message);
+            }
+
+            Gtk.StyleContext.add_provider_for_screen (
```
danrabbit: Instead of adding to the whole screen, we can probably add the style provider for just this widget

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -328,180 +328,159 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
+                Gdk.Screen.get_default (),
+                css_provider,
+                Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION
+            );
+
+            this.height_request = BUTTON_WIDTH;
```
danrabbit: elementary code style is to not use `this` where possible

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -328,180 +328,159 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
-            });
+            Gtk.StyleContext.add_provider_for_screen (
+                Gdk.Screen.get_default (),
+                css_provider,
+                Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION
+            );
+
+            var color_button_remove = new Gtk.Image.from_icon_name ("window-close-symbolic", Gtk.IconSize.MENU);
+            color_button_remove.width_request = 16;
+            color_button_remove.get_style_context ().add_class ("flat");
```
danrabbit: Please use style constants where possible
```suggestion
            color_button_remove.get_style_context ().add_class (Gtk.STYLE_CLASS_FLAT);
```

plugins/pantheon-files-ctags/plugin.vala
lainsce: @jeremypw in an earlier review said to do this. :man_shrugging:
I'll revert to using one var.

plugins/pantheon-files-ctags/plugin.vala
lainsce: If I do that, the custom styling does not apply:

Screenshot from 2019-06-18 18 54 10

lainsce:
> It's a little complicated how you style these widgets. If you're giving them each a unique style class name, why do you need the colors passed in as a variable? It seems like you'd save yourself some trouble to not try to generate a unique stylesheet for each widget and just use a single stylesheet, loaded from a resource, like we do in Calendar

I did it like this because it isn't a case for a global Files CSS place, but only a widget's. Doesn't seem necessary for the scope of this PR.

plugins/pantheon-files-ctags/plugin.vala
```diff
@@ -328,180 +328,125 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
...
         }
 
         private bool button_pressed_cb (Gdk.EventButton event) {
             /* Determine whether a color button was clicked on */
-            int y0 = (get_allocated_height () - BUTTON_HEIGHT) /2;
-            int x0 = BUTTON_WIDTH + SPACING;
+            int y0 = (get_allocated_height () - 16) / 2;
 
-            if (event.y < y0 || event.y > y0 + BUTTON_HEIGHT) {
+            if (event.y < y0 || event.y > y0 + 16) {
```
danrabbit: I don't think you want to get rid of these constants. It's better to have something that makes sense to a future maintainer than just a magic number that isn't clear where it comes from

lainsce: I'll add a block of comments below the opening of this method detailing these.
It's better than to assume they're intrinsic to the button and whatnot (HiDPI button isn't 16 in size, for instance)

jeremypw:
I have not found a way around the signal swallowing problem so for now we will have to stick with the existing method.

I have noticed another downside of this approach in that the colors displayed by the buttons are not necessarily the same as the colors shown in the view.  The former are derived from the theme whereas the latter are hard coded.  Using a different theme causes the buttons to lose any color (and change shape).

You can DRY the code a little more by constructing the buttons in a loop (using the same local variable) adjusting the construction of the buttons accordingly.

```
        private string[] COLORS = {"WHITE", "STRAWBERRY", "ORANGE", "BANANA", "LIME", "BLUEBERRY", "GRAPE", "SLATE"};

...

            for (int i = 0; i < COLORS.length; i++) {
                var color_button = new ColorButton (i, COLORS[i]);
                colorbox.add (color_button);
            }
```

But maybe you should continue to use the TAG_COLORS array for now to fill in the button CSS.

jeremypw:
There are two issues with this solution:

* The color of the buttons is not necessarily the same as the color displayed by the view (at the moment) because the former are derived from the theme and the latter are from a hard coded array of colors.  This is not blocking necessarily but should be fixed at some point.

* This only works properly for the elementary theme.  Other themes show colorless buttons.  Do we need to provide a fallback?

jeremypw:
There does not seem any way around the signal swallowing issue.  We would probably have to use a popover.

lainsce:
> The color of the buttons is not necessarily the same as the color displayed by the view (at the moment) because the former are derived from the theme and the latter are from a hard coded array of colors. This is not blocking necessarily but should be fixed at some point.

This isn't much of an issue, the colors used in the view to me seem like contrast-friendly ones, while the ones used for the buttons just describe what type of color they'll tag the file.

> This only works properly for the elementary theme. Other themes show colorless buttons. Do we need to provide a fallback?

I don't see why we'd need to cater to other themes as Files is designed to be for elementary OS and its theme, only.

> There does not seem any way around the signal swallowing issue. We would probably have to use a popover.

Maybe for a future PR. :wink:

plugins/pantheon-files-ctags/plugin.vala
jeremypw: I think I was commenting on a case where the variable was only used once (or maybe it was used again later and I didnt notice)

jeremypw:
I am OK with the potential color mismatch if @danrabbit is.
I am OK with ignoring other themes but it may be worth making a note in the README that Files is  not intended to work properly with other themes.

jeremypw:
Is it possible to have different styling for the selected color button?  Again, this can be left for another PR.

jeremypw:
I notice that there are fewer colors to choose from now, so some existing color tags will be displayed as white (#fff).

danrabbit:
Yeah I think a small color difference is probably okay since they're shades of the same hue. One is meant for contrast with a black label and the other is for contrast with a white checkmark. (Since we use radios for this in Calendar).

It would be pretty easy to replace the color variables with their values from the elementary stylesheet if this is a blocker for you.

I think I'd still like to see radios used here so that there is an obvious checked selection. I can work on a PR for that against this branch if you'd like.

I agree that the missing colors should be put back

jeremypw:
@danrabbit  I have a draft PR that adds a "checked" style class to the existing ColorButtons (just thickens the border at present) and sets it appropriately but please submit a PR using radio buttons.

jeremypw:
I am still a little concerned that there is no mapping between the color buttons and the `color` number actually stored.  The stored `color` number is an index into the TAGS_COLORS array.  The color buttons are just coded to match

danrabbit:
@jeremypw any ideas about how to resolve this? If I understand correctly the issue is that the menuitem swallows events so we have to guess the widget clicked based on its position? So it seems like getting events from the actual widget clicked would be a prerequisite to having the widget store and return its own values

jeremypw:
@danrabbit One way would be for the position (in order) of the colorbuttons to be mapped to an array containing pairs of hard-coded colors (for the actual tags) and color-names (to be inserted into a parameterised css string.  That way the color-name of the button and the corresponding tag color are specified in the same place.   I agree that if the buttons cannot respond to events there is no point in them storing there own color value once it has been used for construction.

I think you could have the actual tag color be specified by css as well so that the colors of the buttons and tags are the same but that would need some changes to the TextRenderer. That way, the TAGS_COLORS array would contain only the color-names to be substituted into the css.

jeremypw:
If you do not want to use parameterised css then the array could contain style class names instead.

danrabbit:
@jeremypw as far as I can tell, this problem already exists in master. Would it be possible to save a fix for that problem until a follow-up branch?

jeremypw:
@danrabbit  In master the color buttons are drawn with colors taken from the same TAGS_COLORS array that the TextRenderer uses, so tags and buttons have the same color.

But I agree that it is not a blocking issue and can be addressed later if you wish.

danrabbit:
@jeremypw Yeah I feel like the way to go is probably to use the same style class and let the stylesheet handle the actual colors part so that we can make sure contrast is best tailored for each situation and maybe we can do some fun things with selected states too :) But since this would touch a different part of the code base, maybe better for a different PR

jeremypw:
@danrabbit OK.  Have your change requests been met now?

danrabbit:
@jeremypw I think I'm happy with this from a design perspective :) So if you're happy with the code then I think it has a +1 from me

jeremypw:
I have submitted a PR with some minor code fixes but after that it should be mergeable.

cassidyjames:
Screenshot of the new menu with multiple items selected:

Screenshot from 2019-08-05 14 09 07

jeremypw:
Hm - how did I miss that!

PR #977, opened 2019-06-22
Contrasting name text

### LINKED ISSUE #937: Color Labels in Files + Dark Mode = unreadable filenames
elementary OS 5.0, Files, Dark Mode (via System Settings > Tweaks), no labels:

no_labels

Dark Mode with labels:

with_labels

Suggestion: to use dark font color for filenames on color labels in Dark Mode.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/73700595-color-labels-in-files-dark-mode-unreadable-filenames?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #937 

* Relies on unreleased version of Granite for the `Granite.contrasting_foreground_color (Gdk.RGBA background_color)` function, which should be released before this. (CI will fail until Travis CI has latest Granite)

* Also fixes a problem with the ColorWidget size request, although this will be superceded when the Gtk Widget version (#973) is merged.

* The existing tag colors are all pale so result in black text.  To test with darker tag colors, amend the `GOF.Preferences.TAGS_COLORS` array before compiling. Darker tag colors should have white text.

* Works for both light and dark theme variants.

### DIFF
--- a/src/TextRenderer.vala
+++ b/src/TextRenderer.vala
@@ -22,6 +22,8 @@ namespace Marlin {
         const int MAX_LINES = 5;
         private int border_radius;
         private int double_border_radius;
+        private Gtk.CssProvider text_css;
+        private Gdk.RGBA previous_background_rgba;
 
         private Marlin.ZoomLevel _zoom_level;
         public Marlin.ZoomLevel zoom_level {
@@ -69,6 +71,8 @@ namespace Marlin {
 
         construct {
             this.mode = Gtk.CellRendererMode.EDITABLE;
+            text_css = new Gtk.CssProvider ();
+            previous_background_rgba = { 0, 0, 0, 0 };
         }
 
         public TextRenderer (Marlin.ViewMode viewmode) {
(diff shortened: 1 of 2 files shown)


src/TextRenderer.vala
```diff
@@ -126,12 +126,24 @@ namespace Marlin {
                 x_offset += border_radius;
             }
 
+            var provider = new Gtk.CssProvider ();
```
donadigo: I think this could be a little bit more optimized. Instead of creating a new provider every render we can only create it when the result of `contrasting_foreground_color` changed. The provider object also could be also created within the constructor and here we could only call `load_from_data`, we should also probably remove the existing provider if it exists.

jeremypw:
@donadigo : I have tried to implement the optimisation you suggest.  The provider is no longer recreated and only updated when both the background and the contrasting color change.  It is debatable how much of this optimisation is worth the extra code.

A similar optimisation might be possible for the focus css_provider but that is outside the scope of this PR.

PR #988, opened 2019-07-13
Fix selecting pasted files.

### LINKED ISSUE #987: Mark file after copying
I would like to have a visual feedback after copying a small file which can be done like KDE does in Dolphin: marking a file after copying is complete.

jeremypw:
Fixes #987 

* Avoid need for callback when pasting by defaulting to select pasting and dropped files after initial load.
* Avoid polling for tree_frozen when selecting during loading by using notify signal.
* Avoid unneeded boolean member

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -213,9 +213,6 @@ namespace FM {
         private GLib.AppInfo default_app;
         private Gtk.TreePath? hover_path = null;
 
-        /* Rapid keyboard paste support */
-        protected bool select_added_files = false;
-
         public bool renaming {get; protected set; default = false;}
 
         private bool _is_frozen = false;
@@ -249,7 +246,7 @@ namespace FM {
             }
         }
 
-        protected bool tree_frozen = false;
+        protected bool tree_frozen { get; set; default = false; }
         private bool in_trash = false;
         private bool in_recent = false;
         private bool in_network_root = false;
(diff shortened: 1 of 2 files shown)


src/View/AbstractDirectoryView.vala
```diff
@@ -448,16 +448,19 @@ namespace FM {
             GLib.File? focus = focus_location != null ? focus_location.dup () : null;
 
             /* Because the Icon View disconnects the model while loading, we need to wait until
-             * the tree is thawed and the model reconnected before selecting the files */
-            Idle.add_full (GLib.Priority.LOW, () => {
+             * the tree is thawed and the model reconnected before selecting the files.
+             * Using a timeout helps ensure that the files appear in the model before selecting. Using an Idle
+             * sometimes results in the pasted file not being selected because it is not found yet in the model. */
+
+            Timeout.add (20, () => {
```
donadigo: Is there a better way to do this rather than polling for `tree_frozen` every 20ms?
Is it possible to make `tree_frozen` a property and then connect to it here: `notify["tree-frozen"].connect (() => ...`, of course disconnecting from it when we're done?

jeremypw: That could work.  I'll try it.

jeremypw: The problem seems to be that the tree is being unfrozen before the files are retrievable from the model.  The hacky timeout gets around that but I'll try to cure the deeper cause if possible.

jeremypw:
@donadigo : I have implemented your suggestion of using notify instead of a timeout.  Still seems a bit inelegant the way I did it so if there is a better way please let me know.  However, it occurred to me that there is a simpler way of fixing the original issue anyway which means there is a net loss of code.

PR #989, opened 2019-07-15
L10n fixes for Files/Extra

jlnr:
This should solve the merge conflicts for Files/Extra, the last PR only resolved conflicts in the "normal" strings.

https://l10n.elementary.io/projects/files/extra/#alerts


jlnr:
Thanks Dan, and sorry for relying on Jenkins here, I should have tested it locally.

PR #996, opened 2019-07-26
Port simple file operation dialog to Vala, using Granite.MessageDialog

### LINKED ISSUE #995: Empty Trash dialog is GNOME style, not elementary style
Title.

Screenshot from 2019-06-04 15 25 02

jeremypw:
Fixes #995 

The run_simple_dialog code from marlin-file-operations.c is adapted to use Granite.MessageDialog via a new function in StockDialogs.vala. A consequentially unneeded function is removed.

As consequence, all dialogs generated by run_error (), run_warning () and run_question () calls in marlin-file-operations.c will be Granite.MessageDialogs.

Note that many of these dialogs are rarely, if ever, seen because the related issues are now detected before calling marlin-file-operations.  For that reason, no attempt was made to make the code more "Vala-ish".  These dialogs could be shown if a low level error occurred during a file operation however, and when emptying the trash with the sidebar. 

The easiest way to test the linked issue is to try to empty the trash using the Sidebar trash context menu while displaying a folder other than the trash folder.  This calls the marlin-file-operations trash emptying code.  Emptying the trash using the trash plugin while displaying the trash folder does not honour the confirm-trash setting  (this is intentional).


libcore/StockDialogs.vala
```diff
@@ -26,50 +40,63 @@ namespace PF.Dialogs {
...
+                break;
+        }
+
+        var dialog = new Granite.MessageDialog.with_image_from_icon_name (data.primary_text,
+                                                                          data.secondary_text,
+                                                                          image_name,
+                                                                          Gtk.ButtonsType.NONE);
+
+        if (data.button_titles.length == 0) {
+            dialog.add_button (Gtk.Stock.CLOSE, 0);
```
danrabbit: Gtk.Stock.CLOSE has been deprecated since 3.10. I think the solution here is probably to manually create a close button so we don't have to come back and change this for Gtk4

danrabbit:
Other than one little nitpick, looks good :)

Screenshot from 2019-07-26 10 30 42@2x

libcore/StockDialogs.vala
jeremypw: Oops, the C code still uses GTK_STOCK_CANCEL a lot when calling the simple dialog.  I guess I should fix that as well.

PR #999, opened 2019-07-27
Avoid mistranslation of datetime format strings

### LINKED ISSUE #998: Files not using global time format
I have my time format set to AM/PM, but Files uses 24hr in several places. 
Screenshot from 2019-07-26 18 56 17
Screenshot from 2019-07-26 19 00 18

Places that aren't using the globally set time format:

- File names (Like those found in screenshot files taken with the Screenshot app)
- "Modified" column of list view
- Properties of files

jeremypw:
Fixes #998 

Date format strings have been mistranslated (or not updated) in some locales.

For example, in en_GB:

```
#. TRANSLATORS Used when 12h clock has been selected
#: libcore/FileUtils.vala:533
msgid "Yesterday at %-I:%M %p"
msgstr "Yesterday at %-H:%M"
```

This PR replaces all non-translatable (but movable) parts of the strings with %s, which is substituted appropriately in code.

This will require new translations.  An alternative would be to review and correct all translations of relevant strings.

An advantage of this approach is that the formats could be changed without retranslations.

### DIFF
--- a/libcore/FileUtils.vala
+++ b/libcore/FileUtils.vala
@@ -520,32 +520,32 @@ namespace PF.FileUtils {
         switch (now_weekday - disp_weekday) {
             case 0:
                 if (clock_is_24h) {
-                    ///TRANSLATORS Used when 24h clock has been selected
-                    format_string = _("Today at %-H:%M");
+                    ///TRANSLATORS '%s' is a placeholder time in 24hr format. It may be moved but not changed.
+                    format_string = _("Today at %s").printf ("%R");
                 } else {
-                    ///TRANSLATORS Used when 12h clock has been selected
-                    format_string = _("Today at %-I:%M %p");
+                    ///TRANSLATORS '%s' is a placeholder time in 12hr format. It may be moved but not changed.
+                    format_string = _("Today at %s").printf ("%-I:%M %p");
                 }
 
                 break;
             case 1:
                 if (clock_is_24h) {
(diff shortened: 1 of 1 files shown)


ryonakano:
The time format is `%-I:%M %p` (like `3:15 PM`) in English but it is natural to be `%p %-I:%M` (like `午後 3:15`, "午後" means PM) in Japanese. I suppose the time format get unnatural in some language like Japanese once this PR merged because it forbid translators to change the word order of time format itself.

jeremypw:
@ryonakano Hm, that is a drawback of this approach.  I wanted to use the `%r` format string for 12h format like I used `%R` for 24h, but `%r` gives the time including the seconds and timezone instead of just hours and minutes + am/pm.  I assume the result obeys local conventions but this results in a much longer string than desired for the 'informal' setting.
Screenshot from 2019-07-29 13 16 27

I am not sure whether this difference in precision between %R and %r is intentional or a bug in glib.

tintou:
@jeremypw you can probably use Granite function to get the localized format (see get_default_date/time_format in https://valadoc.org/granite/Granite.DateTime.html )

jeremypw:
@tintou Thanks! I did not know (or had forgotten) that function. At least the formats only need to be translated in one place if I use that and its not Files responsibility!

jeremypw:
Much better solution!

PR #1001, opened 2019-07-31
Draw checkerboard background for image items

### LINKED ISSUE #712: White / Light File-Thumbnails are Invisible

jeremypw:
Fixes #712 

Pushed for comment.  No solution to adding a frame or drop shadow to the background using css has yet) been found, but it is possible to add a shadow to the icon itself.  It was considered improper to do this when the icon represents an icon since that may already have a drop shadow incorporated.  At the moment all images have no drop shadow but it would be possible to test if the image is in an icon directory (or is svg?)

### DIFF
--- a/libcore/File.vala
+++ b/libcore/File.vala
@@ -228,6 +228,20 @@ public class GOF.File : GLib.Object {
         return is_desktop_file && !basename.has_suffix (".directory");
     }
 
+    public bool is_image () {
+        if (info == null) {
+            return false;
+        }
+
+        bool is_image = false;
+        unowned string? content_type = get_ftype ();
+        if (content_type != null) {
+            is_image = GLib.ContentType.is_mime_type (content_type, "image/*");
+        }
+
+        return is_image;
+    }
+
     public bool is_trashed () {
         return PF.FileUtils.location_is_in_trash (get_target_location ());
     }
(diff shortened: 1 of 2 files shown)


jeremypw:
Current state of PR:

Screenshot from 2019-08-01 13 36 47

danrabbit:
Looking at the code for the tilerenderer in Music, it seems like you have to subtract the margin size before rendering the background: https://github.com/elementary/music/blob/master/src/Widgets/FastView/TileView/TileRenderer.vala#L102

4jNsY6fCVqZv:
Thanks for fixing this!
I wonder: What will your solutions for PDF's or Open Document file formats look like?
It would be wonderful, if checkerboard or card class can be extended in the sense of these file types too!

PR #1003, opened 2019-08-01
Render background on draw rect

danrabbit:
AFAICT, produces the desired effect:

Screenshot from 2019-08-01 08 39 34@2x

### DIFF
--- a/src/IconRenderer.vala
+++ b/src/IconRenderer.vala
@@ -29,7 +29,6 @@
 namespace Marlin {
 
     public class IconRenderer : Gtk.CellRenderer {
-        private Gtk.CssProvider shadow_provider;
         public Gdk.Rectangle hover_helper_rect;
         public Gdk.Rectangle hover_rect;
         public bool follow_state {get; set;}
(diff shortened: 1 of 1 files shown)


jeremypw:
I wasn't sure whether you wanted a varying size of checkerboard - it looked a bit odd to me. However, if that is the desired effect it easy enough to do.

jeremypw:
Also it does not really solve the linked issue since for most images the checkboard is overwritten by the icon and not visible.  It does solve the drop shadow in a more elegant way though.

danrabbit:
:man_shrugging: 

Screenshot from 2019-08-01 09 03 12@2x

cassidyjames:
Yeah we're really solving two issues:
- white-background images look odd
- transparent background images look odd

Solved by:
- a card-style shadow behind images, making them stand out from the background
- the above, plus checkerboard background

I'm in favor of both of these, and this looks really nice to me!

jeremypw:
I cannot really see the shadow behind the image to be honest.  But maybe its more obvious with mainly white images.  I take it that you do not want a drop shadow below standard icons? There is a slight visual glitch as you zoom past the "normal" thumbnail size and a "large" thumbnail has to be loaded.

jeremypw:
I think I am happy to merge this now and deal with any adjustments later.

Issue #473, opened 2018-05-09
Revise visual representations of focus, selection, etc


danrabbit:
- [ ] Depends on #472 

It's come up that it's a bit unclear and inconsistent how we represent various states between the different views. Attached is a mockup that hopefully clears things up

![bitmap](https://user-images.githubusercontent.com/7277719/39832446-0d3fcffc-537c-11e8-8e55-bb424c18f243.png)

jeremypw:
Yeah, TreeView seems to do it natively whereas IconView doesn't and I am struggling to find the right css to do it on IconView.  The first step was to get the blue rectangle - now I need to get it to turn grey when not keyboard focus.  Notice that in master, it grey all the time and still does not change with keyboard focus.  Both the IconView and TreeView use the same icon and text renderers but the background rectangle is rendered by the widget itself.

jeremypw:
If you want complete consistency between the views, the background rectangle in IconView should turn grey if selected without focus (your issue specified transparent/white)

PR #1004, opened 2019-08-01
Draw frame around prelit/focussed icon.

### LINKED ISSUE #846: Background instead of lightening icon on hover
Lightening the item icon or thumbnail isn't very accessible and can even be problematic for bright thumbnail (see #712). Maybe a better alternative would be to draw a background on hover without modifying the icon/thumbnail. It would also be less subtle and make it more clear if the user click will interact with the target item or the current folder.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/68260108-background-instead-of-lightening-icon-on-hover?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #846.

The first commit draws a subtle frame around the edge of the icon, which is not very obvious until you move it.  A more visualy obvious solution would be to make the frame larger, thicker or different colour.

Suggestions welcome.

### DIFF
--- a/src/IconRenderer.vala
+++ b/src/IconRenderer.vala
@@ -176,10 +176,6 @@ namespace Marlin {
                         }
                     }
                 }
-
-                if (prelit || focused) {
-                    pb = PF.PixbufUtils.lighten (pb);
-                }
             }
 
             if (file.is_image () ) {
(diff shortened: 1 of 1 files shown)


4jNsY6fCVqZv:
Is it possible to share a screenhot?

jeremypw:
I shared an image on Slack with the UX team but they have different ideas so I will withdraw this.

4jNsY6fCVqZv:
What does that mean for the solution of the problem if you withdraw your PR?
What different views were/are there?

jeremypw:
Dan has given some guidance here: https://github.com/elementary/files/issues/473 and I will be trying to implement this.

4jNsY6fCVqZv:
Wow, that looks pretty good!
I'm happy if you can implement it that way.

PR #1006, opened 2019-08-02
Use lookup_icon before choose_icon so hicolor theme searched.

jeremypw:
Fixes https://github.com/elementary/icons/issues/808

### DIFF
--- a/libcore/IconInfo.vala
+++ b/libcore/IconInfo.vala
@@ -90,9 +90,17 @@ public class Marlin.IconInfo : GLib.Object {
 
             var names = ((GLib.ThemedIcon) icon).get_names ();
             var theme = get_icon_theme ();
-            var gtkicon_info = theme.choose_icon_for_scale (names, size, scale, Gtk.IconLookupFlags.FORCE_SIZE);
+
+            /* Looks in default theme and hicolor theme for primary icon name first */
+            var gtkicon_info = theme.lookup_icon_for_scale (names[0], size, scale, Gtk.IconLookupFlags.FORCE_SIZE);
+
             if (gtkicon_info == null) {
-                return new Marlin.IconInfo.for_pixbuf (null);
+                /* Look for fallback icons in default theme only */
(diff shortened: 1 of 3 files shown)


libcore/IconInfo.vala
```diff
@@ -90,9 +90,17 @@ public class Marlin.IconInfo : GLib.Object {
 
             var names = ((GLib.ThemedIcon) icon).get_names ();
             var theme = get_icon_theme ();
-            var gtkicon_info = theme.choose_icon_for_scale (names, size, scale, Gtk.IconLookupFlags.FORCE_SIZE);
+
+            /* Looks in default theme and hicolor theme for primary icon name first */
+            var gtkicon_info = theme.lookup_icon_for_scale (names[0], size, scale, Gtk.IconLookupFlags.FORCE_SIZE);
```
tintou: Why not something like this?
```suggestion
            var gtkicon_info = theme.lookup_by_gicon_for_scale (icon, size, scale, Gtk.IconLookupFlags.FORCE_SIZE);
```

jeremypw: It doesn't work, I am afraid.  It no longer finds hicolor icons in that form.  I do not know why exactly.

4jNsY6fCVqZv: @jeremypw Does PR work in this way or what else does it need to be integrated into Files?

4jNsY6fCVqZv:
@jeremypw Does PR work in this way or what else does it need to be integrated into files?

4jNsY6fCVqZv:
Does PR work in this way or what else does it need to be integrated into Files?

4jNsY6fCVqZv:
@jeremypw Does PR work in this way or what else does it need to be integrated into Files?

jeremypw:
@4jNsY6fCVqZv The PR worked the last time I checked it (some time ago), but I did not get it to work with @tintou 's suggested change.

4jNsY6fCVqZv:
@tintou Is your proposed change mandatory in order to merge this PR?

PR #1008, opened 2019-08-02
Show tag colors in sidebar

### LINKED ISSUE #1002: [Question] Why is the color of a folder/file not reflected in the sidebar?
what it says on the tin!

jeremypw:
Fixes #1002  (maybe).

This is one possible way of addressing the linked issue.
Use Marlin.TextRenderer in SideBar and colorize tagged user bookmark name.

Known bugs:  Does not update sidebar colors if tag color changed in view until sidebar updates e.g. because reordered or something (un)mounted.

Will fix that if determined that this is the way to go.

Adding colored dots to icons is an alternative solution (see #1010).


4jNsY6fCVqZv:
@jeremypw Thanks for working on this! Would you like to share a work in progress screenshot of it?

jeremypw:
I was about to upload screenshot then my internet connection went down :-(  Here it is:

Screenshot from 2019-08-02 19 57 01

4jNsY6fCVqZv:
Well reflects the connection between both views.
And looks kind of noisy at the same time, doesn't it?
What do you think?

PR #1010, opened 2019-08-03
Colorize sidebar icon with tag color.

jeremypw:
Fixes #1008 

Another possible solution for this issue.  

Uses the Marlin.IconRenderer to render sidebar icons and colorize them with tag color where appropriate.

### DIFF
--- a/libcore/AbstractSidebar.vala
+++ b/libcore/AbstractSidebar.vala
@@ -58,6 +58,7 @@ namespace Marlin {
             DISK_SIZE,
             PLUGIN_CALLBACK,
             MENU_MODEL,
+            COLOR,
             COUNT
         }
 
@@ -88,7 +89,8 @@ namespace Marlin {
                                         typeof (uint64),            /* Free space */
                                         typeof (uint64),            /* For disks, total size */
                                         typeof (Marlin.PluginCallbackFunc),
-                                        typeof (GLib.MenuModel)
+                                        typeof (GLib.MenuModel),
+                                        typeof (string)
                                         );
 
             content_box = new Gtk.Box (Gtk.Orientation.VERTICAL, 0);
@@ -110,6 +112,7 @@ namespace Marlin {
             iter = add_place (PlaceType.PLUGIN_ITEM,
                              iter,
                              text,
+                             null,
                              icon,
                              null,
                              null,
@@ -126,6 +129,7 @@ namespace Marlin {
        protected abstract Gtk.TreeIter add_place (Marlin.PlaceType place_type,
                                                   Gtk.TreeIter? parent,
                                                   string name,
+                                                  string? color,
                                                   Icon? icon,
                                                   string? uri,
                                                   Drive? drive,
--- a/libcore/AbstractSidebar.vala
+++ b/libcore/AbstractSidebar.vala
@@ -65,6 +65,7 @@ namespace Marlin {
         protected Gtk.TreeStore store;
         protected Gtk.TreeRowReference network_category_reference;
         protected Gtk.Box content_box;
+        public Marlin.BookmarkList bookmarks { get; protected set; }
 
         protected void init () {
             store = new Gtk.TreeStore (((int)Column.COUNT),
--- a/libcore/Bookmark.vala
+++ b/libcore/Bookmark.vala
@@ -87,6 +87,14 @@ namespace Marlin {
             return gof_file.icon;
         }
 
+        public string? get_color () {
+            if (gof_file != null && gof_file.color > 0) {
+                return GOF.Preferences.TAGS_COLORS[gof_file.color];
+            }
+
+            return null;
+        }
+
         public bool uri_known_not_to_exist () {
             if (!gof_file.location.is_native ()) {
                 return false;
--- a/libcore/BookmarkList.vala
+++ b/libcore/BookmarkList.vala
@@ -270,7 +270,7 @@ namespace Marlin {
             });
         }
 
-        private GLib.List<GOF.File> get_gof_file_list () {
+        public GLib.List<GOF.File> get_gof_file_list () {
             GLib.List<GOF.File> files = null;
             list.@foreach ((bm) => {
                 files.prepend (bm.gof_file);
@@ -279,7 +279,7 @@ namespace Marlin {
         }
 
         private void files_ready (GLib.List<GOF.File> files) {
-            /* Sidebar does not use file.info when updating display so do not signal contents changed */
+            contents_changed ();
             call_when_ready = null;
         }
 
--- a/libcore/File.vala
+++ b/libcore/File.vala
@@ -228,6 +228,20 @@ public class GOF.File : GLib.Object {
         return is_desktop_file && !basename.has_suffix (".directory");
     }
 
+    public bool is_image () {
+        if (info == null) {
+            return false;
+        }
+
+        bool is_image = false;
+        unowned string? content_type = get_ftype ();
+        if (content_type != null) {
+            is_image = GLib.ContentType.is_mime_type (content_type, "image/*");
+        }
+
+        return is_image;
+    }
+
     public bool is_trashed () {
         return PF.FileUtils.location_is_in_trash (get_target_location ());
     }
--- a/libcore/PixbufUtils.vala
+++ b/libcore/PixbufUtils.vala
@@ -17,9 +17,9 @@
  */
 
 namespace PF.PixbufUtils {
-    private uint8 lighten_component (uint8 cur_value) {
+    private uint8 lighten_component (uint8 cur_value, uint power) {
         uint new_value = cur_value;
-        new_value += 24 + (new_value >> 3);
+        new_value += (24 * power + (new_value >> 3));
         if (new_value > uint8.MAX) {
             new_value = uint8.MAX;
         }
@@ -27,7 +27,7 @@ namespace PF.PixbufUtils {
         return (uint8) new_value;
     }
 
-    public Gdk.Pixbuf lighten (Gdk.Pixbuf src) {
+    public Gdk.Pixbuf lighten (Gdk.Pixbuf src, uint power = 1) {
         GLib.return_val_if_fail ((!src.has_alpha && src.n_channels == 3) || (src.has_alpha && src.n_channels == 4),
                                  src);
         GLib.return_val_if_fail (src.bits_per_sample == 8, src);
@@ -46,9 +46,9 @@ namespace PF.PixbufUtils {
             int dst_row = i * dst_row_stride;
             for (int j = 0; j < width; j++) {
                 var width_offset = j * channels;
-                target_pix[dst_row + width_offset] = lighten_component (original_pix[src_row + width_offset]);
-                target_pix[dst_row + width_offset + 1] = lighten_component (original_pix[src_row + width_offset + 1]);
-                target_pix[dst_row + width_offset + 2] = lighten_component (original_pix[src_row + width_offset + 2]);
+                target_pix[dst_row + width_offset] = lighten_component (original_pix[src_row + width_offset], power);
+                target_pix[dst_row + width_offset + 1] = lighten_component (original_pix[src_row + width_offset + 1], power);
+                target_pix[dst_row + width_offset + 2] = lighten_component (original_pix[src_row + width_offset + 2], power);
                 if (has_alpha) {
                     target_pix[dst_row + width_offset + 3] = original_pix[src_row + width_offset + 3];
                 }
--- a/plugins/pantheon-files-ctags/plugin.vala
+++ b/plugins/pantheon-files-ctags/plugin.vala
@@ -325,9 +325,11 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
                 foreach (unowned GOF.File file in files) {
                     if (file.location.has_uri_scheme ("recent")) {
                         update_file_info (file);
-                        file.icon_changed (); /* Just need to trigger redraw */
                     }
+
+                    file.icon_changed ();
                 }
+
             } catch (Error err) {
                 warning ("%s", err.message);
             }
@@ -453,6 +455,19 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
             }
 
             return true;
+        }
+    }
+
+    public override void update_sidebar (Gtk.Widget widget) {
+        Marlin.AbstractSidebar sidebar;
+        if (widget is Marlin.AbstractSidebar) {
+            sidebar = (Marlin.AbstractSidebar)widget;
+        } else {
+            return;
+        }
+
+        foreach (GOF.File gof in sidebar.bookmarks.get_gof_file_list ()) {
+            update_file_info (gof);
         }
     }
 }
--- a/src/IconRenderer.vala
+++ b/src/IconRenderer.vala
@@ -42,7 +42,6 @@ namespace Marlin {
             set {
                 _zoom_level = value;
                 icon_size = Marlin.zoom_level_to_icon_size (_zoom_level);
-                show_emblems = _zoom_level > Marlin.ZoomLevel.SMALLEST;
             }
         }
 
@@ -213,8 +212,7 @@ namespace Marlin {
 
             style_context.restore ();
 
-            int h_overlap = int.min (draw_rect.width, Marlin.IconSize.EMBLEM) / 2;
-            int v_overlap = int.min (draw_rect.height, Marlin.IconSize.EMBLEM) / 2;
+            var emblem_size = (cell_area.width / 4).clamp (12, 24);
             if (file != null && (selected || prelit) && file != drop_file) {
                 special_icon_name = null;
                 if (selected && prelit) {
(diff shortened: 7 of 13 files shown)


jeremypw:
Screenshot from 2019-08-03 19 10 42

jeremypw:
Adding dots to icon did not result in clear colors for reasonable sized dots relative to the icons.

jeremypw:
This solution was used as there is already a suitable function in Files to colorize icons,  however alternative suggestions are welcome.

4jNsY6fCVqZv:
That doesn't seem visually coherent to me. Because first the text and then the folders are connected with the same color.

Three thoughts on that:
1 How do your previous solutions work if I bookmark and tag files in addition to the folders?
2. can you test what it looks like if the colored dots are placed next to the folder names or file names? So instead of putting a dot over the icon.
Examples would be Notes-up or the new Tasks app https://github.com/elementary/tasks/pull/8
3 Another idea would be to fundamentally rethink the concept. What if users no longer tag the folder name or file name with color, but the folder or file icon itself? Standard folders of course have a uniform color. The same goes for new folders. Only if they are tagged, they change their color. I imagine this to be useful for folders, unfortunately I can't imagine it for files right now.

jeremypw:
Just trying things in order of code complexity, but all solutions will require the tag-color to be communicated to the sidebar renderer(s) so a lot of the code is not wasted.  Colorizing the name would be more coherent but could be made less busy with dots or something.

1. Bookmarked files receive a generic icon in the sidebar (no thumbnailing) so this would be colorized in the same way at the moment although it would be simple enough to only colorize folders. The colorization could be lightened easily.

2. Do you mean instead of the icons or as part of the name column? The Notes Up app uses a SourceList subclass for the sidebar which is built from widgets not rendered by CellRenderers like the Files sidebar so the code cannot be directly imported unfortunately.

3.  This might conflict with the intention to colorize system folder icons.

jeremypw:
Another possible solution using a colorized favorite emblem on icon.

Screenshot from 2019-08-04 12 00 11

jeremypw:
For consistency, could use favorite emblems in view as well.

Screenshot from 2019-08-04 12 26 35

jeremypw:
@cassidyjames Some possible implementations of this. Obviously the position and appearance of the tags is open to suggestion.  It is easier to tag either the icon or the text in the same way both the view and the sidebar as they (now) share the same renderers.

4jNsY6fCVqZv:
@jeremypw I love your idea that with the heart an emblem comes into play, which also expresses that I favor a certain file or folder. That wasn't the case yet. And I think much better than my idea with the colored dots. And to your question from above. Yes exactly, I meant colored dots as part of the name column.

Are there currently other apps in elementary OS that use "Favorites"? That would also be a good impulse for the new Tasks App or Friends App, wouldn't it?

As you already write: I would make the position of the tags even more coherent. So either emblem on the icon like in the sidebar or emblem left above the icon like in the view (is that how you call the two views?)

It would actually be important to me that you insert Open Document files or PDFs or photos into your test folder to see how this works with the emblems as bookmarks with these file types.

Another side note: I've seen that I can add bookmarks using the right-click context menu, but I can't remove them the same way, can I? Why not? Doesn't that make sense?

4jNsY6fCVqZv:
In your screenshot I see the folder with the lock emblem. Maybe this is also a good position for the heart? Or are there even other emblems that could express something similar?

jeremypw:
@4jNsY6fCVqZv Glad you like the favorite emblem idea.  The position is open to discussion.  Bottom left may be better as it is not used by anything else in the view.  Top left is used by the helper and there could be up to four file property emblems on the right side (although I've never seen more than two used).  It may be possible to combine tagging and file property emblem code somehow, if it is decided that tag emblems are the way to go. I'll look into what other favorite icons exist and are used in other elementary projects.

You can remove bookmarks from a context menu, but it is the *bookmark* context menu.  

You can only unfavorite bookmarks by removing the color tag from the corresponding folder in a view at the moment but a bookmark context menu option could be added.

lainsce:
What if there was a sidebar place to access every color tag, and when clicked, it'd show all files with tag:red. That way, we don't clutter the bookmarks part of the sidebar.

4jNsY6fCVqZv:
@jeremypw Lower left sounds good! It should not overlap with other emblems.
There is an extra bookmark context menu? Since the bookmarks haven't got an extra colored label yet, I probably didn't notice that yet :-) Or are all entries in the sidebar = bookmarks?
As written, an additional option in the view would be possible. But maybe there are more elegant visual solutions? 
Thank you for researching elementary OS projects for alternative emblems and other "favorites" solutions!

4jNsY6fCVqZv:
@lainsce Thank you for your contribution to this discussion! Do you want to try out how your idea would look visually?

lainsce:
@4jNsY6fCVqZv It'd roughly look like this in the sidebar.

Screenshot from 2019-08-04 14 21 49

jeremypw:
@4jNsY6fCVqZv All the entries in the sidebar except categories have a context menu, but its content varies depending on the entry.

jeremypw:
@lainsce  That looks cool, but extends the functionality of the sidebar beyond a navigation aid.  It would take a lot of new code to implement a "FilterView" like this although it is something I have considered in relation to search results.   

At the moment views can only display one unfiltered folder (or in the case of e.g. recent:// a virtual folder).  I assumed that "all files tagged red" meant red-tagged files*anywhere* in the file-system?  Or did you mean just the current view?

This is getting bit off-topic for this PR - maybe raise a wishlist issue where it can be discussed?

lainsce:
Well, it is a wishlist thing alright, but it'd unclutter the bookmarks.
And yeah, all files tagged red in the places the user can modify.

jeremypw:
By "unclutter the bookmarks", do you mean you could tag a number of bookmarked locations e.g. "red" and replace them with a single "red" bookmark tag?

lainsce:
Yep.

jeremypw:
Tag at bottom left, bookmark to tagged jpg:

Screenshot from 2019-08-04 19 32 35

jeremypw:
It may look better with the sidebar tag in a separate column, but that will take a bit more code...

4jNsY6fCVqZv:
How is it that hearts are of different sizes?
If I recognize this correctly between folder/jpg and PDF.

How does it look with files from the group discussed here?
https://github.com/elementary/files/issues/846
https://github.com/elementary/files/issues/712

> It may look better with the sidebar tag in a separate column, but that will take a bit more code...

I don't understand that yet. Right now I don't see any colored tags in the sidebar of your last screenshot.

4jNsY6fCVqZv:
By the way: I think it looks better from version to version :smile:

jeremypw:
The emblem size is supposed to be related to the size of the icon so that it does not overwhelm it if a non-scalable icon is zoomed.  Because some icons have a transparent border the size is not always what is expected.  An alternative is to have the emblems at the edge of the cell area and all the same size, but in that case they may be some distance from the icon.  It is not easy to get a reasonable compromise.

jeremypw:
In the last screenshot you should see small heart-shaped emblems overlaying the bookmark icons related to tagged folders.

jeremypw:
Maybe easier to see at a lower zoom level:

Screenshot from 2019-08-04 20 01 42

4jNsY6fCVqZv:
Oh okey, I see it now. Can you please try a variant where in all views the emblem is displayed in the lower left corner next to the icons? Then the icons won't be overlaid as much, will they?

4jNsY6fCVqZv:
>  An alternative is to have the emblems at the edge of the cell area and all the same size, but in that case they may be some distance from the icon.

It's worth a try, isn't it?

jeremypw:
I will not have time for a couple of days now, but I'll upload a screenshot when I can.

4jNsY6fCVqZv:
No stress, a good thing takes time!

4jNsY6fCVqZv:
@lainsce I would like to share a feedback about your screenshot as well. Unfortunately GitHub can't give direct answers to posts like GitLab, can it?

I think it's a wonderful idea, and I'd be happy if you'd create a new issue for it, so that we can further develop the visual inspirations and their practical implementation there! What do you think?

jeremypw:
Latest version:  the helper, tags and emblems are sized and positioned more consistently.  At the moment, they touch the edgest of the cell area unless that would cause then to be separated from the icon. It would be possible to allow a certain gap between emblems etc and the icon if desired but it is undesirable to have them visually disconnected, which would happen with small icon at a large zoom level if they always appeared at the cell edges (see second screenshot).

May overlap but not separate from icon:
Screenshot from 2019-08-06 12 20 29

Fixed to cell edges:
Screenshot from 2019-08-06 12 32 30

4jNsY6fCVqZv:
Screenshot 1: Looks very good already! Do you have an example where overlaps can occur?  And what would that look like?
Currently I don't see any coherence between the positioning in the view and in the sidebar. The icons in the sidebar are still overlaid by the emblem. Do you still want to adjust that?

Screenshot 2: Yes, with the distance the distance is already very large. Auch hi Also here there is the visual gap between the view and the sidebar.

I'm just asking myself: How do these two solutions work in different (but hidden in the UX) zoom levels?

4jNsY6fCVqZv:
I just had time to take another look at the two screenshots. Only now did I realize that you were trying to show what a file/folder looks like when different emblems are displayed at the same time. I notice that especially in the .svg & .jpg examples the lock varies very much in its position. To be even more precise: With the links it doesn't look right, because there two emblems coincide. Can the elegant loosening be done with four corners, no need to remain unused, right?

This gives me an idea: If we have found a good position, can you please also match the positions of the emblems at the other corners with the position of the new heart emblem?

jeremypw:
@4jNsY6fCVqZv The vertical position of the emblems depends on the order in which they are added to the emblem list added to the file object which is determined by the order in which file properties are tested in code.  So "is_symlink" is tested first and, if true that emblem drawn first (at the bottom). If that symlink is also readonly then that emblem is added next.  However, a normal readonly file will have the lock icon at the bottom since it is the only one.

The position of helper, tag and emblems are already matched, but unfortunately the different icons corresponding to these may have differing amounts of clear space at the edge (as do the icons to which they are applied) so they never look *exactly* in the same position.

The "appointment-missed.svg" icon even has an emblem built into it, which makes it look odd with added emblems.

The code already allows for up to four emblems and the fourth one would occupy the "unused" top right corner.  I agree that if we limit the number of emblems to two then they could be put in the right side corners but we might want to implement e.g. a "shared folder" emblem in future.

I'll post some more screenshots for other zoom levels and view modes.

jeremypw:
List View, small icons:

Screenshot from 2019-08-06 16 53 14

Column view: smallest icons:

Screenshot from 2019-08-06 16 53 51

jeremypw:
Note that in master, emblems are suppressed for the smallest icon size as they tended to obscure the icon excessively.  I have attempted to make everything usable at the smallest icon size but it is a compromise.

jeremypw:
I agree I need to change the sidebar layout so that there is some space around the icon where the tag can be put without being (completely) superimposed on the icon.

4jNsY6fCVqZv:
@jeremypw Thanks for sharing!

> The "appointment-missed.svg" icon even has an emblem built into it, which makes it look odd with added emblems.

Is this a default icon in the sense of the freedesktop.org icon naming spec? Or could the icon be improved elegantly to solve this problem?

>  but unfortunately the different icons corresponding to these may have differing amounts of clear space at the edge (as do the icons to which they are applied) so they never look exactly in the same position.

Wouldn't it make sense to adjust the free space at the edges of the icons?

>  I agree that if we limit the number of emblems to two then they could be put in the right side corners but we might want to implement e.g. a "shared folder" emblem in future.

You want to keep the upper right corner free for the emblem of shared folders?

jeremypw:
@4jNsY6fCVqZv 

> Is this a default icon in the sense of the freedesktop.org icon naming spec? Or could the icon be improved elegantly to solve this problem?

I am not sure - this icon appears in other themes with a different appearance but still has a badge of some kind in the bottom right corner.  In principle elementary theme could redesign it but there are several other icons with the same characteristic.  I think it is something we will have to live with for now.

> Wouldn't it make sense to adjust the free space at the edges of the icons?

There is no efficient way of doing this that I am aware of - you would have to analyse each pixbuf and  it is likely to hit performance more than is justified. But the UX team may have a different opinion.

>You want to keep the upper right corner free for the emblem of shared folders?

To achieve this, we just need to make sure that the "shared folder" emblem is added to the list last, which is quite simple.

jeremypw:
Adjusted sidebar color tag position:

Screenshot from 2019-08-07 16 26 58

lainsce:
What if the tag was put after the filename? It'd mimic what we do for git status in Code while not pushing every icon off-layout.

jeremypw:
@lainsce That would be possible but require more code - the reason it was done this way was to reuse the same icon renderer that draws the icons (and tags) in the main view.  It also gives coherence between the sidebar and view.

I am just trying out different things at the moment; ultimately its up to @elementary/ux of course.

jeremypw:
Of course, the view color tag could also be put after the name.  I did a different PR where the sidebar shared the view text renderer.

4jNsY6fCVqZv:
By the way, who is currently part of the UX team? @elementary/ux

jeremypw:
@4jNsY6fCVqZv  See https://github.com/orgs/elementary/teams/ux

4jNsY6fCVqZv:
@4jNsY6fCVqZv Sorry, I don't think I have the permission to open your link. Probably the teams are hidden, aren't they? I'm just getting a "404 page not found" error.

jeremypw:
Oh, sorry - I did not realise that. 

Screenshot from 2019-08-09 07 10 43

4jNsY6fCVqZv:
No problem, thanks! Was only asked out of curiosity.

4jNsY6fCVqZv:
> Of course, the view color tag could also be put after the name. I did a different PR where the sidebar shared the view text renderer.

Can you please link the other PR?

jeremypw:
@4jNsY6fCVqZv Sorry, I deleted that PR - I did not want to distract potential reviewers from this one which is further advanced.

4jNsY6fCVqZv:
I understand that very well! I'm curious what the review team will say!

jeremypw:
@cassidyjames Np - the code that communicates the tag color to the sidebar can be re-used.  I'll work on a PR implementing your suggestion.

4jNsY6fCVqZv:
@cassidyjames May I ask what exactly irritates you about this PR? It was always communicated to involve the UX team.

What do you mean by a more iterative direction?

4jNsY6fCVqZv:
@cassidyjames @cassidyjames May I ask what exactly irritates you about this PR? It was always communicated to involve the UX team here. Did you discuss the topic in a team or is that just your current perception after you have read yourself into the PR?

What do you mean by a more iterative direction?

For comparison with Code status-style dot: I don't know if a status expresses the same quality as a bookmark. In this context the heart symbol already makes sense for me, because it is probably self-explanatory for the users, while the status point could be a concept users are familiar with from communication apps (online/offline etc.).

What do you think it takes to introduce a new symbol into elementary's bookmark workflow? From a UX perspective, what does it take to continue working with the existing PR?

cassidyjames:
@4jNsY6fCVqZv nothing irritated me, I was just trying to follow the thinking and was a bit confused about how the solution here was arrived at. I followed the comments and PRs but regret I was unable to get to commenting on this earlier.

Marking a file as a color is not the same as bookmarking it, and bookmarking isn't the same as marking it with a color. The two should not be conflated. The only time a color dot would show on a bookmark in the sidebar is if the user themself already marked a folder with a color, so there's little risk of being confused. At the same time, we already use stars for favoriting/bookmarking elsewhere, so if we were to use a symbol for bookmarking, I'd probably prefer that. **But marking a file with a color is not the same as bookmarking** so that's not super relevant.

By more iterative, I mean making small, individual improvements like first adding the color to a dot in the sidebar instead of introducing a new symbol, using that symbol in the sidebar, and changing how color-tagged files look throughout all of Files. It's _much_ harder for someone on the UX team to review a change when it touches several parts of the workflow, because we have to spend time understanding the original problem being solved and how the presented solution was arrived at. That's part of the reason it took me so long to review this; it was a daunting conversation and PR to follow compared to a single, small change.

4jNsY6fCVqZv:
@cassidyjames 

> But marking a file with a color is not the same as bookmarking

For me, this issue is exclusively about bookmarking a file or folder. My original concern was never to colorize a file or folder. Please have a look at https://github.com/elementary/files/issues/1002  Both views should be visually coherent for me. Unfortunately, they are not yet.
With my issue I wanted to achieve an improvement. And that's how the idea of tagging came about. 

Which native apps currently use the stars for bookmarking? For me, that would also be a conceivable improvement on this question. At least if I can set the stars in different colours, because bookmarking should also express different qualities for me. For example, priorities, rankings or subject areas.

Another question would be whether the heart might be the more beautiful alternative to the star. However, that is not the priority of my concern.

I also want to ask you: Is your feedback your personal feedback or did you discuss it in the UX team?

jeremypw:
The conflation of the original issue with indicating tagged/favorited files in the main view is probably my fault: I wanted to reuse as much code as possible.  If the sidebar tagging/favoriting is done in a different way than colorizing the text or adding an emblem to the icon then there is no incentive to touching the main view.

jeremypw:
@4jNsY6fCVqZv This PR will be replaced by another one implementing @cassidyjames suggestion of a dot (similar to Code) so the issue will be addressed in due course.

4jNsY6fCVqZv:
@jeremypw What's your personal opinion on that? After all, you've put a lot of creativity and work into this PR so far, haven't you?

jeremypw:
I am happy to defer to the UX team on matters of design and, in any case, they have the final say.

4jNsY6fCVqZv:
I'm curious to see how the issue develops and look forward to a shared screenshot with real code. Then I can see if this resonates with my original request.

4jNsY6fCVqZv:
@jeremypw Is there already an idea how to proceed with the solution of the original issue?

jeremypw:
Sorry, I think the project leaders are busy at the moment with more important matters to do with re-engineering the release process and other things - wishlist level issues are not a high priority I guess.

PR #1014, opened 2019-08-07
[In progress] Git support

jeremypw:
Initial attempt to enable git awareness in Files.
The code is largely copied from the equivalent Code code.


libcore/gof-directory-async.vala
```diff
@@ -235,6 +249,30 @@ public class Async : Object {
...
 
+    public void update_git_status () {
+        if (git_repo == null) {
+            return;
+        }
+
+        var options = new Ggit.StatusOptions (Ggit.StatusOption.INCLUDE_UNTRACKED, Ggit.StatusShow.INDEX_AND_WORKDIR, null);
+
+        try {
+            /* FIXME This seg faults for unknown reasons */
```
vjr: Have you tried changing the `check_each_git_status` function from private to public?

jeremypw: Thanks Vishal - David Hewitt advised that the problem is that I had not initialised the Ggit library before using this function.  Confusingly, all the other Ggit library functions that I used worked OK so this was not easy to spot.

jeremypw:
I presume this is failing Travis CI because need to add libgit2-glib-1.0 dependency to "deb-packaging" branch.

jeremypw:
Withdrawing in favour of a plugin approach.

PR #1016, opened 2019-08-08
Git plugin

### LINKED ISSUE #877: Integrate Git status
I could have sworn we had this filed somewhere but didn't find it. Since Files is a more advanced workflow (as it deals with files directly versus "content" via the default content apps), I wonder if it would make sense to integrate Git awareness and status into Files. This could be super handy for managing repos with a lot of assets to see a visual indication of status, especially when the contents of the repo aren't all "code" (and so don't make sense to manage via the Code app).

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/69464111-integrate-git-status?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #877


jeremypw:
I'll try.

tintou:
Looks great, I wonder why Marlin.Plugins.Base isn't a GObject subclass? I would then allow to use a `construct` and not have this `initialize` or is there a thing that I'm not understanding?

jeremypw:
Yeah, I noticed that - I thought it was something to do with the way modules are loaded and I didn't want to mess with it.  I could experiment.

jeremypw:
Ah, there is a `new` call in `module_init' function - I missed that.  Making Plugin.Base a GObject does not seem to break anything - I guess it was omitted for compactness originally?   Or we could just initialize in `module_init`?

tintou:
Rebased and modified, we can probably iterate in the future, we also need a review from UX

jeremypw:
Now I have basic code approval from tintou, I'll work on completing the UX.   Looking at Code code, it uses "user-available" and "user-away" so I can easily use the same.  I'll leave out the "Git information" option for now. It can always be added later if there are requests for more detailed info.

jeremypw:
@cassidyjames I have made the requested changes.  I notice that the yellow dot has fairly poor contrast against the buff folders.

cassidyjames:
@jeremypw I think it's fine; it doesn't really overlap the folders and the consistency with other developer tools is nice.

PR #1023, opened 2019-08-11
Confirm dialogs always shown if called.

### LINKED ISSUE #1022: [Feature Confirmation] 'Shift + delete' behaviour under below conditions.
Please look at the below scenario that I just observed -

Conditions::
The files to be deleted needs to be from same folder.

Scenario 1::
File (select) -> (Right click) delete -> trash(open via clicking the trash icon) -> (Right Click)delete the file just deleted in trash which is from the folder in previous step ::    Here dialogue box is not present as a warning.

After executing the above scenario. 
Come to the folder from which files were deleted  in Scenario 1. 
Now perform Scenario 2.

Scenario 2::
File (select) -> shift + delete(any file) : No dialogue box appears as a warning.

jeremypw:
Fixes #1022 

The file operation confirmation functions should do one thing - show the dialog.

### DIFF
--- a/libcore/marlin-file-operations.c
+++ b/libcore/marlin-file-operations.c
@@ -1212,13 +1212,10 @@ confirm_delete_from_trash (CommonJob *job,
     int file_count;
     int response;
 
-    /* Just Say Yes if the preference says not to confirm. */
-    if (!should_confirm_trash ()) {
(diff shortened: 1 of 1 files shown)


jeremypw:
Incorporated in #1024 so closing.

PR #1026, opened 2019-08-13
Window: Fix first run coords

danrabbit:
Checks that the window coordinates are not `-1` (the default) before moving. This way the first time you open Files it spawns its window in the center of the primary display.


tintou:
That diff would be smaller with a simple:
```vala
if (default_x == -1 || default_y == -1) {
    return
}
```

jeremypw:
@tintou:  That would prematurely exit the function.  There are some more lines before the function ends (not shown in unexpanded diff).

PR #1032, opened 2019-08-22
Do not create duplicate Trash tab from dock

### LINKED ISSUE #981: Multiple Files instances open up
If you open Files app from Trash Doklet Icon, all subsequent clicks on that icon will open a new instance of Files pointing to trash folder, I think that correct behavior at this moment should be to bring to front the Files window already opened instead open another instance of the same app.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/76510904-multiple-files-instances-open-up?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1104: Check for equivalent existing tab when locating downloaded files
This issue seems very similar to this: [#265](https://github.com/elementary/terminal/pull/265).

### Steps to reproduce:
1. Open Firefox.
2. Download anything.
3. Click on 'Open containing folder'
4. Opens a new download tab and highlights the file.

### Expected Behaviour:
1. Locates the file in the same tab if it is already open.
2. Auto-scrolls the window to bring it into view. (usually take manual scrolling to locate the highlighted file)

jeremypw:
Fixes #981 
Fixes #1104 

When opening a location in Files from the commandline, dock or otherwise, do not create a duplicate tab if Files is already open with a tab at that location.  Instead, just present the Files window with that location tab active.


jeremypw:
Hmm, don't know how I missed that fairly obvious fault in logic!  Hopefully now correct.

src/View/Window.vala
```diff
@@ -422,7 +427,25 @@ namespace Marlin.View {
         }
 
         private void add_tab (File location = File.new_for_commandline_arg (Environment.get_home_dir ()),
-                             Marlin.ViewMode mode = Marlin.ViewMode.PREFERRED) {
+                             Marlin.ViewMode mode = Marlin.ViewMode.PREFERRED,
+                             bool ignore_duplicate = false) {
+
+            if (ignore_duplicate) {
+                bool is_child = false;
```
donadigo: This doesn't need to be initialized.

src/View/Window.vala
```diff
@@ -449,6 +472,30 @@ namespace Marlin.View {
             content.add_view (mode, location);
         }
 
+        private int location_is_duplicate (GLib.File location, out bool is_child) {
+            is_child = false;
+            string parent_path = "";
+            string uri = location.get_uri ();
+            /* Ensures consistent format of protocol and path */
+            parent_path = PF.FileUtils.get_parent_path_from_path (location.get_path ());
+            int existing_position = -1;
```
donadigo: Wouldn't it be better to initialize this to 0 and then increment `existing_position` at the end of the `foreach`?

jeremypw: Done

src/View/Window.vala
jeremypw: Done

PR #1036, opened 2019-09-01
Mixed mode

### LINKED ISSUE #1021: Single click on folders, double-click on files (miller columns)
If you have enabled double-click mode and use the miller columns, you would have to press a folder only once to open it, while keeping a double-click required to open files.
I really think this could improve the time you need to navigate if you use miller-columns. 

This would fit to #690.

jeremypw:
Fixes #1021 

Only one click behaviour - the single-click setting is lost.  Now a single click on a folder (without mod key) always activates (navigates), while single click on a file always selects.  To select a folder use rubberband selection or click with a mod key or use helper. To activate a normal file, double-click.

As before, double-clicking on a folder in Miller View makes that folder the root of the view.


KomalleanaByzantis:
If my opinion matters, I am strongly opposed to this change. I prefer to only click on files and folders once to reduce the strain on the fingers. Please do not remove the setting.

jeremypw:
@KomalleanaByzantis Everyone's opinion matters! Do you do open a lot of files by clicking on them in Files then? If the files are of a similar type then you might consider opening them using the default app for them.  You can still navigate around with a single click with this PR.

KomalleanaByzantis:
@jeremypw Many of the files that I open are documents; the default application for opening them is Files. My media library is of such great variety and structured in such a specific way that using Photos or Videos would be a regression; I would be less productive than I am when I use Files. If your suggested change is implemented, I would be forced to click the left mouse button more often despite that reducing strain is important to me.

jeremypw:
OK, thanks for the explanation - it is always interesting to learn more about how different people apps are used in practice.  This is just an experimental PR in response to complaints about the default single-click mode but I guess it will be impossible to please everyone with only one mode.

KomalleanaByzantis:
@jeremypw I would certainly not rule out the possibility of this mode being useful to some. However, if it is to be implemented, I believe it would be most suitable to make it optional so that one does not force established users to change their habits. You could argue that single-clicking could be supplanted by a mostly keyboard-based use of the application where the mouse purely serves the purpose of multi-selecting files and folders; however, not everyone prefers to browse her or his files in this manner.

If accidentally opening files is an issue, it is likely only an issue in the views that are not the Grid View, as Grid View offers ample space to deselect files. In my opinion, the most consistent and graceful solution would be the following:
- Ensuring that single-clicking or double-clicking files and folders is made consistent across all views based on the user setting of whether to use single-clicking or double-clicking to open a file or folder
- Implementing a key that, when held while clicking on a file or folder with the single-clicking option enabled, merely selects the file or folder and deselects everything else

For example, with the option to single-click a file or folder to open it being disabled, one would select a file or folder by clicking on it once, no matter the view. Contrariwise, with the option enabled, one would open a file or folder by clicking on it once. This proposed alternative would ensure consistent behaviour across the views and would make Files a less confusing application.

jeremypw:
@KomalleanaByzantis Thanks for your input.The only (intended) inconsistency I know of between views is that in Miller view (column view), double-clicking while in single-click mode has a special action that is not accessible in double-click mode (it opens the current folder as the root of the view).  Whether full consistency is more value than having that special action I am not sure.

There is already a modified click that selects the clicked item instead of acitvating it when in single-click mode (and deselects all else) - `<Alt>PrimaryClick`.

KomalleanaByzantis:
@jeremypw Ah, I see. I clearly do not know enough about keyboard shortcuts. If more people knew of that key, and if double-clicking to open files and folders were enabled by default for all views, then perhaps there would be no issue. A toggle somewhere in the Mouse Settings could enable single-clicking for files and folders, and no matter the preferred setting, one would double-click while holding Alt to make a folder the root in the Column View. What do you think of this suggestion? It would not constitute feature bloat, as it is a solution that appeals to all parties in a situation where no one size fits all.

Perhaps a formal introduction to the operating system would help in situations like these, such as a tour like that of Windows XP. [Remember that?](http://gtu.ge/Arch/Faculty/Multimedia/Resources/MS/XP/Tours/htmlTour/start_fr.htm) Haha. On a serious note, if the normal assumption is that a new user of elementary OS is already well-versed in the use of computers, then perhaps that assumption should change. Not everything is inherently intuitive, including the Alt key. At any rate, I may attempt to create such a guided interaction myself one day if elementary does not.

jeremypw:
There is this repo: https://github.com/elementary/onboarding so work is underway to make the learning curve less steep for inexperienced users.  

You can get help with the system shortcuts by pressing the `Super` key ("Windows key", to the left of the `Alt` key). I think it would be good if this could also show the shortcuts relavent to the active app or there was some other consistent way of exposing all the shortcuts at least those not already exposed by a tooltip or menu. This has been proposed here: https://github.com/elementary/shortcut-overlay/issues/53 but it is currently blocked because of technical difficulties.

Probably the best place to discuss alternatives to this PR (as opposed to minor amendments) is in another (or an existing issue) that describes the proposed alternative mouse handling in some detail.

I think it would be difficult to persuade the elementary UX team to approve a third option for Files mouse handling (in general user UX options are deprecated).  A new Mouse setting would have to be relevent to all apps not just Files. However, the good news is that resistance to an AppMenu in Files to expose global UI settings like "hidden files" and "click-mode" has reduced and such a menu would now be considered.

jeremypw:
OK, for now I have retained the original modes to lessen the barrier for merging.  There are now three click modes - single, double and mixed.  Implementing these is not too complicated, fortunately.  This required changing a boolean setting to an enum.  I will produce a separate PR exposing these settings in an AppMenu to aid testing.

jeremypw:
Closing in favour of #1067, which includes this.

cassidyjames:
Just an update from the UX team for those lurking: we had a call and discussed this. One of the most persistent things we hear from folks (especially switchers) is that opening files with a single click can be jarring and unexpected. A way we talked about resolving this while still keeping the single-click navigation through folders ended up sounding similar to this mixed-mode branch. We have also discussed in the past how having two or three different click-modes unnecessarily complicates the codebase and leads to other issues, especially on less-tested hidden settings. So something we want to try and see how it feels in practice:

- **A new hybrid/mixed mode as the only click mode.** This simplifies the codebase and forces us as designers to ensure this works better for more people, instead of having the sort of clutch of "well, there's a hidden setting…"

- **Single-click to navigate folders.** This is the current default behavior, and retains the sort of web browsing metaphor and how navigating through other UIs works.

- **Single-clicking a file shows Properties.** Since Files is a file manager and not an app launcher, it might make sense that the "final destination" when navigating to a file is to show the details about that file, not necessarily to open a different app.

- **Add an "Open in Foo"** suggested action button to Properties, to make a single-click flow still work well for opening files while making it more explicit this is opening a new app.

- **Properties as a right-side pane.** This could be in a follow-up branch to keep things simpler. But putting properties in a side pane is consistent with what we've done in other apps (see: Music), and feels less invasive than opening a dialog on top of the current view. It especially will feel nice in the Column view, but should work well everywhere. It could also sort of act as a "preview" mode if we get higher-res thumbnails/previews in the properties… but that's probably getting ahead of ourselves.

- **Double-click a file to open it in the default app.** This makes the behavior to leave Files for another app a little higher-barrier and less likely to happen accidentally, while also making it familiar/easily discoverable for those coming from other platforms.

matfantinel:
@cassidyjames this is a great improvement, and I really like the idea of the Properties side pane. I agree opening the dialog would be a bit jarring and could confuse users. 

I'd just add that, with the side pane, we'd also be able to select the file with a single click too. As in the side pane, when shown, would always display the selected file's properties. I believe one of the File Manager's most used functions is to move files, so this would make expose two important features on a single click(properties and selection), while also helping avoid accidentally opening files. Sounds like a win-win for me!

jeremypw:
@cassidyjames One issue not explicitly addressed is how to show the properties of a folder?  I am assuming we are proposing to change the "show on hover" behaviour to "show on select" (not "show on click") ?  This would mean to get the properties of a folder you would need to click on the selection helper right?  Or select using the keyboard.

jeremypw:
The showing of properties will be addressed in a separate branch.

cassidyjames:
I have been trying this for a bit and really like where it's going UX-wise. I think it will feel even better if we get a Properties sidebar on file select, as then something still happens when clicking files. Of course something to figure out there will be if the sidebar causes a reflow, or overlays, or what. But I do think that can be addressed in a future branch.

matfantinel:
Can this just be compiled from source in Hera? I have been forcing myself to use single-click mode for a while now but still haven't got used to it. Would love to be a lab rat for this

libcore/Enums.vala
```diff
@@ -72,6 +72,12 @@ namespace Marlin {
         INVALID
     }
 
+    public enum ClickMode {
+        SINGLE,
+        DOUBLE,
+        MIXED,
+    }
+
```
cassidyjames, line 80: Is this needed anymore?

cassidyjames:
> Can this just be compiled from source in Hera?

We also are using a newer LibHandy in the current version than is available in Hera. Sorry about that! That's the cost of switching development over to elementary OS 6… just something else to look forward to in the new release! :nerd_face:

danrabbit:
With this branch I'm getting a crash when clicking folders in Column view

cassidyjames:
@danrabbit ah yeah I can confirm that here as well.

BAProductions:
> The showing of properties will be addressed in a separate branch.

Great idea, But I think mixmode should be the default miller view & work in reverse where a single open the folder & a double click to open a file like on Mac OS

libcore/Enums.vala
jeremypw: Its gone ...

janxkoci:
> if the sidebar causes a reflow, or overlays, or what

Borrowing an idea from the Monitor app, you could have the sidebar always there. When opening new Files window or when no file/folder is selected, the sidebar could show properties of current folder.

jeremypw:
After creating a quick mockup, it is pretty annoying to have the properties widget appearing, resizing and disappearing as the selection changes if that causes the view to resize/reflow.  It needs to be decided how much info to put in the properties "sidebar" as this will govern how large it needs to be.  I guess it will need a shortcut to show/hide it.  If we are only going to show the same info as at present then we may as well stick with the overlay.

BAProductions:
> After creating a quick mockup, it is pretty annoying to have the properties widget appearing, resizing and disappearing as the selection changes if that causes the view to resize/reflow. It needs to be decided how much info to put in the properties "sidebar" as this will govern how large it needs to be. I guess it will need a shortcut to show/hide it. If we are only going to show the same info as at present then we may as well stick with the overlay.

There no need decided how much info to put in the properties "sidebar" simply make it resizable & let the user decided how much info  in the properties "sidebar" based on how wade they make the resizable container

janxkoci:
Apologies for my a bit arrogant comment above - I'm no UX designer of course (although my [OCPD](https://en.wikipedia.org/wiki/Obsessive%E2%80%93compulsive_personality_disorder) makes me believe at times that my opinion is expert enough).

I suggested that the sidebar is the key to usability, but I was only thinking about the usability **for me**. And I should have mentioned that my use case may not be a typical use case.

For start, I actually disable the single-click to open files & folders. It's like the first thing I do after installing the system. In the past, I had computers with different operating systems and having one machine with single-click actions and another one with double-click actions would probably drive my muscle memory cray-cray. Now that I run only elementary OS on all my machines, I could maybe change that and try out the single-click actions. Although there are still other computers and I already scroll in the wrong direction on other people's Windows machines, out of habit.

I'm not even in favour of permanent sidebar with file properties etc, as I tend to keep some app windows at conveniently small sizes, so it's easier to multitask and see multiple windows at once. I like the current minimalist design of Files, and permanent sidebar would force me to enlarge the Files' window to accommodate the new sidebar. On the other hand, if this is the direction the development will go, I personally see the sidebar as the only option that I (and my muscle memory) can live with :smiley: 

To conclude, I personally like the current status quo, I use list view as my preferred default view to see the properties I need and I just panic a little when I see changes like this. But I will get over it. You guys make some awesome software here and I trust you and your decisions. Please keep going :wink:

jeremypw:
> With this branch I'm getting a crash when clicking folders in Column view

Should be fixed now.

cassidyjames:
@jeremypw yeah, resizing/reflowing the view could definitely get annoying. I have to look at what we do in Music again and see if a similar pattern would work there—but I do think doing this in phases would make sense. Using this mixed mode branch has become pretty natural to me already even without the properties sidebar.

jeremypw:
Music has a resizable properties sidebar in IconView which appears when there is a selection (and disappears when there is not), causing a reflow. However, the sidebar does not change size according to its contents.

cassidyjames (dismissed):
This seems to break dragging files, at least in icon mode

jeremypw:
> This seems to break dragging files, at least in icon mode

Confirmed for primary drag although secondary drag is working.  Will fix.

jeremypw:
Removing "Needs Design" label as design input has been received.

cassidyjames:
@jeremypw in using this a bit more, I wonder if we should change how the list view works as well. I think it might feel more natural to expand/collapse folders on single-click instead of navigating into them. Double-clicking them could navigate into them. This way it's a lot easier to expand and collapse instead of aiming for the tiny little expander icon.

jeremypw:
@cassidyjames  I think this would be confusing and break the consistent "click on folder to navigate" behaviour.  You can expand the folder by clicking anywhere to the left of it, not just on the arrow so the target is no that small (larger than the "helper" icon). It would also be inconsistent with any apps using the standard Gtk TreeView widget with expanders.

I would prefer to focus this PR on navigate/select and perhaps raise an issue to consider any changes to the expand folder behaviour.

janxkoci:
Funny, I actually tap right arrow on my keyboard to expand folders in list view (my default view btw) instead of mousing to that arrow icon.

This makes me wonder - does the keyboard navigation change in any way with these changes?

jeremypw:
@janxkoci This PR does not change keyboard navigation.

cassidyjames (commented):
I'm happy with these changes UX-wise but would still like a more thorough code review from @elementary/desktop-developers

meisenzahl (approved):
The changes look good to me! I'm not familiar with the codebase though. Maybe someone else should also review 😅️

leonardo-lemos (approved):
Code looks good for me. Approved.

suchipi:
> If you can give detailed reasons why mixed-mode cannot be used for your work flow (and your work flow cannot be adapted to it)

My workflow could probably be adapted to it with enough work, but it's very inconvenient, because I use macOS at work and Windows 10 for gaming, so I'm used to single-clicking on folders to select them (in Finder and Explorer). If I *only* used elementaryOS, it wouldn't be too bad, but since I switch back and forth, it's pretty jarring.

Also, it looks like the link to further discussion didn't make it into your last comment. Could you repost it?

jeremypw:
The missing link is: https://github.com/elementary/files/discussions/1729

The need for some users to switch oses is, imo, probably the strongest argument for an option. What options (if any) are available on Windows and MacOS?

suchipi:
There doesn't seem to be an option in Windows at all; there's only double-click mode. In Finder, there's *mostly* only a double-click mode, but when you use the columns view, a single click on a folder when there's no column open on the right opens a new column to the right, but once there's a column open on the right, a single click on a folder selects it instead. For what it's worth, I've always found that behavior of Finder confusing, so I avoid using the columns view.

ChildishGiant:
I've got pretty used to mixed mode already but I see no harm in keeping the option for users who want to keep what they're used to or dual boot etc

jeremypw:
I do have a lot of sympathy for users of two opinionated OSes with different modes of operation, neither of which has an option to change it!

andrewdbate:
Sorry to post a comment, but I have to say I'm affected by this too. I just can't get used to this change, and for me the only resort has been to install Nemo. I certainly miss the column view, but I have just not got used to this “single click to enter directories”. I keep entering directories and having to hit back.

This change doesn't help users with disabilities who now must be extra careful when moving the mouse around.

One reason I grew tired of Gnome was the opinionated and dictatorial design philosophy where it is deemed that the user must not be given any choice, and where Gnome UX design decisions are justified by “it simplifies the code”, and I fear that Elementary OS is now going in the same direction.

I was surprised to see how such a tiny code change (the PR diff) could cause such a massive headache.

janxkoci:
@andrewdbate see also the extended discussion here: https://github.com/elementary/files/discussions/1729

jeremypw:
If you are mainly interested in selecting folders then you may find it easier to use ListView by default.  In this view you can single-click anywhere under the Size, Type or Modified columns, without a modifier, to select a folder.

andrewdbate:
@jeremypw Thanks, Jeremy. I had discovered that behavior though trial-and-error, although I had assumed that behavior was a bug / unintended. I'll be using Ctrl + click because I don't want to change views or move the mouse away just to make a selection... but at this point I've already set Nemo as my default file manager. Other file managers don't make me think that much... I'm still convinced Finder on the Mac has it right.

PR #1051, opened 2019-09-27
MarlinFileOperations: Remove the need of a custom f() printf function

tintou:
This reduces code complexity, making it less error-prone and easier to translate into Vala


libcore/marlin-file-operations.c
```diff
@@ -1327,29 +1265,31 @@ report_delete_progress (CommonJob *job,
...
-        /// The singular/plural form will be used depending on the remaining time (i.e. the %T argument).
-        time_left_s = f (ngettext ("%T left",
-                                   "%T left",
-                                   seconds_count_format_time_units (remaining_time)),
-                         remaining_time);
+        /// TRANSLATORS: %s will expand to a time like "2 minutes". It must not be translated or removed.
+        /// The singular/plural form will be used depending on the remaining time (i.e. the %s argument).
+        time_left_s = g_strdup_printf (ngettext ("%s left",
+                                                 "%s left",
+                                                 formated_time),
```
jeremypw: IS this right? `formated_time` is a `gchar` but `ngettext` expects `ulong` as 3rd parameter. The `format_time` function has already used `ngettext`.  Strangely, the code compiles??

libcore/marlin-file-operations.c
```diff
@@ -1279,21 +1216,22 @@ confirm_delete_directly (CommonJob *job,
...
+        prompt = g_strdup_printf (_("Are you sure you want to permanently delete \"%s\"?"), basename);
+        g_free (basename);
     } else {
-        prompt = f (ngettext("Are you sure you want to permanently delete "
-                             "the %'d selected item?",
-                             "Are you sure you want to permanently delete "
-                             "the %'d selected items?", file_count),
-                    file_count);
+        prompt = g_strdup_printf (ngettext("Are you sure you want to permanently delete "
+                                           "the %'d selected item?",
```
jeremypw: I did not know you can concatenate strings like this!

libcore/marlin-file-operations.c
tintou: Oops, fixed, the code compiles because we remove the C warnings, I've re-enabled the warnings and fixed some issues on-the-go

PR #1061, opened 2019-10-06
Fix breadcrumb animation

Exalm:
* Use add_tick_callback() instead of counting frames manually
* Use easeOutCubic interpolator
* Don't clip the arrow at the end to avoid blinking
* Don't stop the animation until the element is fully hidden, i.e. animate it going under the arrow of the previous element as well.
* Align arrow position to pixel grid.

Peek 2019-10-06 23-45_1

It still blinks when changing location though, but that's a different issue.

### DIFF
--- a/libwidgets/Chrome/BasicBreadcrumbsEntry.vala
+++ b/libwidgets/Chrome/BasicBreadcrumbsEntry.vala
@@ -584,6 +584,11 @@ namespace Marlin.View.Chrome {
             }
         }
 
+        private double ease_out_cubic (double t) {
+            double p = t - 1;
+            return 1 + p * p * p;
+        }
+
         private uint make_animation (Gee.Collection<BreadcrumbElement> els,
                                      double initial_offset,
                                      double final_offset,
@@ -594,7 +599,8 @@ namespace Marlin.View.Chrome {
             var anim = add_tick_callback ((widget, frame_clock) => {
                 int64 time = frame_clock.get_frame_time ();
                 double t = (double) (time - start_time) / LOCATION_BAR_ANIMATION_TIME_USEC;
-                t = 1 - t.clamp (0, 1);
+                t = 1 - ease_out_cubic (t.clamp (0, 1));
+
                 anim_state = final_offset + (initial_offset - final_offset) * t;
 
                 foreach (BreadcrumbElement bread in els) {
--- a/libwidgets/Chrome/BasicBreadcrumbsEntry.vala
+++ b/libwidgets/Chrome/BasicBreadcrumbsEntry.vala
@@ -552,7 +552,7 @@ namespace Marlin.View.Chrome {
         private void replace_elements (Gee.ArrayList<BreadcrumbElement> new_elements) {
             /* Stop any animation */
             if (animation_timeout_id > 0) {
-                Source.remove (animation_timeout_id);
+                remove_tick_callback (animation_timeout_id);
                 animation_timeout_id = 0;
             }
             old_elements = null;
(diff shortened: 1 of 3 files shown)


libwidgets/Chrome/BreadcrumbElement.vala
```diff
@@ -197,6 +198,10 @@ public class Marlin.View.Chrome.BreadcrumbElement : Object {
             icon_to_draw = PF.PixbufUtils.lucent (icon_to_draw, 50);
         }
 
+        if (offset_px >= width) {
+            room_for_text = false;
```
Exalm: Hm, and this is not correct, I've missed icon here.

Exalm: I've redone this with masking. See commit descriptions for details, there's sth weird going on with clipping + `render_layout()`.

Exalm:
Sorry for the churn, found a few code style issues. :)

libwidgets/Chrome/BasicBreadcrumbsEntry.vala
```diff
@@ -584,33 +584,35 @@ namespace Marlin.View.Chrome {
...
+            var anim = add_tick_callback ((widget, frame_clock) => {
+                int64 time = frame_clock.get_frame_time ();
+                double t = (double) (time - start_time) / LOCATION_BAR_ANIMATION_TIME_USEC;
+                t = 1 - ease_out_cubic (t.clamp (0, 1));
+
+                anim_state = final_offset + (initial_offset - final_offset) * t;
 
+                foreach (BreadcrumbElement bread in els) {
+                    bread.offset = anim_state;
+                }
```
jeremypw: Lint: blank line after loop clause

libwidgets/Chrome/BreadcrumbElement.vala
```diff
@@ -79,6 +79,23 @@ public class Marlin.View.Chrome.BreadcrumbElement : Object {
         this.icon_info = icon_info;
     }
 
+    private Cairo.Surface? get_mask (double x1, double y1, double x2, double y2, int scale, Cairo.Path? clip_path) {
+        if (clip_path == null)
+            return null;
+
```
jeremypw: Lint: missing braces

libwidgets/Chrome/BreadcrumbElement.vala
```diff
@@ -250,6 +279,15 @@ public class Marlin.View.Chrome.BreadcrumbElement : Object {
...
         }
 
+        var group = cr.pop_group ();
+        cr.set_source (group);
+        if (mask != null) {
+            cr.mask_surface (mask, clip_x1, clip_y1);
+        } else {
+            cr.paint ();
+        }
+        cr.restore ();
```
jeremypw: Lint: Missing blank line after if clause

jeremypw:
The difference from master is quite hard to see on my system, I am afraid.  The redrawing of the unchanged breadcrumbs causes a slight flicker even when only adding or removing the final directory.  It may be that master is worse on other systems where the frame-rate is not 60Hz?  Nevertheless using add_tick_callback() is an improvement.

jeremypw:
It probably needs a more extensive change to this widget e.g. by using an Overlay or Revealer and separating the BreadCrumbs from the Entry.

Exalm:
> The difference from master is quite hard to see on my system, I am afraid. The redrawing of the unchanged breadcrumbs causes a slight flicker even when only adding or removing the final directory. 

This only touches the appear/disappear animation. I have elementary OS in a VM, and the difference is quite noticeable tbh. :)

Peek 2019-10-07 21-11

The flickering is still there indeed.

Exalm:
The other flicker seems to be caused by the fact the pathbar briefly becomes insensitive, so it's not really drawing-related.

Exalm:
Hm, I've resolved all discussions, why does github still say there are changes requested?

jeremypw:
OK, maybe I am looking at the wrong thing.  You need to dismiss the previous review and request a new one (if the reviewer does not notice your changes). Its best to keep a separate commit when addressing review issues rather than force push so the reviewer can see what has changed.

Exalm:
Ah, I see. That's quite different from the workflow I'm used to (GitLab doesn't have reviews, only separate discussions, so you just resolve them + GNOME usually doesn't squash commits so you correct them and force-push because they will be merged as is, so "fix this and that" is unwanted), but makes sense when the commits are squashed :) Will know, thanks!

PR #1063, opened 2019-10-09
RFC: Disable location bar when loading without making it insensitive

### LINKED ISSUE #1062: Path bar sometimes flickers when changing folders
See https://github.com/elementary/files/pull/1061#issuecomment-539094011

Peek 2019-10-07 21-11

When changing a folder, path bar flickers a bit. It's caused by the fact it's set to insensitive while the next folder loads. It makes sense when the folder is remote, but the loading usually happens fast enough on local locations to last just a couple frames.

Exalm:
Just disable keyboard and mouse input instead, so that it doesn't flicker. I don't particularly like this tbh, but at least it's better than flickering.

Fixes #1062

### DIFF
--- a/src/View/Widgets/BreadcrumbsEntry.vala
+++ b/src/View/Widgets/BreadcrumbsEntry.vala
@@ -20,6 +20,9 @@
 
 namespace Marlin.View.Chrome {
     public class BreadcrumbsEntry : BasicBreadcrumbsEntry {
+        /** Support disabling the entry during long operations without making it insensitive **/
+        public bool working { get; set; }
+
         /** Breadcrumb context menu support **/
         ulong files_menu_dir_handler_id = 0;
         Gtk.Menu menu;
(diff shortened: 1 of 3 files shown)


danrabbit:
This seems like a bad idea. If this widget needs to be insensitive, then we need to communicate that visually to users.

I guess a better question is does it really need to be insensitive? If so, can we delay the insensitivity by a few hundred miliseconds to avoid the flicker on fast loading directories?

jeremypw:
Adding (another) arbitrary timeout doesn't seem like a good idea either; this PR is at best a quick temporary fix.  The better solution is to make sure that the pathbar behaves appropriately whenever it is clicked on, but this might need more work and more careful testing.  There are `cancel ()` functions which should help.

jeremypw:
I am going to leave this unmerged for now as I am unsure that the issue is important enough in itself to warrant adding another quick fix to Files involving a timeout.  Hopefully, someone will cleanup the pathbar code in a way that will also fix this issue. See also #1066 for a possible (but controversial) quick fix.

PR #1074, opened 2019-10-20
Lose unused is_drag_pending function

jeremypw:


jeremypw:
> Seems like you can also remove the `drag_has_begun` variable

Good point - removed.

PR #1083, opened 2019-10-23
UndoManager: Make undo() and redo() asynchronous

tintou:


libcore/marlin-undostack-manager.c
```diff
@@ -370,18 +370,21 @@ marlin_undo_manager_is_undo_redo (MarlinUndoManager *manager)
  * Redoes the last file operation
 ** ****************************************************************/
 void
-marlin_undo_manager_redo (MarlinUndoManager *self,
-                          GtkWidget *parent_view,
-                          MarlinUndoFinishCallback cb,
-                          gpointer callback_data)
+marlin_undo_manager_redo (MarlinUndoManager   *self,
```
jeremypw: Maybe rename `marlin_undo_manager_redo_async` ?

libcore/marlin-undostack-manager.c
```diff
@@ -481,25 +484,39 @@ marlin_undo_manager_redo (MarlinUndoManager *self,
...
 
 /** ****************************************************************
  * Undoes the last file operation
 ** ****************************************************************/
 void
-marlin_undo_manager_undo (MarlinUndoManager *self,
-                          GtkWidget *parent_view,
-                          MarlinUndoFinishCallback cb,
-                          gpointer done_callback_data)
+marlin_undo_manager_undo (MarlinUndoManager   *self,
```
jeremypw: Rename `..._async`?

jeremypw:
Is there anything new we need to inhibit in the main context while undo/redo is ongoing asynchronously?

jeremypw:
Are there circumstances where the undo/redo needs to be cancelled?  Is it a good idea to cancel an undo half-way through?  What does it mean?  Should those parts of the undo completed be reversed?

jeremypw:
Where the undo operation is itself async (e.g. UNDO_MOVE) the task completion callback triggers *before* the operation actually completes.  Is this intended?  This causes the "after_undo_redo" Window function to be called while the file operations are ongoing which does not seem right.  Although, checking the master code I think the same issue occurs there.  I think the MarlinUndoFinishCallback fuinction should be called in the `undo_redo_transfer_callback` function when appropriate.  I am not sure why some file operations are async and other user sync GLib functions (which would not show and progress window).  Its a long time since I looked at this part of the code :-/

tintou:
> Where the undo operation is itself async (e.g. UNDO_MOVE) the task completion callback triggers _before_ the operation actually completes. Is this intended? This causes the "after_undo_redo" Window function to be called while the file operations are ongoing which does not seem right. Although, checking the master code I think the same issue occurs there. I think the MarlinUndoFinishCallback fuinction should be called in the `undo_redo_transfer_callback` function when appropriate. I am not sure why some file operations are async and other user sync GLib functions (which would not show and progress window). Its a long time since I looked at this part of the code :-/

I tried to keep the same behavior as in master, I agree that the completion is asynchronous but that can be reworked later :)

tintou:
> Are there circumstances where the undo/redo needs to be cancelled? Is it a good idea to cancel an undo half-way through? What does it mean? Should those parts of the undo completed be reversed?

In some cases it might make sense, if you copy a folder over a high-latency network and undo the operation you also might want to cancel the operation I assume. We might want to give a NULL value here for the time being, but having it supported in some places might be a good thing regarding the future

libcore/marlin-undostack-manager.c
tintou: It is not mandatory to call it *_async as there is a *_finish anyway so whatever you perfer

jeremypw: I guess it only matters where there are two similar functions, one sync and one async like in Gio so just leave it.

PR #1084, opened 2019-10-24
git: Do not assume that we have access to child files

tintou:
Opening a folder with a git submodule freeze files

### DIFF
--- a/plugins/git/plugin.vala
+++ b/plugins/git/plugin.vala
@@ -59,7 +59,8 @@ public class Marlin.Plugins.Git : Marlin.Plugins.Base {
             Ggit.StatusFlags new_flag = Ggit.StatusFlags.CURRENT;
             try {
                 if (file.is_directory) {
-                    git_repo.file_status_foreach (new Ggit.StatusOptions (0, Ggit.StatusShow.WORKDIR_ONLY, null), (path, status_flags) => {
+                    var pathspec = "*".concat (file.basename, "*");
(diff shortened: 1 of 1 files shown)


plugins/git/plugin.vala
```diff
@@ -46,17 +46,44 @@ public class Marlin.Plugins.Git : Marlin.Plugins.Base {
...
             } catch (GLib.Error e) {
                 return; /* If this fails then unlikely to be able to get git_status */
             }
 
-            var git_status = update_git_status (file.location, git_repo, file.is_directory);
-            switch (git_status) {
+            Ggit.StatusFlags new_flag = Ggit.StatusFlags.CURRENT;
+            try {
+                if (file.is_directory) {
+                    git_repo.file_status_foreach (new Ggit.StatusOptions (0, Ggit.StatusShow.WORKDIR_ONLY, null), (path, status_flags) => {
```
jeremypw: This matches too many files resulting in modified emblems being applied to directories which contain no modified files.
On solution is to include a pathspec parameter `{pathspec}` instead of `null`, where `var pathspec = "*".concat (file.basename, "*");`.

jeremypw:
Thanks! I am not sure now which way of concatenating strings is better - using `concat ()`, `string.join ()` or just adding them in this context.  Please use the method you prefer.

jeremypw:
@tintou Is this issue still occurring with the latest master? Do you have an example of a repo I can use to test this?

jeremypw:
Issue not occurring in current master.

PR #1087, opened 2019-10-24
Fix DnD with XDirectSave

### LINKED ISSUE #1082: Dragging file out of archive breaks subsequent internal DnD

### To Reproduce
1. Compress some files using "Compress" context menu option.
2. Open the resulting archive by clicking on it.
3. Drag a file out of the archive and drop into original view (choose "Replace" in dialog)
4. Start to drag another file *in the view* using primary button.
5  DnD breaks - there is no drag icon and when the drag button is released the drag does not end.
6 Escape from the problem by pressing `<Alt>F4`

### Expected behavior
DnD behaves normally

https://gfycat.com/jovialgreenblackrussianterrier

Occurs with Elementary Juno unstable.

-I'm using the latest version from git that I've manually compiled

see also #890

jeremypw:
Fixes #1082 

Extra care is taken to take into account which signals are emitted in which order during drag and drop and making sure that the correct destination related data is reset at the correct times after dropping or leaving the destination view.

Changes are commented to assist review.

It is intended in future to separate the drag source and drop destination related code into separate clases/interfaces to clarify what is going on.


jeremypw:
Closing because of excessive conflict with another commit refactoring DnD.  Need to start again.

PR #1090, opened 2019-10-27
Add .github folder and some other default files

jeremypw:


jeremypw:
Oh, OK - I misunderstood how these work then.  I suppose a lot of issue reporters are just ignoring the templates then - I assumed they were not seeing them.  I'll close this.

PR #1091, opened 2019-10-27
Cleanup ViewMode string handling

jeremypw:
Use ViewMode enum functions instead of hard-coded strings in Window and ViewSwitcher.

### DIFF
--- a/libcore/Enums.vala
+++ b/libcore/Enums.vala
@@ -60,8 +60,6 @@ namespace Marlin {
         }
     }
 
-
-
     public enum ViewMode {
         /* First three modes must match the corresponding mode switch indices */
         ICON = 0,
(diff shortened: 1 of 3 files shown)


libcore/Enums.vala
```diff
@@ -60,16 +60,46 @@ namespace Marlin {
...
         /* First three modes must match the corresponding mode switch indices */
         ICON = 0,
         LIST = 1,
         MILLER_COLUMNS = 2,
         CURRENT,
         PREFERRED,
-        INVALID
+        INVALID;
+
+        public string to_string () {
```
tintou:
```suggestion
        public unowned string to_string () {
```

jeremypw:
It occurs to me that, if we bump the minimum gtk3 version to 3.22.25  (its currently 3.22.30 in Juno), then we do not need to use a string type action target anymore, we can use the enum value as a target. I'll try it and see.

libcore/Enums.vala
jeremypw: OK, done.

jeremypw:
See #1094 for a better solution.

PR #1093, opened 2019-10-27
Delint libwidgets

jeremypw:
`io.elementary.vala-lint -d [path-to-libwidgets]` now returns only warnings

### DIFF
--- a/libwidgets/View/SearchResults.vala
+++ b/libwidgets/View/SearchResults.vala
@@ -915,12 +915,13 @@ namespace Marlin.View.Chrome {
             return false;
         }
 
-        string ATTRIBUTES = FileAttribute.STANDARD_NAME + "," +
-                            FileAttribute.STANDARD_DISPLAY_NAME + "," +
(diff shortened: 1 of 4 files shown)


libwidgets/View/SearchResults.vala
```diff
@@ -915,12 +915,13 @@ namespace Marlin.View.Chrome {
             return false;
         }
 
-        string ATTRIBUTES = FileAttribute.STANDARD_NAME + "," +
```
tintou: The previous one with:
```
        const string ATTRIBUTES = FileAttribute.STANDARD_NAME + "," +
                            FileAttribute.STANDARD_DISPLAY_NAME + "," +
                            FileAttribute.STANDARD_CONTENT_TYPE + "," +
                            FileAttribute.STANDARD_IS_HIDDEN + "," +
                            FileAttribute.STANDARD_TYPE + "," +
                            FileAttribute.STANDARD_ICON;
```
was more memory efficient

jeremypw: OK, reverted.

PR #1112, opened 2019-11-29
Make create operation use GTask API to use native Vala async support

tintou:
This will also ease the port of the remaining parts to Vala

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -939,19 +939,36 @@ namespace FM {
 
             /* Block the async directory file monitor to avoid generating unwanted "add-file" events */
             slot.directory.block_monitor ();
-            Marlin.FileOperations.new_file (this as Gtk.Widget,
-                                            null,
-                                            parent_uri,
-                                            null,
-                                            null,
-                                            0,
-                                            create_file_done);
+            Marlin.FileOperations.new_file.begin (
+                this as Gtk.Widget,
+                null,
+                parent_uri,
+                null,
+                null,
+                0,
+                null,
+                (obj, res) => {
+                    try {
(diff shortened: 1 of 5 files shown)


src/View/AbstractDirectoryView.vala
```diff
@@ -939,19 +939,36 @@ namespace FM {
...
             /* Block the async directory file monitor to avoid generating unwanted "add-file" events */
             slot.directory.block_monitor ();
-            Marlin.FileOperations.new_file (this as Gtk.Widget,
-                                            null,
-                                            parent_uri,
-                                            null,
-                                            null,
-                                            0,
-                                            create_file_done);
+            Marlin.FileOperations.new_file (
```
jeremypw: `.begin` missing?

src/View/AbstractDirectoryView.vala
```diff
@@ -2411,12 +2428,21 @@ namespace FM {
             /* Block the async directory file monitor to avoid generating unwanted "add-file" events */
             slot.directory.block_monitor ();
             var new_name = (_("Untitled %s")).printf (template.get_basename ());
-            Marlin.FileOperations.new_file_from_template (this,
-                                                          null,
-                                                          slot.location,
-                                                          new_name,
-                                                          template,
-                                                          create_file_done);
+            Marlin.FileOperations.new_file_from_template (
```
jeremypw: `.begin` missing?

jeremypw:
Looks good.  I'll give it some testing.

src/View/AbstractDirectoryView.vala
tintou: oops, indeed, it was written with the ton of warnings :)

PR #1120, opened 2019-12-05
FileOperations: ensure that all functions are async

tintou:
This allows to use the native Vala syntax for async and provides a decent base to get rid of GIOScheduler


tintou:
@jeremypw oops, indeed, I forgot a to change the callback of `marlin_file_operations_copy_move_link` in the undostackmanager, it is now fixed and working :tada:

Issue #1101, opened 2019-11-06
Permanent file deletion dialog should have .destructive-action on the Delete button


Exalm:
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Describe the bug

![Screenshot from 2019-11-06 12-55-07](https://user-images.githubusercontent.com/1510457/68279248-ebc04080-006a-11ea-91ef-081da40c2964.png)

The Delete button does a dangerous action that cannot be undone. I think it makes sense to have it in red.

### To Reproduce
1. Try shift-deleting any flie

- [x] I'm using the latest version from git that I've manually compiled

cassidyjames:
If someone is looking at touching this dialog up, I have some copy thoughts. :)

>#### Permanently delete “Videos”?
>Deleted items are not sent to Trash and are not recoverable.
>
>    `Cancel`  **`Delete`**

mcclurgm:
I'd like to give this a shot, but I would appreciate some design help.

Since the delete dialog uses a generic dialog from the `run_warning` function, we would either have to special-case it somehow or make larger changes to the dialog codebase. It looks like half the functions that call `run_warning` are confirmations of destructive actions and could benefit from a similar change (like delete from trash). But many others aren't (like error messages from copying files). 

My instinct is to create a new type of dialog function like `run_confirmation` that would be for destructive actions like this. I think that this sounds semantically different from a status warning, and the other destructive confirmations could benefit from this fix. Is this a good way to go about a fix here, or is it too big?

jeremypw:
The simplest way would be to apply the destructive-action style to any button titled `DELETE` or `DELETE_ALL` .  You would need to define the button title constants in a way that makes them available to both marlin-file-operations.c and to StockDialogs.vala.  One way to do this is to move the definitions into StockDialogs.vala but outside any namespace (if you namespace them you would need to amend their use in marlin-file-operations.c accordingly).

mcclurgm:
That makes sense. I have one reservation about this method though. "Delete" and "Delete All" are not the only actions that are destructive and could use this. One example is "Empty Trash," which isn't given a constant. I could of course create a constant for it. But this begs the question of whether we want to define a constant for every possible destructive button label, and then check for every one of them.

jeremypw:
[...]

PR #1131, opened 2019-12-22
Show warning dialog on attempt to open file in trash

### LINKED ISSUE #1129: Trying to open file in trash should display error message
## Feature

Trying to open a file in the trash simply doesn't do anything in the GUI, providing no feedback of what went wrong or how to fix it. I think this could potentially leave users confused, or believing that the app isn't responding to their actions. 

The action does produce an actionable error message--but only in console output:
> ** (io.elementary.files:22835): WARNING **: 09:54:09.782: AbstractDirectoryView.vala:835: Cannot open file in trash

I think that it would be helpful to include a dialog or banner of some sort to indicate this in the app's GUI. 

**Existing work**

One possible approach would be to follow the Finder's design. It opens a dialog explaining the error and suggesting how to proceed if the user actually wants to open the file.

Image 12-21-19 at 10 05 AM

I'm not sure if this is a place for it, but elementary has infobars that could be used to convey the same information. That's a possible alternative design we could use.

jeremypw:
Fixes #1129 

Problems opening/executing a file not in trash already show a warning dialog so for consistency a dialog should be shown when in trash.


danrabbit:
@cassidyjamea are we using typographical quotes for this?

jeremypw:
Looking at Cassidy's suggestion for #1101 it looks like double quotes should be used.

jeremypw:
I have now used typographical double quotes (U+201C and U+201D).

getsnoopy:
It seems counter-intuitive to allow a user to open a file from the Trash, only to tell them that it's not possible. Wouldn't it be better to just dim the items when viewed in the Trash with a banner at the top saying opening is not possible and/or open their properties dialog when a user tries to open them?

jeremypw:
@getsnoopy  The context menu in trash already leaves out impossible/inappropriate actions.  This PR was to fix the issue raised that if you use mouse (double-click) or keyboard  (select + Space) then nothing happens and no explanation is given.

jeremypw:
Dimmed items are already used for hidden files.  Using a banner would take up space.  Having two different results for the same action on the same thing would probably be objected to by the design team.

getsnoopy:
In that case, I think a banner at the top would be the best. It wouldn't really take up much space (just a line of text), and it wouldn't have to pop up with the same information in a dialog over and over (that the user can't open a file).

jeremypw:
I would wait for @cassidyjames opinion first - he may have other ideas and UX team approval would be needed before merging.

cassidyjames:
I think an infobar could make sense here since it applies to all the items in the trash; I don't see the advantage to waiting until the user tries to do something to tell them they can't.

src/View/AbstractDirectoryView.vala
```diff
@@ -827,7 +827,12 @@ namespace FM {
                     }
                 }
             } else {
-                warning ("Cannot open file in trash");
+                PF.Dialogs.show_error_dialog (
+                    ///TRANSLATORS: '%s' is a quoted placehorder for the name of a file. It can be moved but not omitted
+                    _("The file “%s” cannot be opened because it is in the Trash").printf (file.basename),
+                    _("To use this file, first drag it out of the Trash"),
```
danrabbit: This copy should probably say something more about "restore" the item instead of being too specific with a drag operation since there are multiple ways to restore a trashed file

jeremypw: Strictly speaking you do not have to "restore" the file to its original location - just move it out of the trash.

jeremypw:
> I think an infobar could make sense here since it applies to all the items in the trash; I don't see the advantage to waiting until the user tries to do something to tell them they can't.

On the other hand, once the user knows they cannot perform this action, the infobar is wasting space - unless you have a "Do not show again" checkbox - and that means another setting. 

@danrabbit  What is your opinion?

cassidyjames:
We had a discussion on Slack and came to a potential solution. I forgot we had planned to use an infobar for Housekeeping settings anyway, and yeah, maybe it's not the best use of space here. Instead, we could offer a more useful dialog that has a suggested action, instead of just an error.

>### “filename.ext” must be restored before opening
>
>Files cannot be opened from the trash. Restoring “filename.ext” will move it back to [/ugly/path/ugh/gross/](#).
>
>`Restore` &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; `Cancel`  **`Restore and Open`**

Some notes on the path there:
- Ideally it could be a link to open the location so the user can see where the file will end up
- If the path in in the user's home directory, we should make it relative to their home. i.e. `/home/cassidyjames/Pictures/Screenshots/` could read as [Pictures/Screenshots](#), `/home/cassidyjames/Downloads` could just read as [Downloads](#).

danrabbit:
Yeah as I'm thinking about it, I wonder who would want to restore here but then not open the file. That seems unlikely

jeremypw:
Another option would be for `Restore` (or another button) to open up the file-chooser in `Save As` mode with the original name and folder (if it still exists - not sure where otherwise) This would give the opportunity to choose both the location and name of the restored file.  The problem comes when you have a selection of files with different origins (or Restore All).  I guess there is little alternative to move them all back to their original locations (although the user may not necessarily remember where that was).  If the files are of different types then `Restore and Open` would not apply.

jeremypw:
I think `Restore and Open` is fine when the dialog shows in response to an attempt to open a file (but will fail if the original location no longer exists)

danrabbit:
@jeremypw Maybe we could show "Save as" if the original location no longer exists, but showing it always feels a little convoluted.

It's kind of a curious case of why someone is opening a file in the trash anyways. If we assume that they either deleted the file by accident or deleted it and then realized they still need it, I'm not sure that in either of those cases the original location of the file is called into question unless it doesn't exist anymore.

For example, Ellen just finished her final paper for Econ 101 and has mailed it off to her professor. She decides to delete the folder she has created for the class because she's finally done. 15 minutes later she realizes she made a mistake and needs to send in a correction.

In this case, since the original folder was deleted, I think it makes sense to ask where she wants to move the file to. However, if she had just deleted that one paper, I would imagine it makes sense to just put it back in her Econ 101 folder and we probably don't need to ask.

@cassidyjames  ^

cassidyjames:
@danrabbit ah yeah that makes sense; I agree that "save as" only really makes sense if the original location has been deleted or is otherwise unreachable. Do we _just_ want to handle restoring in this case, or also opening still? It's a good point regarding multiple file selection; opening those could be super annoying—if we still have "Restore and Open" for a single file (which I think would be super convenient), we could just have a "Restore" suggested action for multiple file selections?

jeremypw:
I think "Restore" or "Restore and Open" are fine when the dialog shows where the file will restored. At the moment this is not the case for using the context menu but I guess that is a different issue.  You can find out the original location from the PropertiesWindow although that is not very obvious.  A selection of files with multiple origin gives "Unknown" in the Properties Window.

Only showing the "Save As" dialog when the origin is not accessible is OK by me.

I guess we are dealing with corner cases here so it may not be worth making the solution too complicated.

jeremypw:
Blocked for release of v4.4.0

jeremypw:
Having the restored file open automatically is not as straightforward as it seems since the move file operation happens asynchronously.  Working on it (on and off).

jeremypw:
Just restoring the file(s) is easier.

PR #1133, opened 2019-12-23
Fix drag out of archive

### LINKED ISSUE #1082: Dragging file out of archive breaks subsequent internal DnD

jeremypw:
Fixes #1082 

Also some simplification, better variable names, comments.

Significant changes are:

* Return success even if XDnDDirectSave returns "E"
* Separate clearing of source and drag data
* Process "drag-leave" and "drag-drop" in correct order

Other changes should only be code cleaning - lose unneeded variables, rename some variables for clarity, added comments, reorder some lines for efficiency.

### DIFF
--- a/libcore/DndHandler.vala
+++ b/libcore/DndHandler.vala
@@ -200,6 +200,7 @@ namespace Marlin {
                     case 'E':
                         /* No fallback for XdndDirectSave stage (3), result "E" ("Error") yet.
                          * Note this result may be obtained even if the file was successfully saved */
+                        success = true;
                         break;
                     case 'S':
                         /* XdndDirectSave "Success" */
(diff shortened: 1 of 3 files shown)


tintou:
It really looks cleaner that way

PR #1140, opened 2020-01-02
Drop colorization and highlighting of focused and hovered icons

jeremypw:
As discussed with @danrabbit  on Slack and partially resolving #473 

The background for selected/focused cells is also made the same as for TreeView.  The only problem with this is that neither view indicates where the keyboard focus alone is - when the cell is unselected it also loses the "focus" selector for some reason.  This is not much problem in practice as the items are auto-selected when the cursor lands on them.


danrabbit:
It looks like with this branch there's a small inconsistency between views. The column and tree view seem to have the same styling and look as expected, but the icon view doesn't seem to distinguish between selection and keyboard focus. Compare for example:

Screenshot from 2020-01-03 18 07 24@2x
Screenshot from 2020-01-03 18 07 29@2x
Screenshot from 2020-01-03 18 07 17@2x

jeremypw:
@danrabbit :  I have looked into this and there does not seem to be a simple way to fix it using just css.  It is a feature of IconView (which draws the background cells) caused by it (a) not having a css selector  "row", which GtkTreeView uses for coloring the background of selected but unfocused rows and (b) the "focus" selector being "true" for all IconView background cells regardless of whether they actually have keyboard focus (it seems to relate the whole view having keyboard focus.  This contrasts with the "hover" selector which does change on an individual cell basis.  This may be a bug in GtkIconView.  

The only solution I can think of is make the Icon and Text cells fill the whole of the background cell and get them to imitate the appearance of GtkTreeView using custom css per renderer.

danrabbit:
@jeremypw Ah okay. Yeah I was having trouble with trying to do it myself. For just the grid view, what about making the big background grey always and only drawing the blue text background on the focused element? Is that possible?

jeremypw:
Yes, it is possible for the style of the icon renderer and/or the text renderer to track the keyboard focus.  We could probably also tweak the styling of the TreeView to match that of IconView if you wanted.

jeremypw:
@danrabbit The problem with using a blue text background to show cursor position is that it conflicts with color-tagging,  especially since one of the color-tags is the same color as the selected background.  Do you have any other suggestion? E.g. drawing a frame or underline.

jeremypw:
PR appearance:

Icon View: First item has focus, second item hovered, first two items selected.
Screenshot from 2020-01-23 10 21 54

jeremypw:
Blocked for release of v4.4.0

PR #1149, opened 2020-01-10
Remove `file://` from path shown in pathbar; update comments

### LINKED ISSUE #1146: file:/// prefix in folder path
## Prerequisites
- [+] I have searched open and closed issues for duplicates.

## Describe the bug
In version 4.3.0, an inconvenience appeared: instead of the usual folder path, the path with the prefix file:/// is displayed.

### To Reproduce
1. Go to any folder. I will cite /home/platon/miniconda3/bin as an example.
2. Click at the right of the breadcrumbs.
You will see the path like file:///home/platon/miniconda3/bin

### Expected behavior
In previous versions, the path began with /home. It was much more convenient if you needed to copy the path to a terminal or another program.

## Platform Information
elementary OS 5.1

Please check what applies:
- [+] I'm using the latest released stable version

jeremypw:
Fixes #1146 

The relevant functions now default to include the `file://` protocol to avoid possible regressions, but when being called for displaying the path in the pathbar remove this protocol, if present.  Also remove it from paths being pasted or dragged into a window that accepts text.

Comments are updated to reflect current behaviour


jeremypw:
Blocked for release of v4.4.0

PR #1150, opened 2020-01-10
Add destructive button to file operation dialogs

### LINKED ISSUE #1101: Permanent file deletion dialog should have .destructive-action on the Delete button

mcclurgm:
Fixes #1101. 

--
Original comment:
Design is in progress. I also need to add more button types to get destructive style. (Suggestions are welcome! Currently planning on "Empty Trash.")


libcore/StockDialogs.vala
```diff
@@ -88,7 +101,12 @@ namespace PF.Dialogs {
         } else {
             var response_id = 0;
             foreach (string title in data.button_titles) {
-                dialog.add_button (title, response_id++);
+                dialog.add_button (title, response_id);
+                if (title == DELETE || title == DELETE_ALL) {
+                    var button = dialog.get_widget_for_response (response_id);
+                    button.get_style_context ().add_class ("destructive-action");
```
danrabbit: It's better to use the constants for these so we get deprecation warnings in case they ever change or are removed :)

```suggestion
                    button.get_style_context ().add_class (Gtk.STYLE_CLASS_DESTRUCTIVE_ACTION);
```

mcclurgm: Good to know, thanks! I added that.

mcclurgm:
Come to think of it, I think some of the other constants could be destructive: `REPLACE` and `REPLACE_ALL` (although they don't seem to be used anywhere?). Should these also be considered destructive actions? I could add the style to them as well, but it would make the if statement rather long.

jeremypw:
I just triggered a REPLACE? dialog by moving a file into a folder which already had a file with the same name.  It already had destructive-action styling (see FileConflictDialog.vala).  I don't think there is a REPLACE_ALL button (there a checkbox for applying an action to all conflicting files)

mcclurgm:
Sorry I should have been more specific--I meant that the REPLACE actions don't seem to be used in `marlin_file_operations.c`. As far as I can tell, it got moved to `FileConflictDialog.vala` and picked up the destructive styling in that class. So it shouldn't be my problem to implement destructive-styled REPLACE actions (at least I don't think).

mcclurgm:
I just added an `EMPTY_TRASH` case. I think that's it for destructive actions, unless we want to add `REPLACE` and `REPLACE_ALL` anyway. So I think this is ready for a real review.

This is only very tangential, but I don't think any of these actions (`EMPTY_TRASH`, `REPLACE`, `REPLACE_ALL`) are handled in `marlin_file_operations.c` anymore, and could potentially be written out of the codebase. This is based only on searching the repo for function calls, but I don't see where they are in use anywhere. Conflicts have been moved to their own Vala class and the trash seems to be a plugin now. Of course this isn't an issue for this PR, but it is something I noticed. If they're removed, I could get rid of the relevant cases in `StockDialogs.vala`.

mcclurgm:
Great, makes sense. Would it be worth opening up a new issue for this?

jeremypw:
@tintou is already currently working on cleaning up and porting marlin-file-operations so I do not think a separate issue is required, thanks

PR #1152, opened 2020-01-11
File copy remaining time fix

markodp:
Copy time was not displayed properly when bigger than 60 seconds. Time unit was always displayed in seconds .. for example "10 seconds" was shown instead of "10 minutes".

I moved seconds_count_format_time_units() function to be used only for translations, which to me makes more sense.


jeremypw:
I agree the existing code is questionable, not least in the names of the functions.  As the two functions are similar and are used in combination, would it make sense to combine them and have one sensibly named function that returns two results (1) a string containing the time and units and (2) a bool indicating whether the string should be regarded as plural?  One of the results would have to be an `out` parameter.

tintou:
@jeremypw you can't treat plural/singular with a bool unfortunately. It's why `ngettext` exists ;) some languages require more than two cases :)

jeremypw:
@tintou Oh, thanks - I should have checked,  In that case the out parameter would have to be `ulong`.

markodp:
Great idea to combine them, thanks.

On another note, i am not really sure about the part where there's "hours + minutes". This does not makes sense to me, even in english. For example, if i use the output in context:
`there ARE '1 hour, 1 minute' left` but should be `there IS '1 hour, 1 minute' left`
or even
`there ARE '1 hour, 50 minutes' left` should be `there IS '1 hour, 50 minutes' left`
(correct me if i'm wrong, since english is my second language)

So i think in this case, plural form is dependent on hours and function should return only hours.

I know that in code there is no first part, where it makes difference, only " left" is appended after time. But it might be significant in some other languages.

Thoughts?

jeremypw:
I think you are right about hours + minutes - I did not understand it either and I am a native English speaker!  Maybe in some other language the verb form depends on the sum of hours and minutes?? In anycase I would probably phrase it differently such as "The time remaining is <time string> " or "<time string> until complete", which, in English, definitely does not depend on the number of hours or minutes.

I think that you should just fix the units problem in this PR and then submit another PR addressing the singular/plural problem after consulting with 118n.

markodp:
In that case, PR is ready for review.

PR #1160, opened 2020-01-15
AbstractDirectoryView: Create context menus natively

### LINKED ISSUE #629: Clean up context menu
The current context menu is in a seemingly random order with separators everywhere, effectively separating everything and making it look really cluttered.

screenshot

Instead, I think we should rethink the groupings of items into a few categories:

**Act on the currently-selected or being-clicked-on thing**
- Open in
- (Contractor items)
- Set a color
- Properties

**Act on the current folder**
- New
- Sort by
- Bookmark

**App-wide**
- Show hidden files    

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/63742373-clean-up-context-menu?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #629 

This branch doesn't change the position of the contractor or color tags items because they are currently plugins and that means they just get shoved at the end of the menu. I'd like to revisit this in a future branch.

I broke most of the submenus into their own private classes, but making the whole menu its own class seems like it's going to be a lot more work. This is a monster class so I think it would be worth it, but maybe not for this PR.

I also chose not to tackle the Granite.AccelLabel in this branch because it's already pretty big

I think there's several items that should be insensitive instead of not packed, but that also feels like it can wait for another branch since it will likely be significant work.

This will need to be tested:
- [ ] In local folders
- [ ] In remote folders
- [ ] In the Trash
- [ ] In Recent
- [ ] In read-only folders
- [ ] With stuff on the clipboard
- [ ] With an empty clipboard
- [ ] With folders/files that have already been bookmarked
- [ ] With folders/files that have not yet been bookmarked


danrabbit:
@cassidyjames it looks like it would be non-trivial to invert the logic for "Hide Thumbnails". I think that would be a good candidate for a separate branch

cassidyjames:
@danrabbit that's fair. Then we should instead restore the :ballot_box_with_check: behavior to match with master at least.

danrabbit:
Filed the inverted logic as https://github.com/elementary/files/issues/1163

src/View/AbstractDirectoryView.vala
```diff
@@ -1834,364 +1837,396 @@ namespace FM {
...
+                } else if (in_recent) {
+                    menu.add (new Gtk.SeparatorMenuItem ());
+                    menu.add (new SortSubMenuItem ());
+                    menu.add (new Gtk.SeparatorMenuItem ());
+                    menu.add (show_hidden_menuitem);
+
+                    if (slot.directory.is_local) {
+                        menu.add (hide_local_thumbnails_menuitem);
+                    } else if (slot.directory.can_open_files) {
+                        menu.add (show_remote_thumbnails_menuitem);
```
danrabbit: Is it reasonable to think the recent folder will ever be remote?

cassidyjames: I'm not familiar enough with how that's implemented, but I don't think the entire directory could be remote… it's a virtual directory. I guess maybe theoretically items inside of it could be remote?

danrabbit: Yeah this is more a question for @jeremypw. My guess is that this is unreachable code but I'd like to ask before removing it

danrabbit:
@jeremypw oddly the radiomenuitems don't seem to be triggering the action at all. But changing them to check items that draw as radios seems to fix the issue

jeremypw:
Yes, I have a feeling that maybe RadioMenuItems need separate actions, not detailed actions (?).  If you create the RadioMenuItems separately (ungrouped) then they all appear as active (ticked), whereas doing exactly the same with CheckMenuItems shows the state correctly.  Possibly a bug in Gtk?   I guess we will have to use the work around for now.  I'll spend a little more time testing this (probably tomorrow) then it should be mergeable.

jeremypw:
@danrabbit  Is it intended to move to AccelLabels at some point?

PR #1184, opened 2020-01-23
Recognise afc protocol

### LINKED ISSUE #1119: [Juno + Hera] Files does not mount iPod in the desired way (wrong protocol)
## Prerequisites
- [X] I have searched open and closed issues for duplicates.

## Describe the bug
When mounting an iPod (in this case, 2nd-gen iPod Touch, 2009), Files uses the `gphoto://` protocol, mounting it as a Photo device (only DCIM folder is visible), instead of using the `afc://{ipod_uuid}` protocolol, which makes music files accesible.

### To Reproduce
1. Connect iPod
2. Mount iPod

3. Access iPods's file panel
4. Check protocol on the address bar (e.g. `gphoto2://[usb:002,004]/`)

5. Open Terminal and type `idevicepair pair`

6. Copy the iPod's resulting UUID
7. In Files, replace `gphoto://... ` with `afc://{resulting UUID number}`
8. See that the iPod's actual filesystem is now accesible

### Expected behavior
Connecting the iPod should result in Files using the `afc://` protocol to sync music with Noise or Rhythmbox instead of mounting it as a camera using the `gphoto://` protocol.

## Platform Information
 - OS: elementary OS
 - OS Version: Juno
 - Hardware info:
   * System: ThinkPad T410s
   * iPod: 2nd-gen Touch

- [X] I'm using the latest released stable version

jeremypw:
Fixes #1119 

I am not sure whether this is enough to fix the linked issue - request iPod user to test.


jeremypw:
Blocked for release of v4.4.0

danrabbit:
It sounds like from the [latest comment](https://github.com/elementary/files/issues/1119#issuecomment-633798339) that this branch doesn't fix the linked issue. Gonna convert to draft until we can confirm the fix

jeremypw:
@danrabbit Thanks - looks like I forgot to deal with the odd `:3/` ending to the uri  produced by this protocol and gvfs.  It is a valid uri so sanitizing does not remove it.  It works OK as a folder name suffix on a local drive so we do not want to strip it arbitrarily.  I guess just stripping it if the protocol is `afc` and the file is a folder will do it - a bit of a hack though.

fabiozaramella:
It works, but devices get duplicated in the sidebar. I can't attach a screenshot at the moment but testing with an iPad results in the iPad being recognized twice, once as a camera and once using afc protocol, both under the Devices section. Also, it get recognized with afc also under the Network section.

jeremypw:
@fabiozaramella Thanks for looking at this, I needed an Apple user to test it. Which category do think is correct for the AFC protocol?  I would have said "Devices", in which case I do not think it is a bug for it to appear twice under that category, once for each supported protocol.  I do not have an iPhone but presumable the locations accessible via each protocol differ?  The camera protocol is is presumable "PTP" or "MTP"?

I'll look into why the AFC protocol is appearing under two different categories which is a bug.

fabiozaramella:
> @fabiozaramella Thanks for looking at this, I needed an Apple user to test it. Which category do think is correct for the AFC protocol? I would have said "Devices", in which case I do not think it is a bug for it to appear twice under that category, once for each supported protocol. I do not have an iPhone but presumable the locations accessible via each protocol differ? The camera protocol is is presumable "PTP" or "MTP"?

It seems to be mounted as a photo camera using the gphoto2 and as an external memory using afc, but yesterday it showed a bunch more of files using the same protocol and today just the Chrome folder, i don't know why...

> I'll look into why the AFC protocol is appearing under two different categories which is a bug.

Also I can't reproduce this today, as you can see from the screenshots below. It seems to happen randomly.

iPad as afc
Schermata del 2020-11-06 19 04 51

jeremypw:
> Also I can't reproduce this today, as you can see from the screenshots below. It seems to happen randomly.

Does it make a difference whether or not the afc connection is mounted/ejected when Files starts?  I can add code to ensure that the mounts shown under "Network" are not already showing under "Devices".

fabiozaramella:
> Does it make a difference whether or not the afc connection is mounted/ejected when Files starts? I can add code to ensure that the mounts shown under "Network" are not already showing under "Devices".

How do you want me to test it? Just to be sure 😅

jeremypw:
I suggest:
* Set Files to just show "Home" tab and close.
* Plug in iPad  
* Start Files - the "Documents on iPad" device should not be mounted (no "eject button).   Does Network also show "Documents on iPad"?
* Mount "Documents on iPad" by clicking on the Device. Location displays in view. Does Network now show "Documents on iPad"?
* Close Files and then open Files again
* "Documents on iPad" should still be mounted. Does Network now show "Documents on iPad"?

To be honest, it may be better for me to address this in the new ListBox Sidebar (which did not exist when this PR was first written).  Even better someone with an Apple device :stuck_out_tongue_closed_eyes:

fabiozaramella:
> * Set Files to just show "Home" tab and close.

- [x] Done 

> * Plug in iPad

- [x] Done 

> * Start Files - the "Documents on iPad" device should not be mounted (no "eject button).   Does Network also show "Documents on iPad"?

- [x] No, it doesn't show up. This is the result when i plug in the iPad but i don't mount it: 
iPad plugged in but not mounted
> * Mount "Documents on iPad" by clicking on the Device. Location displays in view. Does Network now show "Documents on iPad"?

- [x] No, this is what i get after i mount it:
Documents on iPad  is mounted

> * Close Files and then open Files again

- [x] Done

> * "Documents on iPad" should still be mounted. Does Network now show "Documents on iPad"?

- [x] It is still mounted but "Documents on iPad" doesn't persist under Network as it used to do before, idkw. Still it shows up for a fraction of a second, I've tried to record it using Peek but it doesn't show up in the final GIF...
 
> To be honest, it may be better for me to address this in the new ListBox Sidebar (which did not exist when this PR was first written). Even better someone with an Apple device

@jeremypw Tell me if you need some more testing 😅️

jeremypw:
@fabiozaramella Thanks for the testing.  So, it seems that under your testing,  the afc protocol does not appear both under Devices category and the Network category except momentarily.  This sounds like a timing issue to do with various signals from the VolumeMonitor.  The Sidebar gets updated several times during startup as different Drives, Volumes and Mounts get reported by the monitor.  I thought that I had throttled this updating at some point but looking at the current code this does not seem to be the case - presumably that was in a PR that was never merged.  I'll probably include that in the ListBox Sidebar PR.  I'll need to port the AFC protocol stuff there as well.

I think I'll close (but not delete) this PR for now.  I'll let you know when the AFC stuff needs testing in the ListBox Sidebar.  Thanks again for your help.

jeremypw:
Reopening as this does not actually touch Sidebar code.

jeremypw:
As this does not actually touch Sidebar code and at least partially fixes the issue I think it is still worth considering merging it and the fixing Sidebar issues in a separate PR.

fabiozaramella:
> As this does not actually touch Sidebar code and at least partially fixes the issue I think it is still worth considering merging it and the fixing Sidebar issues in a separate PR.

Nice, so is it ready for review? ☺️

jeremypw:
Yes, I think it can be reviewed again now with the narrow objective of mounting AFC devices so that they are usable even if not in the expected place in the sidebar.

fabiozaramella (approved):
@jeremypw I'm not opposed to merge your changes and deal later with any problem that may arise. 👍🏽️

jeremypw:
Thanks - there won't be a stable Files release for a while and I do not think this will cause any regressions for those not experiencing the linked issue and will improve things for those who are.  I can pull this into the ListBox Sidebar branch and work on any glitches there.

PR #1247, opened 2020-02-29
Do not update pathbar after loading

### LINKED ISSUE #761: Breadcrumbs sometimes incorrect at startup
* Arrange for two tabs to be restored, one showing a folder that is slow to load (e,g. large number of files) and one that is quick to load.  Make the quick one the second tab and make it active when Files is closed.

* Re-open Files.  The breadcrumbs will show the the path of the slow folder even though it is not the active tab.

This is because the 'done_loading' signal from the slow folder is causing the pathbar to update even though it is coming from a slot that is no longer active.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/66270594-breadcrumbs-sometimes-incorrect-at-startup?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fix #761 

A simple solution.  It is theoretically possible that under rare circumstances where the final location of the view is not the same as the requested location the pathbar may be incorrect but I have not been able to find a way of triggering this.  In any case, this seems preferable to a bug occurring on many startups for many people and can be fixed with a more complex solution if the need should arise.


cassidyjames:
@jeremypw was this a regression in 4.4.0? or a pre-existing issue?

jeremypw:
@cassidyjames  Thanks for the review. This is a regression but has been around quite some time - I think from when the loading of the tabs was made asynchronous.

jeremypw:
Blocked until v4.4.1 released

PR #1253, opened 2020-03-04
Fix view frozen after renaming with blank

### LINKED ISSUE #1252: After creating a new file/folder without some characters in the input field you cannot delete it, context menu does not work either
## Prerequisites
- [x] I have searched open and closed issues for duplicates…and did not find a similiar one.

## Describe the bug

In Files 4.4.1 you can not delete a new folder, when you just create it without some characters in the input field. Context menu also does not work as long as you not restart io.elementary.files.

### To Reproduce
1. Create a new file/folder with context menu or keyboard shortcut and make sure it does not have a name (delete all characters in the input field) and press enter...now the name is 'New file' or 'New folder'
2. Now you cannot delete this new file or folder, only a restart of files helps.

### Expected behavior
Deleting the new file/folders works as expected with [DEL] or with the context menu.

Please share a screenshot of the System Settings > About screen.
Bildschirmfoto von 2020-03-04 18-27-11

Please check what applies:
-->

### LINKED ISSUE #1255: Strip filename on rename
Brought up in #1253 

Right now it's possible to create a name with leading and trailing spaces, which seems to be far more likely to be accidental (and problematic) than useful. I'm not sure which filesystems would allow this, but I think we could safely disallow it in Files—if a user needs it (and knows what they're doing), they could use Terminal.

jeremypw:
Fixes #1252 
Fixes #1255 

Note that this also prevents a user naming a file with only spaces even though this seems to be legal with some filesystems.   It also removes any leading or trailing space. Should the user be restricted in this way?  If not then the `.strip ()` functions can be removed.

Some further cleanup of renaming is possible but left for another PR.

### DIFF
--- a/data/io.elementary.files.appdata.xml.in.in
+++ b/data/io.elementary.files.appdata.xml.in.in
@@ -62,6 +62,8 @@
       <description>
         <ul>
           <li>Fix device icon sometimes missing</li>
+          <li>Fix occasional view freeze after renaming</li>
+          <li>Improve renaming logic when dealing with leading/trailing whitespace</li>
           <li>Updated translations</li>
         </ul>
       </description>
--- a/src/Dialogs/PropertiesWindow.vala
+++ b/src/Dialogs/PropertiesWindow.vala
@@ -322,7 +322,7 @@ public class PropertiesWindow : AbstractPropertiesDialog {
         /* Only rename if name actually changed */
         original_name = file.info.get_name ();
 
-        if (new_name != "") {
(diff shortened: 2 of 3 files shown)


cassidyjames:
@jeremypw yeah I think it's reasonable to disallow beginning, trailing, and only-spaces in the app even if they may be valid on some filesystems. There's more potential for confusion/breakage than need to allow such a niche naming scheme. There's always Terminal if someone really needs it and knows what they're doing. :)

jeremypw:
@cassidyjames Agreed.  I have modified this PR to also fix #1255 now as little extra code was required.

PR #1262, opened 2020-03-17
Sidebar: Do not assume that every Icon is a ThemedIcon

tintou:
A CD-ROM can actually return a FileIcon


jeremypw:
@tintou: Good spot!  Is a CDROM the only case returning a FileIcon?  In which case, should a different fallback be used?

tintou:
I already saw this behavior with a simple USB drive (with the default content)

jeremypw:
OK, its not obvious how to choose a better default icon and the important thing is to prevent a crash.

PR #1284, opened 2020-04-16
Exclude fuse filesystems from git plugin scanning to prevent apparent hanging

### LINKED ISSUE #1156: Opening folder mounted with sshfs that contains a git repo causes long load times
I have a folder I mount with `sshfs` so I can edit the files on my local machine. Trying to open the folder causes files to hang. It eventually loads, but I really only want git status on files that are physically located on my local hard drive.

### To Reproduce
1. Mount a folder containing a git repo with `sshfs`
2. Open folder in files
3. Wait for files to scan
4. Keep waiting
5. Give up and just use the terminal

### Expected behavior
This is expected with the current implementation, but I think the option to disable the git status in a directory would be nice.

## Platform Information
image

jeremypw:
Fixes #1156 

This excludes all FUSE file systems (as well any other filesystem whose type cannot be determined) from git plugin scanning.  Is this too draconian?


vjr:
This may be good for SSHFS but draconian for NTFS?

vjr:
(posted comment too quickly)

Also, is there another way of distinguishing between remote versus local filesystems that might be more appropriate to exclude? NTFS being local should be fine for git plugin speed, no?

jeremypw:
@vjr True, but as yet I have not found a programmatic way of distinguishing sshfs and ntfs - both have the FileAttribute.FILESYSTEM_TYPE set to "fuse".  Hopefully someone will advise a way to do it.

Another possibility I thought of was to only exclude folders under `/mnt` and require users to mount problematic folders there and local ones elsewhere but this does not seem satisfactory.

jeremypw:
Is it unreasonable for the plugin to not support git repos on fuse type filesystems? It does not prevent their use - you just do not get the status info in Files.  Presumably most devs would clone repos onto a local native filesystem.   I do not think the remote dir in the linked issue had a repo in it but I'll check.

jeremypw:
I also tried using FileAttribute.FILESYSTEM_REMOTE but this seemed to return `true` only where the URL included a remote protocol (which Files already excludes) - this returned `false` for a remote mount mounted with sshfs on `\mnt`

vjr:
> Is it unreasonable for the plugin to not support git repos on fuse type filesystems? It does not prevent their use - you just do not get the status info in Files. Presumably most devs would clone repos onto a local native filesystem. I do not think the remote dir in the linked issue had a repo in it but I'll check.

I guess this seems reasonable.

jeremypw:
The issue remote_dir did have a repo on it but the reporter was satisfied to clone it to a local drive to circumvent the problem.

jeremypw:
I think it best to avoid locking up the interface even if it means (at least temporarily) restricting the capabilities. Code has the same problem.

plugins/git/plugin.vala
```diff
@@ -89,6 +90,29 @@ public class Marlin.Plugins.Git : Marlin.Plugins.Base {
             return;
         }
 
+        FileInfo info;
+        try {
+            info = directory.get_target_location ().query_filesystem_info (FileAttribute.FILESYSTEM_TYPE);
```
tintou, line 95: Ideally you can put everything inside the try {} block so that it is clear what it is required for
```suggestion
            var info = directory.get_target_location ().query_filesystem_info (FileAttribute.FILESYSTEM_TYPE);
            if (!info.has_attribute (FileAttribute.FILESYSTEM_TYPE)) {
                debug ("GIT PLUGIN: no filesystem type info - ignoring");
                return;
            }
            
            unowned string fs_type = info.get_attribute_string (FileAttribute.FILESYSTEM_TYPE);
            if (EXCLUDED_FS_TYPES.contains (fs_type)) {
                warning ("GIT PLUGIN: excluded filesystem type %s", fs_type);
                return;
            }
```

plugins/git/plugin.vala
```diff
@@ -89,6 +90,29 @@ public class Marlin.Plugins.Git : Marlin.Plugins.Base {
...
+        }
+        catch (GLib.Error error) {
+            if (!(error is IOError.CANCELLED)) {
+                warning ("GIT PLUGIN: Error querying %s filesystem info: %s", directory.uri, error.message);
+            }
+
+            info = null;
```
tintou, line 101: Info will be set to NULL on error so we can return directly as we already know that we will return later on
```suggestion
            return;
```

jeremypw: Yes, good point, you don't need the debug message in this case as you will get the warning message anyway.

PR #1288, opened 2020-04-18
Show overlaybar in all view modes.

### LINKED ISSUE #1285: File info popover does not show if files are in List View
## Prerequisites

## Describe the bug
When using files with list view, if I hover over a file it does not show the file information on the bottom right of the window like what happens with the other views. 

imagem

### To Reproduce
1. With list view active, go to any file
2. Hover over the file
### Expected behavior

Should appear the file information on the bottom right of the window.
## Platform Information
imagem

## Additional context
Could be usefull to show some more information about the file under "Properties".
imagem
This is the properties window from an image file. Why not show the width and height for example?

jeremypw:
Fixes #1285 

The overlaybar shows extra information compared to list view columns and there seems no clear reason to suppress it.

### DIFF
--- a/data/io.elementary.files.appdata.xml.in.in
+++ b/data/io.elementary.files.appdata.xml.in.in
@@ -63,6 +63,7 @@
         <p>Minor updates:</p>
         <ul>
           <li>Fix an issue that prevented file modification times from showing</li>
+          <li>Show file info overlay in List View as well</li>
           <li>Updated translations</li>
         </ul>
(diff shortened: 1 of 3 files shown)


tintou:
Is there anything else using `showbar` ? If not you can maybe remove it completely

jeremypw:
@tintou Not at the moment - I was being conservative - but I guess it can always be reintroduced if the need arises.

PR #1289, opened 2020-04-19
Plugin: Files Preview

marbetschar:
This pull request attempts to eventually add a quick file preview to elementary Files. To accomplish this goal, the plugin leverages the DBus Service provided by the Nautilus File Previewer aka "Sushi".


marbetschar:
As of now this PR contains a quick proof of concept, adding file preview functionality using the Sushi DBus Service as well as the Files plugin facility. There is still work to do in terms of Design and Usability:

Files-Preview

cassidyjames:
The problem with previewers is that we already do thumbnailing, and opening Photos or Code or whatever the “real app” is almost always ends up being just as fast as or faster than opening a previewer app that has to be designed to preview any arbitrary file. 

A previewer app feels like a weird and ultimately fruitless workaround for slow apps, and implementing it is destined to turn into some sort of “handle all the things” app instead of letting individual apps handle their own file formats.

marbetschar:
I understand your line of thought and I'm aware you guys held this discussion over and over already. So I did not assume this pull request to be merged anyways and I'm fine with this :)

But is there a way we can add support for defining a keyboard shortcut to a plugin which provides a `context_menu` entry?

marbetschar:
I love to experiment a bit more with the sushi file previewer to see where I end up with because I fell in love with the Preview App on macOS back then. But this seems to be only worth it, if its somehow possible to trigger the (3rd party) plugin with a keyboard shortcut.

jeremypw:
Can an accelerator be added to the context menu widget?

marbetschar:
@jeremypw I did not found a way to bind an accelerator to a plugin function along with the possibility to retrieve the currently selected files.

jeremypw:
@marbetschar I have looked into this a little more and there a number of problems to be overcome to enable the preview to triggered by a shortcut.
It is possible plugins to set an accelerator key for a one of their functions but it needs adding to master an application level action as well as another hook for plugins (afaict).  

A more difficult issue is obtaining what file (if any) is selected when the shortcut is triggered.  The plugin has no way of knowing this when the context menu is not showing because it has no access to any symbols define in `\src` (and vice versa). Moreover, for performance reasons the selection information is only updated when needed, not when the selection changes so at the moment it has to be requested.  Updating the plugins with the current selection everytime it changes may reduce performance for large selections.  I'll see if I can think of an acceptable simple solution.

marbetschar:
@jeremypw thanks a lot!!!

jeremypw:
@marbetschar  I have something working up to calling the previewer.  Do you know what to put in the `windowHandle` parameter of the `show_file` method?

jeremypw:
Seems to work with a windowHandle of `0` - The preview appears in the center of the screen though.

jeremypw:
`<Alt>space` does not work as an accelerator though.  Maybe it is already intercepted by the system? What alternatives do you suggest?

marbetschar:
@jeremypw best would be `space` because that's what macOS as well as GNOME uses. But I guess thats not possible, so maybe `<Alt>Enter`...?

Otherwise take whatever you like :)

jeremypw:
The first step would be to get the necessary changes to master approved.  The accelerator is set by the plugin.

jeremypw:
Looks like "SelectionEvent" is not being emitted by the previewer in response to arrow keys in the version present in Hera.  This means it is not possible to browse files just by pressing the navigation keys (the previewer grabs focus from Files).  This feature was added to a later version of gnome-sushi (which will not build on Hera).  It would only be possible to browse within files that are already selected anyway since Files plugins cannot access or control the View.  Anyway I'll push a branch soon for comment.

jeremypw:
In Files `space` causes the file(s) to open in the relevant application.  In my branch I have used `space` on the assumption that anyone installing this plugin wants the previewer _instead of_ the native behaviour.

marbetschar:
Maybe we get some better event handling in Sushi for Fossa - we‘ll see :)

OJFord:
> The problem with previewers is that we already do thumbnailing,

But it's a lot smaller; essentially just 'prettifying' rather than being utilisable for anything, unless I'm missing something/

> and opening Photos or Code or whatever the “real app” is almost always ends up being just as fast as or faster than opening a previewer app that has to be designed to preview any arbitrary file.

Assuming that was the correct file to open. On MacOS I find it a very handy way to quickly see a bit more about the file, next file, next file, no, ah yes this one, and _then_ open. Or not, maybe I didn't need to open it, just to check that it was what I thought it was.

> A previewer app feels like a weird and ultimately fruitless workaround for slow apps, and implementing it is destined to turn into some sort of “handle all the things” app instead of letting individual apps handle their own file formats.

Again, I don't think it's about avoiding a slow app, to me it's a different use case. I want to open the app to 'read'/'use' (as applicable) or edit the file. I want to preview it when I'm primarily thinking in terms of files, where they are in the filesystem; not their contents. I just might need a quick glimpse at the contents in order to do that file management, but I don't want to open & close the app (however fast it is to do so) so many times.

Real scenario: I downloaded my statements for the last couple of years from my electricity supplier, they were PDFs with supremely unhelpful names. I wanted to check the date at the top of the first and last ones, maybe one in the middle, (on MacOS I would actually probably have opened the first one and then just pressed 'down' repeatedly to flick through them all quickly) to verify that they did at least sort chronologically and what the start/end months were that I'd downloaded, so that I could then rename them more appropriately. I was thinking in terms of files and directories, not PDF contents, not reading, not editing them.

But in order to do that _file management_ in my _file manager_ (sorry to belabour that..) I did need to have a brief peek at a small part of the contents. `evince` starts up plenty fast, but still wasn't as handy as a floating pop-up preview window would've been.

cassidyjames:
As a note: if we go the route of https://github.com/elementary/files/pull/1036#issuecomment-665761699, then integrating a preview into the right sidebar might make more sense than a pop-up window; in that flow, the natural "end" state within the Files app is to show information about the files, reasonably including a higher-resolution preview. It also seems like it would be easier to ensure it behaves as expected, instead of throwing a dialog open over top of the Files window, obscuring your navigation (this is what I have always hated about the macOS preview flow).

jeremypw:
To embed a preview we would need a previewer widget instead of handing it off to an external service like Sushi, which results in a separate window.  It would be good if Sushi worked like a thumbnailer and provides a URL to the preview for the calling app to display.

marbetschar:
FWIW: there has been put some toughts into how a Gtk.Preview Widget may work here https://wiki.gnome.org/Projects/GTK/Preview

jeremypw:
Interesting - I wonder if anyone is working on it - that post is a couple of years old.

jeremypw:
That specification goes further than a "super-size" thumbnail in that e.g. pages of a document can be turned. It seems that once you start allowing actions on the file you may as well open an app.  What exactly do we want out of a preview?

jeremypw:
@marbetschar I do not think there is much change of a solution based on an external Sushi preview being acceptable - I've closed my related PRs.  Do you mind closing this one?

peteruithoven:
I just wanted to add one more argument on why a preview (sidebar or dialog) could be an improvement. Applications don't seem to know the order of the files in Files. This means that when you for example go to the next photo in Photos you often get a different photo than you'd expect. This would be solved with a more Files integrated preview.

There is the following issue to solve this for Photos specifically. 
https://github.com/elementary/photos/issues/14

follower:
Losing Mac OS's "Quick Look"/QuickLook functionality when moving to Elementary was definitely a downside so I'd like to address a couple of the aspects raised here:

#### What benefit does Quick Look-style preview provide over an application?

> @cassidyjames said:
>
>  [snip] opening Photos or Code or whatever the “real app” is almost always ends up being just as fast as or faster than opening a previewer app
>
> [snip] A previewer app feels like a weird and ultimately fruitless workaround for slow apps,

I concur with other comments about the _purpose_ of "Quick Look"/"Preview"/"gnome-sushi" style previews being *different* to the purpose of opening a file in an editing/creating/viewing app.

**There's a *reason* why I don't want to open the file in an application in this situation.** You've correctly identified that *one* of the possible reasons might be poor performance of app startup & in that situation I'd agree that solving the performance problem would be preferable.

However, there's at least two other scenarios unrelated to performance:

 * **There *isn't* a default app *or* there's more than one applicable app for the file** in question--for example, if I created a `.obj` or `.fbx` 3D model in Blender to later use in my game I'm making in Godot and maybe I'm wanting to email the file or something. And maybe there's a couple of variations of the model so I just want to check if it's variant A or B--that's where on a Mac I'd hit space bar and take a "quick look".

   **I don't _want_ to edit or import the file, so starting Blender or Godot in this scenario isn't useful**--even if the applications started immediately they're not designed as mere "viewers" so the workflow isn't conducive to my current task.

   Related is the case of something like a `.md` MarkDown file--if I'm using Quick Look it's because I want to look at the content--not edit it in the default editor which may or may not support e.g. formatted preview.

 * **Opening the file in the default application interferes with the current state of that application**.

   With Quick Look I hit space bar and *bam* in the center of the screen I can immediately see the content of the file I'm interested in.

   If the file opens in the default application does it load in this workspace or a different one or in a tab, oh, wait, the application opened the new file in a window/tab obscuring something, or is still busy compiling, or, or, or...

   And, sure, even if all the situations which are just "bad software" get fixed, there's always more situations which can't be controlled--maybe I'm running out of swap space or something.

   This also requires me to even *think* about what the currently configured default application even is--when that's not what I actually care about, I just want to take a quick look at the content during some other task I'm undertaking.

#### Does QL-style preview require a monolithic preview app to support it?

> [snip] that has to be designed to preview any arbitrary file.
> 
>  [snip] and implementing it is destined to turn into some sort of “handle all the things” app instead of letting individual apps handle their own file formats.

**It's not necessarily required that "Quick Look" style functionality is implemented in a single monolith of an application**, nor implemented outside of the native application itself.

On the Mac "Quick Look" functionality can be implemented in one of three ways:

 * **OS Built-in** -- for common formats used by multiple applications, e.g. plain text, multimedia etc.
 * **By native application** via QL plugin -- for application specific file formats the application can bundle a Quick Look plugin to render a preview.
 * **By 3rd party** QL plugin -- this may be for a format used by multiple applications or a "better" implementation of support that is application specific or for a format that doesn't have the application installed on this computer etc.

#### Wrapping up

Hopefully this is a useful perspective to help inform why Quick Look-style functionality is desirable and maybe shape some of the discussion about what form it could take. (Or at least it's skim-able if you're taking a...quick look. :) )

Thanks to all involved for the work in this PR & related code/issues to explore how best implement such functionality.

follower:
> @jeremypw 
>
> @marbetschar Do you know what to put in the `windowHandle` parameter of the `show_file` method?

FWIW the [gnome-sushi docs](https://gitlab.gnome.org/GNOME/sushi/-/blob/master/README#L16) mention:

> `ShowFile(s: FileUri, i: ParentXid, b: CloseIfShown)`
> [snip]
> If a XID [a.k.a. `ParentXid` a.k.a. `windowHandle`] is specified, the Sushi window is made transient to that window.

At a guess I'd suspect that may impact preview window position and/or focus aspects when the Files window XID is supplied as the value?

> @cassidyjames wrote:
>
> integrating a preview into the right sidebar might make more sense than a pop-up window;

While a right sidebar preview may be useful in and of itself, for my purposes it probably wouldn't replace a "Quick Look" style preview due to relative size.

Given how I tend to use Files (in a non full height/width window) the size of any preview in the side bar would be significantly smaller & thus less readable than the equivalent pop-up window could be (from full width or height up to full screen).

[Just noticed a response to my previous comment will respond to that separately.]

follower:
@janxkoci Thanks for your response & follow-up question.

> Would a proper thumbnailing be acceptable for you as a viable alternative to Preview?

I guess that would depend in part on what is considered a "thumbnail"--to my mind "thumbnail" implies that the resulting image is of such a size that, for example, any text would be too small to be readable (and/or intentionally rendered as ["greeked"](https://www.dictionary.com/browse/greeked-text)).

If the thumbnail was limited by the size of the right sidebar and the side bar was by default smaller than, say, half the width & 80% of the height of the screen then it wouldn't be large enough to be useful. (And my current Files window size wouldn't meet that size requirement.)

> The current direction is to provide high-res thumbnails in a sidebar 

I would say that for me (as someone primarily writing code & installing random software) most of my "Quick Look" use is viewing either raw plain text, rendered markup or PDF files. So, essentially, QL functions as a easily-access read-only viewer application/overlay. (I haven't used any of the later QL versions that added e.g. annotation functionality but some people seem to like that aspect.)

Also, for e.g. PDF files, QL renders multiple pages which you can scroll through. (It does something similar for text files but with an upper limit on file size, from memory.) 

Even for something like CSV files QL can render them in a table/grid which benefits from the larger display area a pop-up window would provide.

So, for my most common use cases my initial thinking is that a high-res thumbnail in a sidebar wouldn't be sufficient.

(One thought that just occurs to me now is that *maybe* a preview that was rendered within the confines of the Files window could be an option? e.g. maybe it slides in from the side so rather than a "side bar" it fills most of the window and you could click a "==>" arrow to close the preview/thumbnail. Or maybe the preview grows from the position of the file icon to fill (most of?) the Files window but not outside it--which might eliminate concerns related to the pop-up nature of the preview. But not sure about the UX implications of all that.)

marbetschar:
> At a guess I'd suspect that may impact preview window position and/or focus aspects when the Files window XID is supplied as the value?

Yes, that's exactly what I think this does. Unfortunately I found no API to retrieve the Files apps' XID back then - and even worse: If I recall correctly the XID is X11 specific. This means this will not work on Wayland anyways and inevitably lead to even more problems down the road.

Leaving phylosophical questions aside, there is serious work needed in libsushi upstream to get a feature like Quick Look merged into elementary eventually (e.g. lack of control of its behavior and look).

Feel free to take over this PR and digg into libsushi's code - but don't expect any quick fixes. This is quite a hard problem to solve.

follower:
#### XID Retrieval

> @marbetschar wrote:
>
> Unfortunately I found no API to retrieve the Files apps' XID back then 

During a brief investigation I found a reference to one approach [here](https://valadoc.org/gstreamer-pbutils-1.0/Gst.PbUtils.InstallPluginsContext.set_xid.html) but I haven't tried it:

```
##include <gtk/gtk.h>
##ifdef GDK_WINDOWING_X11
##include <gdk/gdkx.h>
##endif
...
##ifdef GDK_WINDOWING_X11
  xid = GDK_WINDOW_XWINDOW (GTK_WIDGET (application_window)->window);
##endif
```

Other possible leads:

 * <https://valadoc.org/gdk-x11-3.0/Gdk.X11.Window.get_xid.html>
 * <https://gitlab.gnome.org/GNOME/sushi/-/blob/5f2995abf97ca6d068046fedf598bb9a82dc88d3/src/libsushi/sushi-utils.c#L157-168>
 * <https://gitlab.gnome.org/GNOME/sushi/-/blob/30ec9b835cadfb6e95cc87fc376e65f5946d3b23/src/ui/application.js#L69>

#### Wayland support

> and even worse: If I recall correctly the XID is X11 specific. This means this will not work on Wayland anyways and inevitably lead to even more problems down the road.

Conincidentally, I noticed this `gnome-sushi` commit "[ Switch to using ExternalWindow class for external parenting ](https://gitlab.gnome.org/GNOME/sushi/-/commit/286a75274cd2d991f308af58365ade93d144f9d4)" from a couple of years ago which includes this commit message:

> The code is lifted from xdg-desktop-portal-gtk. This will allow
us to parent the window to an external Nautilus window even under
Wayland.

So this potential issue might already be resolved.

#### Arrow keys to change selected file

As a side note I also found these links related to using arrow keys to change the previewed file (when used with Nautilus, at least):

 * "Export a Visible property over D-Bus"
   * https://gitlab.gnome.org/GNOME/sushi/-/commit/30ec9b835cadfb6e95cc87fc376e65f5946d3b23
   * "This is needed to help Nautilus update the preview file when we're open and its selection change."
 * "Allow left/right arrow keys to change a file to preview"
   * https://gitlab.gnome.org/GNOME/sushi/-/issues/5

janxkoci:
> "Allow left/right arrow keys to change a file to preview"

So how does this work in List view or Column view? In those I would use Up/Down..

marbetschar:
A note on XID retrieval: Just came across [this snippet](https://github.com/elementary/calendar/blob/ed5ceffd928eb801e5a363a969e00158fef76663/src/MainWindow.vala#L190-L192) - so this seems to do the trick on X11. However, as noted earlier this wont work in Wayland.

jeremypw:
Closing as there is no immediate intention to implement a File Previewer.

PR #1291, opened 2020-04-21
Sushi experiment

jeremypw:
Experimental branch - not for merging


marbetschar:
That's pretty cool stuff!! The styling of Sushi just looks out of place and its window jumps around after starting - but otherwise this works suprisingly well :)

jeremypw:
Maybe if the correct window handle parameter is passed the jumping around would be fixed.  I believe that (on X) the `xid` of the FIles window is needed. Not sure what happens with Wayland though.

jeremypw:
Doesn't look like there is a way to get the xid of a Gdk.Window anymore.  Anyway Sushi needs to be made backend agnostic.

jeremypw:
If the UX team were to OK a preview function we would probably want it to be in a sidebar not provided by a third party app so I am closing this for now.

PR #1295, opened 2020-04-24
Prefer dark

danrabbit:


jeremypw:
There are (at least) a couple of issues with Files and dark theme:
* Outline of breadcrumbs not visible
* Outline of view switcher buttons not visible.

danrabbit:
There are still outstanding issues here with icons that I consider blockers. I don't think this is ready for review

cassidyjames:
@danrabbit we have a couple open on the icons repo, but those are outside the scope of Files imho. Is there anything in Files that's still blocking it?

danrabbit:
@cassidyjames Not that I'm aware of, but I still wouldn't consider merging this until those issues are fixed. I'm going to move this back to draft so it doesn't get accidentally merged pre-maturely

cassidyjames:
Blocked by https://github.com/elementary/icons/issues/853

danrabbit:
Unblocking per conversation in standup

cassidyjames (approved):

Issue #1307, opened 2020-04-30
Clone Repository with Ctrl + V (Git)


marbetschar:
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Feature

If I paste in Files (`Ctrl + V`) and there is a valid Git URL in the clipboard (an URL ending with `.git`, ), it would be awesome if files would clone the repository to the current folder.

## Use Case

During my work I use a few different IDE's and Editors. Because each of these work a bit differently to setup a new project, I often find myself firing up the console, navigate to the my Development directory and execute a simple `git clone URL` with an URL copied to clipboard before from GitHub/GitLab/etc.

As we have a Git integration in Files now, it would be awesome if we can simply clone a project by pasting it to a directory.

jeremypw:
This could be tricky at the moment as the git code is a plugin and plugins do not (presently) have a hook for pasted files.  May be able to add a context menu item though.

jeremypw:
If the shortcuts for plugins stuff I did for the Sushi plugin could get merged then it should be possible to assign a shortcut to cloning.

I agree about giving some indication of progress and errors.  There is also a question of whether we should support clone options and which (which would presumably require a dialog before the actual cloning).

marbetschar:
not sure about cloning options, I don't think thats necessary. In fact, I wasn't even aware there are any options up to this point to be honest 😅️

jeremypw:
Showing a progress bar is problematic as I do not think the relevant information and signals are available for Ggit library.  May be able to provide a spinner though.

jeremypw:
Even a spinner and cancel button do not work easily - the clone_repository function is not asynchronous and cannot be interrupted.  So for the moment I am just showing an InfoBar.

jeremypw:
The linked PR is closed as it seems better to implement this in Code if possible.  I'll keep the branch around in case a specific use case for Files is made to the project leaders.

jeremypw:
Closing as cloning repositories is now implemented in Code.

PR #1310, opened 2020-04-30
Add Git Clone menu option

jeremypw:
Partly addresses #1307 except it provides a context menu option not a keyboard shortcut.


plugins/git/plugin.vala
```diff
@@ -167,6 +167,65 @@ public class Marlin.Plugins.Git : Marlin.Plugins.Base {
             }
         }
     }
+
+    public override void context_menu (Gtk.Widget? widget, List<GOF.File> files) {
+        var menu = (Gtk.Menu)widget;
```
tintou, line 172:
```suggestion
        unowned Gtk.Menu? menu = (Gtk.Menu)widget;
```

jeremypw:
@tintou Do you know how to have a spinner and cancel button work when Ggit.Repository.clone () is synchronous and does not take a cancellable as a parameter?  Is it even possible?

marbetschar:
Any chance to get this merged? I'd LOVE to see it in Files 😍️

jeremypw:
Looking into using GLib.Task to make the cloning cancellable.

jeremypw:
Looks like you cannot actually cancel the cloning operation using GTask, as far as I can see.  You can "cancel" the task but the cloning continues to run in its own thread so the files are still downloaded.  But at least the Files UI is not blocked.

jeremypw:
I think there is a way to effectively cancel the cloning but it is a bit of a hack.  If you trash the ".git" folder then cloning ends with an error (!).

jeremypw:
One question to be decided is whether, when cloning into an empty folder from the background menu, a new root folder should be created with named from the git repo.  At the moment cloning into a location creates the .git in that location, i.e. that folder becomes the root folder.  It is therefore left to the user to create a suitable empty root folder first.  It might be convenient for Files to do that automatically in this circumstance (and not when cloning from the selection context menu).

Probably should consider having a similar feature in Code.

marbetschar:
I guess it would be best to do the same as git clone does when executed as terminal command: create an empty subfolder using the xyz filename from the url: GitHub.com/group/xzy.git

jeremypw:
> I guess it would be best to do the same as git clone does when executed as terminal command: create an empty subfolder using the xyz filename from the url: GitHub.com/group/xzy.git

Yes, it is a bit annoying that `GGit.Repository.clone (origin, target, options)` does not already do this.

jeremypw:
I don't think it is possible to implement git cloning using `<Ctrl>V` of a git url without moving a lot of git stuff out of the plugin into libcore.  Do we really want to add a dependence on libgit2-glib to the core?

marbetschar:
If the context menu action does the job but without keyboard shortcut, I think we are already 90% there - and the shortcut isn’t too important at all in my opinion.

So if it works as context action and without forcing further dependencies to core it seems good enough to me.

What do you think?

jeremypw:
I agree.  I suspect that it would be much more difficult to get it merged if it involved substantial changes to the core.

marbetschar:
@jeremypw any chance you have some time to resolve the conflicts, so this PR can be re-reviewed? 🤗️

jeremypw:
@marbetschar I have resolved the conflicts and also allowed the plugin to clone local repositories as well.

marbetschar:
@jeremypw just installed this on Odin and it seems to have some problems with detecting Git repositories for SSH urls (just put the URLs into clipboard using the copy button on GitHub while in Firefox):

```
** (io.elementary.files:35374): WARNING **: 14:08:31.433: plugin.vala:291: Not a git repository - git@github.com:marbetschar/time-limit.git

** (io.elementary.files:35374): WARNING **: 14:08:44.063: plugin.vala:291: Not a git repository - git@github.com:elementary/files.git
```

FWIW: The HTTP URLs do work without issues. Although the text (or icon) alignment for the progress indication seems to be off vertically:

Screenshot from 2021-02-08 14-12-00

jeremypw:
@marbetschar Thanks for the further testing - I must admit I had not considered SSH urls till now.  I'll see if I can implement them.

jeremypw:
Looks like GitHub omits the `ssh://` protocol - which explains why it does not work.  I guess I could have the cloner assume the protocol is SSH if the url begins in the form `[user@]host.xz[:port]` (?)

jeremypw:
Another thing that this PR does not address if where the `git clone` command requires user feedback.

jeremypw:
@marbetschar I have now set up so that I can clone GitHub repos using SSH in the terminal successfully.  However, this PR, using the libgit2 package,  gives an error "Authentication required but no callback provided".  There does not seem to be a simple solution to this (may be a bug in the package) as there is no parameter for a callback in the `Ggit.Repository.clone ()` command.

Would you consider it worthwhile merging a functionality limited cloning local and https urls for now and addressing ssh later?

jeremypw:
I have removed the code trying to vaildate the URL.  It is now just passed to the `Ggit.Repository.clone ()` and an error shown if anything fails.

jeremypw:
The alignment problem seems to be a style sheet issue - I cannot find anything in the code that would be causing this.

marbetschar:
@jeremypw

> Looks like GitHub omits the ssh:// protocol - which explains why it does not work. I guess I could have the cloner assume the protocol is SSH if the url begins in the form [user@]host.xz[:port] (?)

Yes, from what I can tell this should be Ok. Maybe its fine to just check, if the URL ends with `.git` - and then, if no specific protocol is given, use SSH. Protocol wise, thats how I'd expect `git clone` works internally.

> Another thing that this PR does not address if where the git clone command requires user feedback.

I understand. Usually that's just confirming the servers' fingerprint on initial connection. Of course this would be nice to do so in the GUI with a yes/no dialog - but throwing an appropriate message is probably fine for now.

> However, this PR, using the libgit2 package, gives an error "Authentication required but no callback provided". There does not seem to be a simple solution to this (may be a bug in the package) as there is no parameter for a callback in the `Ggit.Repository.clone ()` command. Would you consider it worthwhile merging a functionality limited cloning local and https urls for now and addressing ssh later?

Personally, for 9 out of 10 repositories I use the SSH protocol to clone. Otherwise I'm not able to make any modifications, because HTTPS is anonymous therefore I'm not authenticated - which implies I'm not authorized to make any changes.

Just a wild guess: Is it possible to pass the path to the default public key `~/.ssh/id_*.pub` as option? Maybe thats the missing piece? (usually `id_rsa.pub`, but [GitHub changed this to `id_ed25519.pub`](https://docs.github.com/en/github/authenticating-to-github/adding-a-new-ssh-key-to-your-github-account))

If this does not work, my next best guess is the VAPI might be missing something. Unfortunately valadoc.org seems to be down atm, but maybe @tintou can help out regarding the missing callback argument in `Ggit.Repository.clone ()` of `libgit2`?

jeremypw:
Looks like the the vapi does have ssh credential related stuff - just need to work out how to use it ...

jeremypw:
Looks like the required callback is passed in the FetchOptions of the CloneOptions parameter of the clone functions.  There are some clues here https://gist.github.com/zchee/444c8c20aa7756468d8e.

marbetschar:
@jeremypw thanks for the hint! Just made some small adjustments - now cloning using SSH works as well as cloning with HTTP.

However, the context menu option now seems always to be shown regardless of whats exactly is in the clipboard. No clue why ;)

jeremypw:
@marbetschar  Thanks!  I tried much the same thing but for some reason it did not work for me.  I'll try your version.

jeremypw:
I can confirm that I can now clone repositories from GtiHub with Files using SSH :tada:  

I'll look at the context menu issue now.

jeremypw:
Very basic validation added for clipboard contents.  Providing a comprehensive Regex for valid urls of git repositories is non-trivial. Leave for another PR?

marbetschar:
@jeremypw works like a charm! IMHO your git repository url verification is perfectly fine (at least thats exactly how I would do it too).

IMHO this RP is ready to be merged. Maybe someone from @elementary/ux can do a review?

danrabbit:
I'm kind of hesitant about this because I feel like we might be getting into kind of a scope creep/feature overlap situation with Code. When we're working with git repositories, it seems to me like the expected workflow would be to open Code to create a new project or clone an existing project

jeremypw:
Yes, I think you are right that this should be implemented in Code first (although doing it with Files makes it easy to specify where to create the clone.  I'll close this for now and look at doing it in Code.

marbetschar:
@danrabbit / @cassidyjames I'm aware of the feature overlap situation with Code and I do understand the concerns of feature boilerplate creeping into Files: Basically everything we do on a Computer is somehow Files related - so a line needs to be drawn somewhere. Just wanted to check back if the decision to explicitly _not_ include Git repo cloning is final?

My personal use case: I don't use Code for all projects. Depending on the project there are better suited editors/IDE's. But all of these projects are stored in a Git repo somewhere - and all of them need to be cloned somewhen as first step. So that's the least common divisor. From there on, its very project specific what comes next.

I can use the Terminal for all of these stuff for sure - so its not a deal breaker. It would just have been super convenient to quickly clone from Files.

PR #1314, opened 2020-05-02
Add message_plugin hook and associated action in app

jeremypw:
Allow a plugin to set a target for a new "app.action-message-plugin" action together with, if desired, an accelerator.  The action triggers a plugin hook that passes the target string and the currently selected files to the plugins.  An example of the use of this can be found at #1291.

Shortcuts could also be set for core contracts for example.


jeremypw:
There is no immediate need for this so I will close for now to reduce noise.

PR #1315, opened 2020-05-03
Allow remembering of password for afp:// shares

### LINKED ISSUE #1244: Remember Password forever does not work on remote volumes
## Prerequisites
- [X] I have searched open and closed issues for duplicates.
This bug is similar to #934 but neither is the scenario the same nor is the solution acceptable.

## Describe the bug
When connecting to a auto discovered network server over AFP, files will ask me for a password when connecting to the server. Then I am asked a second time when connecting to a share (which is annoying, but as said in #934 probably is due to the underlying system requirements - it would be fantastic although if this could be fixed). Both times I always choose "Remember forever". But the next time I connect to the server I am asked for my credentials again.

### Expected behavior
I expect the server I specified the password for and marked "remember forever" tu do just that... remember it forever ;)
It is not acceptable in my scenario to bookmark the server as I have various dozens of servers and shares to connect to and it would be a complete mess to bookmark all of them.

## Platform and additional Information
elementary OS 5.1 Hera updated to latest in a mixed office sharing network with mostly Macs and Synology Diskstations.
- [X] I'm using the latest released stable version of files

Best regards
Merlin

jeremypw:
Fix #1244 

This PR works around a bug (or feature) of the STANDARD_TARGET_URI attribute of folders on an AFP server.  It seems that this attribute loses the user name from the target uri and this results in the system dialog asking for a password always appearing even if the "Remember for ever" option is ticked.  This PR causes the username to replaced in the target uri and so allows the remembered password to be retrieved.


dsancheznet:
After checking this branch for several days, I have found the patch to work as expected with no undesired behaviour.

libcore/FileUtils.vala
```diff
@@ -881,6 +881,52 @@ namespace PF.FileUtils {
...
+        /* For afp:// addresses the standard target uri has the user name stripped - we need to replace it */
+        if (Uri.parse_scheme (uri) == "afp") {
+            string origin_host, origin_filename;
+            if (get_afp_user_server_and_filename (uri, out origin_filename, out origin_host)) {
+                string target_host, target_filename;
+                if (get_afp_user_server_and_filename (standard_target_uri, out target_filename, out target_host)) {
+                    return "afp://" + Path.build_path (Path.DIR_SEPARATOR_S, origin_host, target_filename);
+                }
+            }
+        } else if (Uri.parse_scheme (standard_target_uri) == "afp") {
```
danrabbit, line 894: Since we already check if the target URI scheme is `afp` before calling this function, we probably don't need to check again right? It's probably reasonable to assume that we would always check that target URI is `afp` in order to call this function right?

libcore/FileUtils.vala
```diff
@@ -881,6 +881,52 @@ namespace PF.FileUtils {
...
+            /* Probably clicked on an AFP server in network:// view which gives weird gvfs uri like:
+             * network:///dnssd-domain-<host>._afpovertcp._tcp.  We have to get any missing username from
+             * the Environment */
+            string target_host, target_filename;
+            if (get_afp_user_server_and_filename (standard_target_uri, out target_filename, out target_host)) {
+                if (!target_host.contains ("@")) {
+                    /* Cannot get username from origin uri so just use Environment user name */
+                    target_host = GLib.Environment.get_user_name () + "@" + target_host.dup ();
```
danrabbit, line 902: Hrm is that reasonable to assume the computer user name is the same as the afp user name? Maybe in this case where we can't get the username we should be prompting. Or will this fail to prompting for the user name if the uri turns out to be invalid?

libcore/FileUtils.vala
```diff
@@ -881,6 +881,52 @@ namespace PF.FileUtils {
...
+                    return "afp://" + Path.build_path (Path.DIR_SEPARATOR_S, target_host, target_filename);
+                }
+            }
+        }
+
+        return standard_target_uri;
+    }
+
+    private bool get_afp_user_server_and_filename (string uri, out string filename, out string user_server) {
```
danrabbit, line 911: Should this throw a GLib.Error instead of returning true and false? Right now it kind of looks like we just fail silently if this doesn't work out

danrabbit (commented):
Left a few questions!

libcore/FileUtils.vala
jeremypw: Yes, I guess that is safe enough.  Replacing check with a comment.

libcore/FileUtils.vala
jeremypw: I've just added some terminal output - is that enough?  This gets called every time the file updates and we do not want to repeatedly throw error dialogs.

libcore/FileUtils.vala
jeremypw: This gets called every time the file updates and we do not want to repeatedly throw dialogs.  I could do some more testing of error situations - I haven't worked on this since reinstalling so I need to set things up again.

jeremypw: On further thoughts, if this work around fails it will just cause the normal system prompt to show instead of automatically connecting (as now) so I do not think there is any point in throwing our own dialog.

jeremypw:
There is a limitation when navigating to a server without using the Connect Server dialog (by clicking on an icon in the `Entire Network` view) in that there is no option to enter a different user name - Files assumes that the name to use is the current local user name.  However, this is not caused by this PR - it is pre-existing, so it should not block this.

libcore/FileUtils.vala
jeremypw: On even further thought, I think it best to omit the username altogether in this circumstance - this triggers a system dialog requesting the name as well as the password.

leonardo-lemos (approved):
Works as expected. Approved.

PR #1318, opened 2020-05-06
Better small icon view

### LINKED ISSUE #1219: Option to make the list view more compact
## Prerequisites

## Feature
**Problem**
When working with directories containing huge number of files (like a source check out from git etc) the huge space between items is a distraction and makes poor use of the screen real estate.

One could argue that for such use cases, a terminal would be more appropriate; but what about those addicted to nice GUIs from elementary (great work guys)!

**Solution**
Provide an option or an alternate view where the display is more compact.

* Can have tiny icons just enough to identify the file-type
* No thumbnails
* Option to have monospace font would also help (for huge directories with fixed-format filenames)

**Existing work**
Gmail offers a similar approach where it is possible to have the spaced out view (default config), but for those who prefer a compact view, there is an option for that too.

KDE's dolphin can be configured for a compact view.

image

**Describe alternatives you've considered**
* Switch back KDE, but not as nice over all.
* Use dolphin as a flatpak; dark theme doesn't seem to work.

**Additional context**
N/A

jeremypw:
Fixes #1219 

More fine-grained control of the layout of IconView at different icon sizes if implemented and used to increase the compactness of the view at icon sizes below Marlin.IconSize.NORMAL.

* Reduce spacing and padding
* Tighten the focus rectangle
* Reduce the maximum number of lines of the name but maintain a larger minimum width to avoid tall narrow text cell.

The IconView no longer observes the "minimum zoom level" setting, which of little benefit and will be removed.

Some improvements in the layout of multiple emblems with small icons were made.

Little change should be noticeable at normal icon size and larger.


jeremypw:
@danrabbit Is it worth pursuing this?  A number of conflicts have now arisen.

jeremypw:
Closing as multiple conflicts have arisen and no recent activity.

PR #1322, opened 2020-05-07
Appmenu for click mode setting

### LINKED ISSUE #1319: Add a toggle for double/single click in settings
A toggle in settings to enable/disable clicking/single click on files would be a great option that would work for a ton of people. I personally love the single click style, but I am probably one in a million, but it would definitely end a never ending discussion on the subreddit
Thank you

jeremypw:
Fixes #1319 

Minimal implementation.


jeremypw:
I think its best to withdraw this for now to concentrate on the mixed mode only PR.  If that is merged then this becomes redundant.

Issue #1341, opened 2020-06-02
4.4.3 Build fails with both GCC 10 and Clang 10


alucryd:
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Describe the bug
Building 4.4.3 fails for different reasons with GCC 10 and Clang 10. FYI I can "fix" the clang build by temporarily passing `-Wno-return-type`.

### Logs
GCC: https://buildbot.alucryd.xyz/#/builders/213/builds/10/steps/7/logs/stdio
Clang: https://buildbot.alucryd.xyz/#/builders/213/builds/11/steps/7/logs/stdio

## Platform Information
- [ ] I'm using the latest version from git that I've manually compiled
- [x] I'm using the latest released stable version

Arch Linux, GCC 10, Clang 10.

jeremypw:
Did the previous release build OK under the same conditions?

jeremypw:
The Clang problem seems to be due to a quirk in the way the Vala compiler handles `return_if_fail` in a void async function.  If this is replaced by a simple `if` clause then the problem goes away.

jeremypw:
@alucryd I think I fixed the CLang10 issue with the linked #1342 but cannot test it.  Are you able to confirm whether that PR fixes that issue.  

The issue with GCC10 has a different cause and should probably have been raised separately so that it can fixed in a separate PR.

alucryd:
@jeremypw This indeed fixes the clang building, thanks!

jeremypw:
Thanks - I'll look into the GCC10 issue.

jeremypw:
@alucryd I have pushed a draft PR (linked) that may fix the GCC 10 build problem - are you able to test it please?

alucryd:
Sure, it still fails to build, here is the error log: https://paste.xinu.at/vXQaHWjlh8gFw8Rc/

jeremypw:
@alucryd  Thanks for trying!.  Seems I have got rid of some errors so I am on the right track.  Hopefully next commits will compete the fix.

alucryd:
@jeremypw did you get around completing your GCC fix? 4.4.4 still fails to build here.

jeremypw:
@alucryd  Sorry - rather forgot about that one.  I'll have another look at it.

jeremypw:
@alucryd  Is 84ee73f sufficient to fix the GCC problem?  If so I'll revert the PR to that point.  It started failing CI after that and I am having trouble identifying and fixing the problem.

[...]

PR #1342, opened 2020-06-02
Fix Clang10 build

jeremypw:
See #1341 

The Vala compiler does not correctly handle the `return_if_fail ()` in async functions as it gets expanded within a boolean function.  With Clang10 this causes the build to fail.  Not sure what the effect would be if this were actually triggered at run time.   

Only the ctags plugin uses this macro within an async function so it is just replaced with a simple `if` clause (which compiles to correct C code).

### DIFF
--- a/plugins/pantheon-files-ctags/plugin.vala
+++ b/plugins/pantheon-files-ctags/plugin.vala
@@ -183,7 +183,9 @@ public class Marlin.Plugins.CTags : Marlin.Plugins.Base {
     }
 
     private async void rreal_update_file_info (GOF.File file) {
-        return_if_fail (file != null);
+        if (file == null) {
+            return;
+        }
 
         try {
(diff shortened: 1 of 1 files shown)


tintou:
I think that technically, if you put `GOF.File file` as an argument, Vala should throw a warning if file is NULL though

jeremypw:
@tintou My understanding is that Vala can't tell at build time whether the file is null (possibly due to a previous program error) and wont stop the file being referenced (causing a seg fault) at run time.

I think these guards were put in as a quick fix to prevent old undiagnosed bugs causing the color tagger to crash.  They may no longer be required.

jeremypw:
I'll ask the reporter to confirm it fixes the CLang10 issue before merging.

jeremypw:
Reporter confirms this fixes the CLang10 issue.

PR #1343, opened 2020-06-02
ProgressInfoWidget: rewrite

danrabbit:
Screenshot from 2020-06-02 16 26 00@2x

* Solves a ton of compiler warnings about unsafe casts
* Use grid instead of nesting boxes
* Use `width_chars` instead of hardcoding pixel width
* GObject style
* Gtk style class constant


jeremypw:
Files compile without warnings (apart from re zeitgeist vapi) on Hera so I guess you are building on Odin with later versions of the compilers. I havent got an Odin install yet as I havent been able to access the iso so I can only review on Hera.

PR #1360, opened 2020-06-08
Fix building with GCC 10

### LINKED ISSUE #1341: 4.4.3 Build fails with both GCC 10 and Clang 10

jeremypw:
Attempt to fix #1341 (GCC part)

The symbol name "settings" is not used and is replaced by names specific to where it was used in order to try and avoid errors such as 
```
(.text+0x0): multiple definition of `settings'; src/25a6634@@io.elementary.files@exe/meson-generated_Application.c.o (symbol from plugin):(.text+0x0): first defined here
/usr/bin/ld: src/25a6634@@io.elementary.files@exe/meson-generated_View_Window.c.o (symbol from plugin): in function `settings':
```

when compiling with GCC 10.

"settings" is also removed from the Preferences namespace in `marlin.vapi` (This vapi may no longer be needed anyway as certain settings are exposed to the library through the GOF.Preferences object).


jeremypw:
MarlinIconInfoTests failing for no apparent reason :-(

jeremypw:
I can replicate the CI failure locally by uninstalling Files (with its schema) and then trying `ninja test`.  What is puzzling is that the tests have not changed and previous passed (and pass in master).

jeremypw:
Seems the current commit is not sufficient to fix GCC10 compilation (compiles with GCC9.3) so need to make more changes without breaking CI.

jeremypw:
With the latest commit, the PR both builds with GCC10 and passes CI.

src/Application.vala
```diff
@@ -31,10 +31,25 @@ public class Marlin.Application : Gtk.Application {
     private const uint MAX_WINDOWS = 25;
 
     public int window_count { get; private set; }
+    public Settings marlin_app_settings { get; construct; }
+    public Settings marlin_icon_view_settings { get; construct; }
+    public Settings marlin_list_view_settings { get; construct; }
+    public Settings marlin_column_view_settings { get; construct; }
+    public Settings gnome_interface_settings { get; construct; }
+    public Settings gnome_privacy_settings { get; construct; }
+    public Settings gtk_file_chooser_settings { get; construct; }
```
danrabbit, line 40: Should these be constructed statically instead? Then you wouldn't have to reference the instance as in `slot.window.marlin_app.marlin_app_settings` which is kind of verbose

jeremypw: Done.

PR #1369, opened 2020-06-14
core: Port some function to Vala

tintou:
Port some utilitary function from marlin-file-operations.c to FileUtils.vala and leave them as internal.


libcore/FileUtils.vala
```diff
@@ -804,6 +804,178 @@ namespace PF.FileUtils {
...
+    }
+
+    internal int get_max_name_length (GLib.File file_dir) {
+        if (!file_dir.has_uri_scheme ("file")) {
+            return -1;
+        }
+
+        var dir = file_dir.get_path ();
+        var max_path = Posix.pathconf (dir, Posix.PathConfName.PATH_MAX);
+        var max_name = Posix.pathconf (dir, Posix.PathConfName.NAME_MAX);
```
jeremypw, line 849: Need to bump minimum version of `valac` in order to support `Posix.pathconf ()`.  On Hera, fails to compile with valac version 0.40.  These lines compile with valac version 0.48 but build fails due to faults in the zeitgeist.vapi.  Unless this can be fixed we will have to hold off merging this until Odin is released.

tintou: done, Odin is now the target branch anyway :slightly_smiling_face:

libcore/FileUtils.vala
```diff
@@ -804,6 +804,178 @@ namespace PF.FileUtils {
...
+        minutes = seconds / 60;
+        if (minutes < 60) {
+            time_unit = minutes;
+            return ngettext ("%'d minute", "%'d minutes", minutes).printf (minutes);
+        }
+
+        hours = minutes / 60;
+        if (hours < 4) {
+            time_unit = minutes + hours;
+            return _("%s, %s").printf (
```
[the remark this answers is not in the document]
tintou: added

PR #1391, opened 2020-07-13
Stop FileChooser crashing when used as FileChooserNative (e.g. with flatpak)

### LINKED ISSUE #1390: Save file window vanishes when mouse cursor move to the top
When I want to save a file in Firefox (flatpak) the save file window disappears when I move the mouse cursor to the top of the save file window.

I have tried to replicate it in Linux Mint but was unable to do so. I'm not sure if this is a flatpak problem, a problem with elementary, or a flatpak problem with elementary.

I have attached a screen record to show you what I mean.

Thanks

App used: Firefox 78.0.3

[Screen record from 2020-07-11 01.37.32.mp4.zip](https://github.com/elementary/gala/files/4907676/Screen.record.from.2020-07-11.01.37.32.mp4.zip)

jeremypw:
Fixes #1390 

This now fixes the issue "properly".  It turns out it is not valid to assume that the display associated with the entry window is the same as the default display when running from a flatpak.  Using the display associated with the entry_window to make the cursor fixes the issue (at least on X).  I suspect the original code would only work on X anyway and will probably have to be revisited for Wayland.


jeremypw:
To test this PR, 

* Install the FireFox flatpak
* Install this PR
* Log out and back in (or otherwise reload the modules)
* Run FireFox from the flatpak
* Request Save File from FireFox e.g. by right-clicking on an image.
* Move the pointer over the path bar.  The cursor should change appropriately depending on whether the pointer is over a breadcrumb or the text entry region.
* Using the master version of Files, the filechooser crashes immediately the pointer enters the pathbar.

d1moncom:
I have crashing in Firefox and Chromium, when when use FileChooser. Problem is only sometimes. I use Firefox and Chromium without Flatpak!! Web browsers are installed via apt install.

jeremypw:
@d1moncom If you are not using Flatpak then its probably not related to the issue this PR fixes.  Please raise a new issue, identifying the conditions under which the crashes occur. Thanks.

PR #1392, opened 2020-07-14
Reduce minimum icon size to "smallest" in List View

### LINKED ISSUE #1164: Too large icons in List-view
Icons are to too big by default. As such I feel that the 'list-view' becomes less effective.  Using dconf-editor I changed all zoom-levels to 'smallest'. Much better!

jeremypw:
Fixes #1164
There does not seem to be a good reason to maintain this restriction on the user.

In order to gain the benefits in compactness possible with a smaller icon size, the line spacing has been reduced.  At all but the smallest icon size the icon height dominates in anycase. 

In addition, the layout of the icon, helper and emblems has been tweaked so that less of the icon is obscured and enabling showing of the emblems at smallest icon size.


danrabbit:
It looks like at the smallest zoom in list view we lose icon badges such as the symlink or read-only emblems.

It's also weird that this is a dconf setting, so changing this won't actually fix the issue for users with existing installs since we don't overwrite user preferences on update

jeremypw:
How about removing the setting altogether for the List and Column Views?  There is a different PR outstanding that tries to address a similar issue with Icon View.

jeremypw:
Regarding the emblems, this was done to avoid completely obscuring the icon with the overlapped emblems.  In the linked issue I suggested maybe the emblems should be moved to the side (non-overlapping) at the smallest icon size instead of hiding but I did not want to complicate this PR.

danrabbit:
Yeah I agree we should remove the setting. It doesn't really seem like there's a good reason for this to be configurable. We should support the sizes we support :)

My feeling is that we should probably address issues with the smallest icon size before making it generally available for people to use

jeremypw:
I am not sure the hiding is really a problem (or not a new one) as this has happened in Column View for a long time with no issues raised.

jeremypw:
I have now removed the "minimum-zoom" setting altogether for the two tree views.  I have left for IconView as this does not cope well with the smallest icon sizes.  There is another PR outstanding that attempts to produce a usable IconView at small icon sizes - see #1318

jeremypw:
AFTER 
Screenshot from 2020-07-18 19 07 50

jeremypw:
BEFORE
Screenshot from 2020-07-18 19 13 47

jeremypw:
@danrabbit I have attempted to improve the situation with emblems and small icons, as you suggested.  I have increased the xpadding and used the extra space to reduce the amount that the emblems overlap the icon at small icon sizes.

src/View/AbstractTreeView.vala
```diff
@@ -349,6 +352,49 @@ namespace FM {
...
+                        break;
+                }
+
+                try {
+                    view_css.load_from_data ("treeview { -GtkTreeView-vertical-separator: %i px; }".printf (vsep));
+                    get_style_context ().add_provider (view_css, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+                } catch (Error e) {
+                    warning ("Could not style FM.TreeView");
+                }
+            }
```
danrabbit, line 385: I would really prefer not to do this. If the padding is too large for tree view rows, we should address that in the stylesheet and not special case Files styling

src/View/ColumnView.vala
```diff
@@ -56,21 +56,20 @@ namespace FM {
             return false;
         }
 
-        protected override Marlin.ZoomLevel get_set_up_zoom_level () {
-            var zoom = Marlin.column_view_settings.get_enum ("zoom-level");
-            Marlin.column_view_settings.bind ("zoom-level", this, "zoom-level", GLib.SettingsBindFlags.SET);
+        protected override void set_up_zoom_level () {
+            Marlin.column_view_settings.bind ("zoom-level",
+                                                           this, "zoom-level",
+                                                           GLib.SettingsBindFlags.DEFAULT);
```
danrabbit, line 62: I don't really think this is easier to read than having it inline, but if it must be shortened then code style is for it to be done like:
```suggestion
            Marlin.column_view_settings.bind (
                "zoom-level",
                this, "zoom-level",
                GLib.SettingsBindFlags.DEFAULT
            );
```

src/View/IconView.vala
```diff
@@ -86,9 +87,10 @@ namespace FM {
             return tree as Gtk.Widget;
         }
 
-        protected override Marlin.ZoomLevel get_set_up_zoom_level () {
-            var zoom = Marlin.icon_view_settings.get_enum ("zoom-level");
-            Marlin.icon_view_settings.bind ("zoom-level", this, "zoom-level", GLib.SettingsBindFlags.SET);
+        protected override void set_up_zoom_level () {
+            Marlin.icon_view_settings.bind ("zoom-level",
```
danrabbit, line 91: same as above for code style

src/View/ListView.vala
```diff
@@ -198,22 +198,20 @@ namespace FM {
             return tree as Gtk.Widget;
         }
 
-        protected override Marlin.ZoomLevel get_set_up_zoom_level () {
-            var zoom = Marlin.list_view_settings.get_enum ("zoom-level");
-            Marlin.list_view_settings.bind ("zoom-level", this, "zoom-level", GLib.SettingsBindFlags.SET);
+        protected override void set_up_zoom_level () {
+            Marlin.list_view_settings.bind ("zoom-level",
```
danrabbit, line 202: same code style comment as above

danrabbit (dismissed):
Left a few inline comments

I'm not a fan of injecting custom CSS in this way. If the padding for treeview rows is too big, we should address that at the stylesheet level and not make Files inconsistent with other apps

src/View/AbstractTreeView.vala
jeremypw: I agree in principle - this is more by way of a demonstration.  The padding only changes for small icons - can you accomplish that in the stylesheet? Would you add two new classes for the treeview with different padding?

src/View/IconView.vala
jeremypw: Theres a lot of old code where this style needs updating!

src/View/AbstractTreeView.vala
danrabbit: I'm not sure we really need to have two different paddings. Maybe we can open an issue in the stylesheet to reconsider the default and see where we wind up?

jeremypw: The row padding is reduced at small icon size to get the maximum benefit in terms of compactness a well as avoiding an unbalanced appearance (space between rows too large compared with the row content).  There are in fact three different paddings used in this PR (0, 3 and 5 px).  But I agree it is ultimately up to the UX team to arrive at the appropriate compromise. I'll open an issue in stylesheet.  Is this PR mergeable if I remove the CSS or shall I put it on the backburner?

jeremypw:
Smallest icon size without CSS to reduce row padding:
Screenshot from 2020-08-20 10-23-40

jeremypw:
@danrabbit - Is it worth pursuing this?

danrabbit:
Hey sorry I haven't followed up here in a while. I thought I just had was maybe in the list and column views we should treat emblems as their own columns instead of badging the icon:

badges

jeremypw:
A problem with putting the emblems in their own column is that the emblems would be shown  vertically aligned outside the right edge of the filename column in the view therefore potentially far from a short name.   It would also result in up to 4 column headers for usually empty columns (the current implementation caters for up to 4 emblems).

It may be possible to change the TextRenderer class to render the emblems just after the filename, or have the IconRenderer show them differently in the icon column, but I think that would be out of scope for this PR and should be the subject of a new issue.

jeremypw:
@danrabbit So, how do you want to proceed with this PR?

danrabbit (approved):
Does what it intends to do and ditches some unecessary gsettings stuff. We can file other issues and follow up on making list views better in other branches I think :)

PR #1394, opened 2020-07-15
Use proper icon size for back and forward

lainsce:
This fixes the buttons' icon size so it's a button, no icon size hacks :stuck_out_tongue:
this was made apparent after the `go-*` icon PRs I made:
https://github.com/elementary/icons/pull/919 and https://github.com/elementary/icons/pull/920

| Old | New |
|:---:|:---:|
|Screenshot from 2020-07-15 19 39 58|Screenshot from 2020-07-15 19 40 31|


danrabbit:
This is not correct. These are intentionally 24px icons, not 16px

PR #1398, opened 2020-07-23
Make Properties header-title selectable when not writable

### LINKED ISSUE #1397: Can't copy file/folder name if my user don't have write permission
There is no way to copy a file/folder name if my user don`t have write permission on the item. 

If I want to copy an item name on files I usually use the rename action and copy that name, but when I don`t have write permission on the item I cant rename and copy the name. 

File/Folder `properties` don`t have the file name, this would solve the problem.

## Prerequisites

## Feature

- copy a file/folder name when user don`t have ownership 

**Describe alternatives you've considered**

Add the file name on item properties (info)

image

jeremypw:
Fixes #1397 

This provides a simple though not entirely discoverable way of copying the filename when renaming is prohibited.

It is questionable whether this is an important enough function to warrant a special menu entry or shortcut. One possibility would be for the "Rename" menu option to change to "Copy name" when the file is not accessible.

If required that can be added in another PR.


src/Dialogs/PropertiesWindow.vala
```diff
@@ -198,11 +198,16 @@ public class PropertiesWindow : AbstractPropertiesDialog {
...
-        if (!only_one || (only_one && !goffile.is_writable ())) {
+        if (!only_one ) {
             var label = new Gtk.Label (get_selected_label (selected_folders, selected_files));
             label.halign = Gtk.Align.START;
             header_title = label;
-        } else if (only_one && goffile.is_writable ()) {
+        } else if (!goffile.is_writable ()) {
+            var label = new Gtk.Label (goffile.info.get_name ());
+            label.halign = Gtk.Align.START;
+            label.selectable = true;
```
danrabbit, line 208: Latest code style :)

```suggestion
            var label = new Gtk.Label (goffile.info.get_name ()) {
                halign = Gtk.Align.START,
                selectable = true
            };
```

jeremypw: Done.

PR #1403, opened 2020-07-26
Sidebar: Show basename for bind mounts as well as mount name to avoid duplicate device entries

### LINKED ISSUE #1401: Duplicate entries under Devices when there is a bind mount to a device subdirectory

## Describe the bug
If a device is mounted as usual in fstab and also a subdirectory on the device is bind mounted in fstab to a subdirectory of the home directory, then the sidebar shows two apparently identical entries for that device.

### Expected behavior
Only one entry for each mounted device should be shown under "Devices".  Bind mounts are accessible through the local filesystem.

"Data" is a separate NTFS partition mounted to `media/data`.  A Music folder on this device is also bind mounted to ~/Music.
image

## Platform Information
Hera
I'm using the latest released stable version

## Additional context
see https://elementaryos.stackexchange.com/questions/25263/show-hide-used-bar-and-or-entry-in-files-sidebar-for-specific-drives

jeremypw:
Fixes #1401


src/View/Sidebar.vala
```diff
@@ -895,9 +895,14 @@ public class Marlin.Sidebar : Marlin.AbstractSidebar {
             if (mount != null) {
                 /* show mounted volume in sidebar */
                 var root = mount.get_root ();
+                var device_label = root.get_basename ();
+                if (device_label != mount.get_name ()) {
+                    device_label = _("%s on %s").printf (device_label, mount.get_name ());
```
danrabbit, line 900: Maybe a translator comment?

jeremypw: Done.

PR #1407, opened 2020-07-28
Rename 'Personal' to 'Favorites' in Sidebar

### LINKED ISSUE #1406: Change "Personal" to "Bookmarks"
After some discussion with Scott, we realized that it would be a much clearer connection to users to refer to this section as "Bookmarks" instead of "Personal".

lainsce:
This fixes #1406 with my suggestion of using "Favorites" instead as it's less of a browser-ism.

Looks like:
favorites


lainsce:
That should be all places that mention something dealing with Bookmarks (user-facing) so I changed them to Favorites.

src/View/Sidebar.vala
```diff
@@ -580,7 +580,7 @@ public class Marlin.Sidebar : Marlin.AbstractSidebar {
         store.clear ();
 
         iter = add_category (PlaceType.BOOKMARKS_CATEGORY,
-                             _("Personal"),
+                             _("Favorites"),
                              _("Your common places and bookmarks"));
```
danrabbit, line 584: Maybe:

```suggestion
                             _("Your common places and saved files and folders"));
```

@cassidyjames ?

lainsce: Oh, that one passed by me :facepalm: 

Your suggestion makes sense. But I'll wait for Cass's OK.

cassidyjames: The two "and"s feels stilted. Don't we also typically avoid second-person? Uuuuuuuhhhhh… how about just

```suggestion
                             _("Saved folders and files"));
```

lainsce: But the "common places" part I think is important to mention, it's those "Documents, Recents, etc" options. :)

danrabbit: There's also "Recent" and "Home" which are not optional. I guess we don't _need_ to include that in the tooltip, but if we could figure something out about it that would be nice :)

cassidyjames: This is getting really hairy, but I think we've used "plus" before for clarity instead of two "and"s. But we're also using "place" and "folders and files," so maybe this could be simplified.

`Common places plus saved folders and files` or `Common and saved places` could work depending on how granular we want to get.

lainsce: >`Common places plus saved folders and files`

I like this approach. :D

danrabbit: Yeah I'm +1 to calling out "folders and files" specifically so users know they have save files as well

danrabbit: should it be "files and folders" though because alphabet?

lainsce: imo, folders and files retains some hierarchy that mimics the sort order Files does it.

lainsce: Going to go with `Common places plus saved folders and files` here since it seems like it's the preferred choice. :)

lainsce:
Ready for merge :D

PR #1409, opened 2020-07-28
Sidebar: Update network tooltip

cassidyjames:
Follow-up to #1407. Avoids second-person and adds some clarity.


src/View/Sidebar.vala
```diff
@@ -784,7 +784,7 @@ public class Marlin.Sidebar : Marlin.AbstractSidebar {
             /* ADD NETWORK CATEGORY */
             iter = add_category (PlaceType.NETWORK_CATEGORY,
                                  _("Network"),
-                                 _("Your network places"));
+                                 _("Remote storage available on the network"));
```
danrabbit, line 787: I'm not sure this is accurate because this includes local network devices and remote servers

cassidyjames: @danrabbit ah I suppose I was thinking "remote" as in remote from this machine, and "network" like, via the network. But I can understand the particularity here. Thinking on it...

cassidyjames: Would "via" the network actually solve this? Maybe:

```suggestion
                                 _("Remote storage available via a network connection"));
```

danrabbit: I'm not sure I like "Storage" here because this could be a computer on your local network or a shared network folder and yeah those things technically are places to store files, but when you say "Storage" I feel like it's a volume

cassidyjames: Ah yeah that also makes sense. u so smart. Is "files and folders" too wordy here?
```suggestion
                                 _("Files and folders available via a network connection"));
```

danrabbit: Maybe "Devices and folders"? I'm not sure you can save a specific file here, but it could be like "Sally's Laptop" and "Office SMB Share" right?

cassidyjames: @danrabbit we also use "Places" elsewhere. "Places available via a network"? :stuck_out_tongue: But I'm also fine with "Devices and folders".

PR #1411, opened 2020-07-28
Sidebar: format device tooltips

### LINKED ISSUE #1410: Better tooltip format for devices in the sidebar
We should make better use of formatting and multiples lines here. I feel like it would be a big improvement just to put the usage on a second line and use small, dim text (echoing the style of accel tooltip markup)

**/path/to/volume — ext3/ext4**
218 GB Free of 250 GB

danrabbit:
Screenshot from 2020-07-28 16 54 05@2x

Fixes #1410 

* If the rowref isn't valid, there's no point doing any of this, so check that first
* Get rid of a single-use method
* I don't really see a good reason to use stringbuilder here
* Make sure not to include `file://`


danrabbit:
Thoughts on prioritizing free space?

Screenshot from 2020-07-29 11 27 23

PR #1414, opened 2020-07-30
Sidebar interface

jeremypw:
First draft of Sidebar Interface.

- [x] Interface with plugins via libcore
- [x] Interface with Window
- [x] Test with CloudProviders plugin


jeremypw:
Closing as #1453 includes it anyway and has not been separately reviewed.

PR #1416, opened 2020-08-03
Favorites → Bookmarks

cassidyjames:
Follow-up to #1407 after discussion in Slack. Aligning on "Bookmarks" because you might "bookmark" e.g. your tax documents or divorce papers even if they're not your "favorite". :stuck_out_tongue:

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -1957,7 +1957,7 @@ namespace FM {
 
             var bookmark_menuitem = new Gtk.MenuItem ();
             bookmark_menuitem.add (new Granite.AccelLabel (
-                _("Add to Favorites"),
+                _("Add to Bookmarks"),
                 "<Ctrl>d"
             ));
             bookmark_menuitem.action_name = "common.bookmark";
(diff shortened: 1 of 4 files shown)


src/View/AbstractDirectoryView.vala
```diff
@@ -1957,7 +1957,7 @@ namespace FM {
 
             var bookmark_menuitem = new Gtk.MenuItem ();
             bookmark_menuitem.add (new Granite.AccelLabel (
-                _("Add to Favorites"),
+                _("Add to Bookmarks"),
```
cassidyjames, line 1960: So this used to just be "Bookmark". I think "Add to Bookmarks" is probably clearer (especially when translating), but I know either could work.

src/View/Sidebar.vala
```diff
@@ -568,9 +568,11 @@ public class Marlin.Sidebar : Marlin.AbstractSidebar {
...
         store.clear ();
 
-        iter = add_category (PlaceType.BOOKMARKS_CATEGORY,
-                             _("Favorites"),
-                             _("Common places plus saved folders and files"));
+        iter = add_category (
+            PlaceType.BOOKMARKS_CATEGORY,
+             _("Bookmarks"),
+             _("Common places plus saved folders and files")
+         );
```
cassidyjames, line 575: Modernized indentation as long as we were here.

PR #1417, opened 2020-08-04
Implement xdg-desktop-portal for FileChooser

tintou:
This is the initial work on implementing the FileChooser portal


cassidyjames:
@tintou awesome! Could you share what is incomplete here, if anything—or why it's a draft?

tintou:
We need to support all the option keys of the portal https://flatpak.github.io/xdg-desktop-portal/portal-docs.html#gdbus-org.freedesktop.impl.portal.FileChooser

tintou:
Ready for wider testing

danrabbit:
@tintou I've installed from your branch and restarted, but it doesn't appear any of the Flatpak'd apps I have are using the native file chooser. Can you list an example app that can be tested with?

tintou:
You can test the app with GNOME Maps if you import a layer for instance, the occurrences of FileChooserNative are quite rare though

jeremypw:
How much work would it be to add some kind of test harness?

tintou:
@jeremypw Unsure about that because the changes after this PR are probably going to be cosmetic so I don't really know

danrabbit:
Seems to work here! I think we probably want to iterate on this quite a bit with style etc and I'd expect we remove the Gtk module in a future branch as well, but to get the ball rolling I'm in favor of merging :)

jeremypw:
I guess I was imagining a small demo app that used all the portal features.

tintou:
> I guess I was imagining a small demo app that used all the portal features.

Oh yes, that would indeed be useful! I'll do that when I get some time.

jeremypw:
Presumably we would be able to provide a standalone FileChooserDialog after this is merged instead of the hacked Gtk one?  But then any app not using the portal would get the native Gtk FileChooser instead of the elementary one?

tintou:
Yes, that's the plan, we should:
 * remove the GTK module (as it won't be possible to do such things in GTK4 anyway).
 * move the "patches" to the portal dialog.
 * then reimplement a dialog with the widgets of elementary Files to not rely on the GtkFileChooserDialog internals.

filechooser-portal/Main.vala
```diff
@@ -0,0 +1,364 @@
...
+
+    if (show_version) {
+      print ("0.0 \n");
+      return 0;
+    }
+
+    loop = new GLib.MainLoop (null, false);
+
+    try {
+        var session_bus = GLib.Bus.get_sync (GLib.BusType.SESSION);
```
jeremypw, line 346: Not used?

jeremypw:
There are some long (and very long) lines amongst the new code.

jeremypw:
I've tried running `/usr/libexec/io.elementary.files.xdg-desktop-portal` in a terminal and then introspecting the system using
```
dbus-send --session    --dest=org.freedesktop.DBus  --type=method_call    --print-reply   /org/freedesktop/DBus    org.freedesktop.DBus.ListNames 
```
but cannot see the Filechooser portal methods.  I am not an expert in this area - am I doing something wrong?

I was trying to invoke the portal methods using dbus-send but without success.

tintou:
I've added a testing tool:
Build and install it with:
`ninja filechooser-portal/io.elementary.Files.PortalTester` (flatpak-builder and GNOME SDK required)
Then run it with `flatpak run io.elementary.Files.PortalTester`

jeremypw:
I had to build the flatpak manually (after running the two commands given above I got the error message `app/io.elementary.portalTester/x86_64/master not installed`.  The first command gave `[1/1] Generating portal-tester with a meson_exe.py custom command.`).

The app had a couple of problems on my system:
* If a file was actually loaded or saved, rather than cancelling the filechooser, the resulting dialog could not be cancelled and the app had to be killed to exit.

* Trying to open a folder gave the error `GTK_FILE_CHOOSER_ACTION_SELECT_FOLDER is not supported by GtkFileChooserNativePortal because portal is too old`.  Not sure if this is a problem with my system not being up to date though - will investigate.

The FileChooser shown was the unmodified GtkFileChooser rather than the one hacked by Files but presumably that is expected?

jeremypw:
The above commands worked if the `--user` option in the custom target was changed to `--system`.  The flatpak then got installed.  Note that my Files local repo is on a different partition to my home directory.

filechooser-portal/tests/PortalTester.vala
```diff
@@ -0,0 +1,100 @@
...
+                if (id == Gtk.ResponseType.ACCEPT) {
+                    var message_dialog = new Gtk.MessageDialog (
+                        window,
+                        Gtk.DialogFlags.MODAL,
+                        Gtk.MessageType.INFO,
+                        Gtk.ButtonsType.CLOSE,
+                        "This file has been opened: %s",
+                        filechooser.get_file ().get_path ()
+                    );
+                    message_dialog.show_all ();
```
jeremypw, line 43: Need to destroy dialog when closed

filechooser-portal/tests/PortalTester.vala
```diff
@@ -0,0 +1,100 @@
...
+                    filechooser.get_files ().foreach ((file) => { paths = string.join (", ", file.get_path (), paths); });
+                    var message_dialog = new Gtk.MessageDialog (
+                        window,
+                        Gtk.DialogFlags.MODAL,
+                        Gtk.MessageType.INFO,
+                        Gtk.ButtonsType.CLOSE,
+                        "These files have been opened: %s",
+                        paths
+                    );
+                    message_dialog.show_all ();
```
jeremypw, line 65: Need to destroy dialog when closed

jeremypw (changes requested):
See inline comments

jeremypw:
When running the portal test on Odin in a VM, the message dialogs do not appear - because it seems the filechooser response is not detected.  Not sure why this behaviour is different to that found when running on Ubuntu.

filechooser-portal/tests/PortalTester.vala
tintou: Run is not necessary anymore (see https://developer.gnome.org/gtk4/unstable/ch41s02.html#id-1.7.4.4.77 for the reasoning behind its removal in GTK4)

filechooser-portal/tests/PortalTester.vala
tintou: same as above

jeremypw (approved):

jeremypw:
I am in favour of merging this now and dealing with any issues arising in future PRs.

PR #1423, opened 2020-08-12
Replace LibUnity with Granite.Services.Application

### LINKED ISSUE #1038: Ditch LibUnity and Quicklists
In favor of Granite's new functions

danrabbit:
Fixes #1038


decathorpe:
I don't particularly care about plank being optional or not (since it's an integral component of Pantheon and not going away soon), but I think the optional support* should work both ways* if it's there :) Otherwise, just make it mandatory.

danrabbit:
afaict, plank isn't optional but we're checking for a recent enough version to use the poof window

tintou (approved):

PR #1436, opened 2020-08-14
BasicLocationBar : misc cleanup

andirsun:


danrabbit:
We cannot remove the copyright that previous authors have on code. The only reason to remove their copyright lines would be if the code they had written was also removed

PR #1437, opened 2020-08-14
Port get_link_name () and move to FileUtils

jeremypw:
The function `get_link_name ()` is ported to Vala from `marlin_file_operations.c` and moved to FileUtils.vala.

The format for multiple links has been simplified as discussed.

The associated function `shorten_utf8_string ()` was also ported.

Unit tests included.


libcore/FileUtils.vala
```diff
@@ -891,6 +891,81 @@ namespace PF.FileUtils {
...
+                    ///TRANSLATORS: %'d represents 22 or 32 or 42 etc
+                    result = _("%'dnd link to %s").printf (count, target_name);
+                    break;
+                case 3:
+                    ///TRANSLATORS: %'d represents 23 or 33 or 43 etc
+                    result = _("%'drd link to %s").printf (count, target_name);
+                    break;
+                default:
+                    ///TRANSLATORS: %'d represents a number in range 4 - 9, 14 - 19, 24 - 29 etc
+                    result = _("%'dth link to %s").printf (count, target_name);
```
danrabbit, line 958: This doesn't seem right at all for localization. I'm actually not sure of the right way to do this though. Maybe @ryonakano knows

danrabbit (dismissed):
While we're here, I think we should see if this is really the right way to localize these strings

libcore/FileUtils.vala
danrabbit: @jeremypw Maybe we can vastly simplify this and go for "FileName (link)", "FileName (link 2)", etc. I think I saw somewhere you were looking at the copy one too. We could apply the same pattern and do "FileName (copy)", "FileName (copy 2)", etc.

jeremypw: Yes, I am happy to do that - I was thinking there was a lot of fairly involved code for not very much gain.

libcore/FileUtils.vala
```diff
@@ -919,6 +919,52 @@ namespace PF.FileUtils {
...
+            case 0:
+                result = target_name; //First link in directory has same name as source
+                break;
+
+            case 1:
+                result = _("%s (link)").printf (target_name);
+                break;
+
+            default:
+                result = _("%s (link %i)").printf (target_name, count);
```
tintou, line 957:
```suggestion
                result = _("%s (link %d)").printf (target_name, count);
```
We're using %d everywhere ;)

tintou:
Ready to go on my side, neat that you added some tests!

libcore/FileUtils.vala
jeremypw: Naming now simpler

tintou (approved):

PR #1443, opened 2020-08-16
Port get duplicate name () to Vala and move to FileUtils

jeremypw:
Based on #1437, which should be reviewed first

The format for duplicates has been simplified and some code shared with getting link names.


libcore/marlin-file-operations.c
```diff
@@ -124,447 +124,125 @@ static void scan_sources (GList *files,
...
-    }
-}
-
-/* Note that we have these two separate functions with separate format
- * strings for ease of localization.
- */
-
-static char *
-get_link_name (const char *name, int count, int max_length)
+//~ <<<<<<< HEAD
```
tintou, line 128: Looks like a oops ;)

libcore/marlin-file-operations.c
```diff
@@ -124,447 +124,125 @@ static void scan_sources (GList *files,
...
 
-    parse_previous_duplicate_name (name, &name_base, &suffix, &count);
-    result = make_next_duplicate_name (name_base, suffix, count + count_increment, max_length);
 
-    g_free (name_base);
+    return name;
+//~ =======
 
-    return result;
+//~ >>>>>>> port-get-link-name
```
tintou, line 245: oops

libcore/marlin-file-operations.c
```diff
@@ -2360,14 +2039,26 @@ get_unique_target_file (GFile *src,
...
-                              G_FILE_ATTRIBUTE_STANDARD_EDIT_NAME,
-                              0, NULL, NULL);
+    info = g_file_query_info (
+        src,
+        g_strjoin (
+            ",",
+            G_FILE_ATTRIBUTE_STANDARD_EDIT_NAME,
+            G_FILE_ATTRIBUTE_STANDARD_TYPE,
+            NULL
+        ),
```
tintou, line 2049: that's a leak, you can use the const way here:
```suggestion
        G_FILE_ATTRIBUTE_STANDARD_EDIT_NAME "," 
        G_FILE_ATTRIBUTE_STANDARD_TYPE,
```

libcore/marlin-file-operations.c
```diff
@@ -2360,14 +2039,26 @@ get_unique_target_file (GFile *src,
...
+        G_FILE_QUERY_INFO_NOFOLLOW_SYMLINKS,
+        NULL,
+        NULL
+    );
+
     if (info != NULL) {
         editname = g_file_info_get_attribute_string (info, G_FILE_ATTRIBUTE_STANDARD_EDIT_NAME);
+        is_link = g_file_info_get_attribute_uint32 (info, G_FILE_ATTRIBUTE_STANDARD_TYPE) == G_FILE_TYPE_SYMBOLIC_LINK;
```
tintou, line 2057:
```suggestion
        is_link = g_file_info_get_file_type (info) == G_FILE_TYPE_SYMBOLIC_LINK;
```

libcore/FileUtils.vala
```diff
@@ -919,6 +919,142 @@ namespace PF.FileUtils {
         return result;
     }
 
+    public const string COPY_TAG = N_("(copy");
+    public const string LINK_TAG = N_("(link");
+    public const string CLOSING_COPY_LINK_TAG = N_(")");
```
tintou, line 924: In an ideal world, to make translators life easier, it would be something like :
```suggestion
    public const string COPY_TAG = N_("(copy %d)");
    public const string LINK_TAG = N_("(link %d)");
```
and use `_(COPY_TAG).split("%d",2)` to get the previous `COPY_TAG` and `CLOSING_COPY_LINK_TAG`.
I'm unsure that assuming the same closing tag would work for RTL languages.

tintou: I like the direction though ;)

tintou (changes requested):

libcore/FileUtils.vala
jeremypw: Yes RTl languages are problematic.  If you use `split` then wouldn't the closing tag be the __first__ item not the last?  Maybe to make translation easier there should be an opening tag as well as a closing tag, with translator comments provided to explain what function these characters serve.

libcore/marlin-file-operations.c
jeremypw: Fixed

libcore/marlin-file-operations.c
jeremypw: Fixed

libcore/FileUtils.vala
tintou: So you need an even more complicated string matching function with:
`%s (copy %d)`
and do some `split("%s",2)` and look for each part for a `%d`

jeremypw: So, am I correct in thinking strings  of text in RTL languages are stored "normally" (i.e. opening character first) and just displayed RTL?  I believe that in an RTL language an opening parenthesis character "(" is actually displayed as ")"?  Very confusing. :confused:

tintou: According to Google Translate: `File (Copy 1)` is translated to `ملف (نسخ 1)` in arabic

jeremypw: Just pushed a commit with a separate opening tag and translator comments.  This works OK using RTL in an English environment at least (using GTK_DEBUG=interactive). I guess there are some languages that do not use "(" but I do not know which.

jeremypw: Hold on, one of the tests fails.

jeremypw: Fixed the test

jeremypw:
Added a translators comment against another translated string "Copy" in AbstractDirectoryView which is a verb whereas in this PR "copy" is used as a noun so in some languages the translation may be different.

Do we need to use `NC_()` in this case?

tintou:
"Copy" is different from "copy" so no need for NC_ right now ;)

jeremypw:
Hm, Just tried it in Arabic using "Copy" as the tag. Get `new file (نسخ 2).txt` so seems to work.  Only thing is when the filename wraps in icon view the order of the  words gets mixed up but I do not think it is an issue with this PR.  Probably an issue with how Pango wraps mixed direction strings.

jeremypw:
This kind of fails for the corner case of an original file named "(copy)"  - creating a copy of this results in a file named "(copy 2)" not "(copy) (copy)".  However I am not sure it is worth accomodating perversely named files like this.

tintou:
Yeah i'd say that it's a convenience function which works for most people so let's ship it ;)

tintou (approved):

PR #1444, opened 2020-08-16
Port misc string functions to Vala and move to FileUtils

jeremypw:
Ports to Vala:
* custom_basename_from_file ()
* get_path_for_symlink ()

Removes some unneeded/unused code


jeremypw:
May not be needed or may be incorporated into another PR refactoring the construction of duplicate filenames.

PR #1447, opened 2020-08-19
Port and combine code relating to getting target files

jeremypw:
This combines and simplifies the code relating to getting target files when copying, linking and creating files and ports it to Vala.

Because of the intricacy of the original code, it is hard to ensure all corner cases are dealt with but it seems to work well enough in normal situations and when moving files to filesystems where the names are invalid.

As discussed, the naming of duplicates has been simplified and rationalised so that both copy and link duplicates have the same form  - "(copy xx)" or "(link xx)".

I am aware the way the duplicate format strings are constructed might be questionable but this was required to ensure that they could be parsed in any locale to be identified and to extract the digit part reliably.  Alternative suggestions welcome.

Some attempt was made to improve the efficiency of file operations by only determining the maximum name length for the destination file system once per operation instead of for each file.  The assumption is made that the source and destination filesystems do not change during an operation.

Also the opportunity was taken to remove all reference to "debuting files" which have not been used for years and add an overhead.

### DIFF
--- a/libcore/FileUtils.vala
+++ b/libcore/FileUtils.vala
@@ -971,22 +971,23 @@ namespace PF.FileUtils {
         return count;
     }
 
-    ///TRANSLATORS Just translate the word "copy" and leave the placeholders (%s) in their current positions
+    /* We assume all these translations will be valid for filenames in all filesystems */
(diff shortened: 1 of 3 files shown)


jeremypw:
Closing as this seems be superceded by, or overlap extensively with #1443

PR #1453, opened 2020-08-28
Sidebar listbox

### LINKED ISSUE #1375: Rewrite Sidebar as ListBox
ListBox seems better suited here for a few reasons:
* We never use any kind of nesting or headers or columns, so I don't think we use any treeview-specific-features
* We could pack sidebar sections into a revealer for animations
* We could animate expander image rotation
* It would probably be easier to layout and style complex widgets like drives with levelbars

jeremypw:
Closes #1375 
Building on #1414, this reimplements the Sidebar using mainly Gtk.ListBox.

The original Sidebar class has been divided into separate classes for Bookmarks, Devices, Network connections  lists and rows, for and Actions which implement the appropriate interfaces.  Although mainly Gtk.ListBox and Gtk.ListBoxRow have been used the interfaces do not require this so other widgets could be used.  The appearance and layout of the Sidebar is much the same as the original except the "Connect Server" action has been separated out and pinned to the bottom of the window as it is not really a bookmark item and so does not implement that interface.

 - [x] Populate Bookmarks, Devices and Network lists
 - [x] Implement storagebar on device bookmarks
 - [x] Implement tooltips (content to be finalized)
 - [x] Navigate to the appropriate location when clicked or activated with keyboard
 - [x] Sync selected bookmark with view
 - [x] Add bookmark from view context menu and shortcut
 - [x] Add bookmark by drag and drop
 - [x] Remove bookmark by drag and drop if not pinned or permanent
 - [x] Reorder bookmark by drag and drop within its own list if not pinned or permanent
 - [x] Change Trash icon according to state of trash
 - [x] Implement "Connect to Server" Action
 - [x] Implement add plugin item with callback
 - [x] Implement basic "Open in .."  and "Remove" context menu
 - [x] Implement additional context menu items and modify basic items according to row properties
 - [x] Sync expander states with settings
 - [x] Implement rename bookmarks
 - [x] Implement unmount devices
 - [x] Implement show device properties
 - [x] Implement plugin menu models
 - [x] Reflect trash empty/full status in icon
 - [x] Visual feedback of action on dropping during drag motion


jeremypw:
Appearance (on Ubuntu Focal with elementary themes)
Screenshot from 2020-08-28 12-04-48

jeremypw:
Appearance of device storage bars with level colors and NextCloud plugin context menu
Screenshot from 2020-08-29 20 23 53

jeremypw:
The storage bar is now a little thinner and more separated from the label. The original code is here: https://github.com/elementary/files/blob/master/src/View/Widgets/DiskRenderer.vala
The original bar height was 4px,  full > 0.9, warn > 0.75.
After #1455:
NewDiskRenderer

Original (master):
OriginalDiskRendererStyle

jeremypw:
When dropping onto sidebar (favorites section), the bookmarks (appear to) move apart to make space when dropping between.

hanaral:
Are there plans to use Gtk.Listbox for other apps with the sidebar? It could be a little inconsistent to have the Files sidebar with animations and the one in Code without.

jeremypw:
Music would be a better comparison than Code as the former, like Files, is not sorted and implements dragging/dropping rows. Both Code and Music use Granite.Widgets.SourceList for the sidebar, so an alternative question would be should Files use that widget, instead of a ListBox?

danrabbit:
We've attempted to write a modern Granite.Sidebar a few times, but so far none of the attempts have stuck. It might be worth revisiting if we can come up with something that works for Files, Music, Tasks, etc. I'm not sure if it's worth waiting for Gtk.ColumnView in Gtk4 or not. It does seem that having a full tree view as a sidebar is much less common

jeremypw:
Is there a specification for such a Granite.Sidebar?  It does seem quite a hurdle to get such a widget written and approved and then get elementary apps to switch to using it.

danrabbit:
@jeremypw not that I'm aware of. That would probably be a good first step 🙃

jeremypw:
Marking ready for review to attract wider testing.  I am concerned that the code may be over-abstracted in order to facilitate changing the underlying widgets used for the sidebar or to add new categories so suggestions on simplification welcome.  Even so, several hundred lines are saved - more could be saved with a more compact layout.

fabiozaramella:
Seems to work great, good job! 👍🏽️
Just a note on the cosmetic changes for now: categories headers are much taller with this PR than before and compared to what you see in Music and Code sidebars for example. The fading animation is also much slower. 
Apart from this everything works as expected. 😊️

jeremypw:
Yes, this is controlled by the css provided here: https://github.com/elementary/files/pull/1453/commits/8fed462ad52f9a8bc5a01d6c936f94b6ac16cdef

danrabbit:
The padding thing was my fault. I was trying to be too clever. Proposed a fix here: https://github.com/elementary/stylesheet/pull/881

I pushed a commit that should speed up animations a bit

danrabbit:
Found a couple of issues with Drag and Drop:

- [ ] It appears that bookmark order is not saved between sessions when rearranged with drag and drop
- [ ] Dragging a bookmark out of the sidebar to remove it only works if you drag it down, not up

jeremypw:
@danrabbit Latest commit should fix saving reordering of bookmarks.  I could not reproduce the drag out of sidebar issue - any failed drag/drop should result in removal of the bookmark.

jeremypw:
Looking at this again after a period away, there are some further simplifications possible.  I do not think that it is necessary to have separate code in the library to handle bookmarks - this could be incorporated into the BookmarkListBox so we only maintain a single model (list of bookmarks).  But that can wait for another PR if and when this is one is accepted.

danrabbit:
This is such a huge rewrite that I feel like we should merge it sooner than later so we can get wider testing and ensure there are no regressions. So far from what I can tell this is in pretty good shape UX-wise.

libcore/BookmarkList.vala
```diff
@@ -196,25 +201,29 @@ namespace Marlin {
...
+            bool list_changed = false;
+            unowned GLib.List<Marlin.Bookmark> node = list;
+            unowned GLib.List<Marlin.Bookmark> next = node.next;
 
-            if (destination > list.length ()) { // Can be assumed to be limited in length
-                critical ("Bookmarklist: Attempt to move bookmark to out of range index");
-                return;
-            }
+            int index = 0;
+            for (node = list; node != null; node = next) {
```
tintou, line 210:
```suggestion
            foreach (unowned var bm in list) {
```

libcore/FileUtils.vala
```diff
@@ -1108,6 +1106,19 @@ namespace PF.FileUtils {
...
+        string [] supported;
+
+        supported = GLib.Vfs.get_default ().get_supported_uri_schemes ();
+        for (int i = 0; supported[i] != null; i++) {
+            if (supported[i] == protocol) {
+                return true;
+            }
+        }
+
+        return false;
```
tintou, line 1120: return protocol in GLib.Vfs.get_default ().get_supported_uri_schemes ()

libcore/Interfaces/SidebarInterface.vala
```diff
@@ -0,0 +1,51 @@
+/***
+    Copyright (c) 2020 elementary LLC <https://elementary.io>
```
tintou, line 2:
```suggestion
    Copyright 2020 elementary, Inc <https://elementary.io>
```

src/Utils/AppUtils.vala
```diff
@@ -0,0 +1,33 @@
+/*
+ * AppUtils.vala
+ *
+ * Copyright 2020 Jeremy Paul Wootten <jeremy@jeremy-Kratos-Ubuntu>
```
tintou, line 4:
```suggestion
 * Copyright 2020 elementary, Inc. (https://elementary.io)
```

src/Utils/AppUtils.vala
```diff
@@ -0,0 +1,33 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
+ *
+ *
+ */
+
+namespace Marlin {
+    static Gtk.Window get_active_window () {
+        var gtk_app = (Gtk.Application)(GLib.Application.get_default ());
```
tintou, line 26:
```suggestion
        unowned Gtk.Application gtk_app = (Gtk.Application) GLib.Application.get_default ();
```

src/View/Sidebar/BookmarkListBox.vala
```diff
@@ -0,0 +1,234 @@
+/* BookmarkListBox.vala
+ *
+ * Copyright 2020 elementary LLC. <https://elementary.io>
```
tintou, line 3:
```suggestion
 * Copyright 2020 elementary, Inc. (https://elementary.io)
```

src/View/Sidebar/BookmarkListBox.vala
```diff
@@ -0,0 +1,234 @@
...
+                new ThemedIcon (Marlin.ICON_RECENT),
+                true
+            );
+
+            row.set_tooltip_markup (
+                Granite.markup_accel_tooltip ({"<Alt>R"}, _("View the list of recently used files"))
+            );
+        }
+
+        foreach (Marlin.Bookmark bm in bookmark_list.list) {
```
tintou, line 136:
```suggestion
        foreach (unowned Marlin.Bookmark bm in bookmark_list.list) {
```

src/View/Sidebar/BookmarkListBox.vala
```diff
@@ -0,0 +1,234 @@
...
+            }
+        });
+    }
+
+    public override bool add_favorite (string uri,
+                                       string? label = null,
+                                       int pos = 0) {
+
+        int pinned = 0; // Assume pinned items only at start and end of list
+        foreach (Gtk.Widget child in get_children ()) {
```
tintou, line 169:
```suggestion
        foreach (unowned Gtk.Widget child in get_children ()) {
```

src/View/Sidebar/BookmarkListBox.vala
```diff
@@ -0,0 +1,234 @@
...
+        if (bm != null) {
+            insert_bookmark (bm.label, bm.uri, bm.get_icon (), pos);
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    public override bool remove_item_by_id (uint32 id) {
+        foreach (Gtk.Widget child in get_children ()) {
```
tintou, line 191:
```suggestion
        foreach (unowned Gtk.Widget child in get_children ()) {
```

src/View/Sidebar/BookmarkListBox.vala
```diff
@@ -0,0 +1,234 @@
...
+            if (child is SidebarItemInterface) {
+                var row = (SidebarItemInterface)child;
```
tintou, line 193:
```suggestion
                unowned var row = (SidebarItemInterface)child;
```

src/View/Sidebar/DeviceListBox.vala
```diff
@@ -0,0 +1,241 @@
+/* DeviceListBox.vala
+ *
+ * Copyright 2020 elementary LLC. <https://elementary.io>
```
tintou, line 3:
```suggestion
 * Copyright 2020 elementary, Inc (https://elementary.io)
```

src/View/Sidebar/DeviceListBox.vala
```diff
@@ -0,0 +1,241 @@
...
+                true,  //Pinned
+                true   //Permanent
+            );
+
+            row.set_tooltip_markup (
+                Granite.markup_accel_tooltip ({"<Alt>slash"}, _("View the root of the local filesystem"))
+            );
+        }
+
+        foreach (GLib.Drive drive in volume_monitor.get_connected_drives ()) {
```
tintou, line 121:
```suggestion
        foreach (unowned GLib.Drive drive in volume_monitor.get_connected_drives ()) {
```

src/View/Sidebar/DeviceListBox.vala
```diff
@@ -0,0 +1,241 @@
...
+            bookmark_stoppable_or_removeable_drive_if_without_volumes (drive);
+        } // Add drives not otherwise bookmarked
+
+        foreach (Volume volume in volume_monitor.get_volumes ()) {
```
tintou, line 125:
```suggestion
        foreach (unowned Volume volume in volume_monitor.get_volumes ()) {
```

src/View/Sidebar/DeviceListBox.vala
```diff
@@ -0,0 +1,241 @@
...
+            bookmark_volume_if_without_mount (volume);
+        } // Add volumes not otherwise bookmarked
+
+        foreach (Mount mount in volume_monitor.get_mounts ()) {
```
tintou, line 129:
```suggestion
        foreach (unowned Mount mount in volume_monitor.get_mounts ()) {
```

src/View/Sidebar/DeviceListBox.vala
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    private DeviceRow? has_uuid (string? uuid, string? fallback = null) {
+        var search = uuid != null ? uuid : fallback;
+
+        if (search == null) {
+            return null;
+        }
+
+        foreach (var child in get_children ()) {
```
tintou, line 214:
```suggestion
        foreach (unowned var child in get_children ()) {
```

src/View/Sidebar/NetworkListBox.vala
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public void refresh () {
+        clear ();
+
+        if (Marlin.is_admin ()) { //Network operations fail for administrators
+            return;
+        }
+
+        foreach (Mount mount in VolumeMonitor.@get ().get_mounts ()) {
```
tintou, line 76:
```suggestion
        foreach (unowned Mount mount in VolumeMonitor.@get ().get_mounts ()) {
```

src/View/Sidebar/SidebarListInterface.vala
```diff
@@ -0,0 +1,100 @@
...
+        sidebar.path_change_request (item.uri, flag);
+    }
+
+    public abstract void refresh ();
+
+    public virtual uint32 add_plugin_item (Marlin.SidebarPluginItem plugin_item) {return 0;}
+
+    public virtual void clear () {
+        foreach (Gtk.Widget child in get_children ()) {
+            remove (child);
```
tintou, line 40: maybe move this after `destroy_bookmark` as it can be freed after calling `remove`

src/View/Sidebar/SidebarListInterface.vala
```diff
@@ -0,0 +1,100 @@
...
+    public virtual void open_item (SidebarItemInterface item, Marlin.OpenFlag flag = Marlin.OpenFlag.DEFAULT) {
+        sidebar.path_change_request (item.uri, flag);
+    }
+
+    public abstract void refresh ();
+
+    public virtual uint32 add_plugin_item (Marlin.SidebarPluginItem plugin_item) {return 0;}
+
+    public virtual void clear () {
+        foreach (Gtk.Widget child in get_children ()) {
```
tintou, line 39:
```suggestion
        foreach (unowned Gtk.Widget child in get_children ()) {
```

src/View/Sidebar/SidebarListInterface.vala
```diff
@@ -0,0 +1,100 @@
...
+                return;
+            }
+            remove (row);
+            row.destroy_bookmark ();
+        }
+    }
+
+    public virtual bool has_uri (string uri, out unowned SidebarItemInterface? row = null) {
+        row = null;
+        foreach (var child in get_children ()) {
```
tintou, line 60:
```suggestion
        foreach (unowned var child in get_children ()) {
```

src/View/Sidebar/SidebarListInterface.vala
```diff
@@ -0,0 +1,100 @@
...
+                    return true;
+                }
+            }
+        }
+
+        return false;
+    }
+
+    public virtual bool remove_item_by_id (uint32 id) {
+        foreach (Gtk.Widget child in get_children ()) {
```
tintou, line 73:
```suggestion
        foreach (unowned Gtk.Widget child in get_children ()) {
```

src/View/Sidebar/SidebarListInterface.vala
```diff
@@ -0,0 +1,100 @@
...
+            if (child is SidebarItemInterface) {
+                var row = (SidebarItemInterface)child;
```
tintou, line 75:
```suggestion
                unowned var row = (SidebarItemInterface)child;
```

src/View/Sidebar/SidebarListInterface.vala
```diff
@@ -0,0 +1,100 @@
...
+                if (row.permanent) {
+                    continue;
+                }
+                if (row.id == id) {
+                    remove (row);
```
tintou, line 80: Here too, better call remove after `destroy_bookmark` as `remove` decrease its reference count.

src/View/Sidebar/SidebarWindow.vala
```diff
@@ -0,0 +1,289 @@
...
+        if (network) {
+            network_listbox.refresh ();
+        }
+
+        loading = false;
+    }
+
+    /* SidebarInterface */
+    public uint32 add_plugin_item (Marlin.SidebarPluginItem plugin_item, Marlin.PlaceType category) {
+        uint32 id = -1;
```
tintou, line 145: I don't really like returning -1 with uint, can we find a better value like 0 or uint32.MAX (even if -1 is the same, that's more visible that way)

src/meson.build
```diff
@@ -56,6 +54,17 @@ pantheon_files_exec = executable (
...
+	'View/Sidebar/BookmarkListBox.vala',
+	'View/Sidebar/DeviceListBox.vala',
+	'View/Sidebar/NetworkListBox.vala',
+	'View/Sidebar/BookmarkRow.vala',
+	'View/Sidebar/DeviceRow.vala',
+	'View/Sidebar/NetworkRow.vala',
+	'View/Sidebar/PopupMenuBuilder.vala',
+	'View/Sidebar/SidebarListInterface.vala',
+	'View/Sidebar/SidebarItemInterface.vala',
+
```
tintou, line 67: Indentation looks wrong here

tintou (dismissed):
A few minor comments about ownership, I've only reviewed the code, I haven't tested it though

libcore/BookmarkList.vala
jeremypw: Done.  This was a remnant of the old direct port from C

libcore/FileUtils.vala
jeremypw: Much better!

src/Utils/AppUtils.vala
jeremypw: Hmm, left an auto-generated header.  Replaced with one copied from existing file

libcore/Interfaces/SidebarInterface.vala
jeremypw: Done - Need a separate PR to fix all the existing headers.

src/Utils/AppUtils.vala
jeremypw: Done

src/View/Sidebar/SidebarListInterface.vala
jeremypw: But `destroy_bookmark` calls `base.destroy ()` so isn't a problem to reference it afterwards?  If we do not use `foreach (unowned Gtk.Widget ...` then a reference will be placed on the child preventing it being prematurely freed?

src/View/Sidebar/SidebarListInterface.vala
jeremypw: See comment above

src/View/Sidebar/SidebarListInterface.vala
jeremypw: See comment above on similar issue

src/View/Sidebar/SidebarWindow.vala
jeremypw: I think we can use 0.  Real items have an id > 0.

hanaral:
Not sure if this has been fixed in the mean time, but I often have to click multiple times on a sidebar location, and cannot seem to middle/3 finger click to open in new tab at all.

hanaral:
There also seems to be some visual inconsistencies with the new sidebar's dropdown and the SourceList one. SourceList has a hover highlight, a smaller icon and is significantly slimmer than the new sidebar:
Screenshot from 2021-02-06 10-20-44
Screenshot from 2021-02-06 10-22-18

jeremypw:
@hanaral Thanks for the input.  The difference in appearance is because different widgets are used (Gtk.ListBox in Files and Granite.SourceList in Code) and I was trying to emulate the existing appearance.  I tried to make the code widget agnostic by using interfaces so it might be possible bas the sidebar on SourceList but I am not sure it is flexible enough.  Otherwise the some CSS magic might be able to reduce appearance if the design team approve.

jeremypw:
> and is significantly slimmer than the new sidebar:

Do mean slimmer in width, because that is user adjustable?

hanaral:
>> and is significantly slimmer than the new sidebar: ...

@jeremypw As in vertical thickness, the topmost drop down in the code side bar is about half the thickness of the new sidebar’s topmost drop down.

jeremypw:
OK, awaiting input from UX team.

danrabbit:
@hanaral @jeremypw any inconsistencies in listbox and treeview based sidebars would be an issue in the stylesheet :)

tintou (approved):
Looks good here, better land this sooner to catch regressions

PR #1454, opened 2020-08-29
Fix styling and animate

danrabbit:
Screenshot from 2020-08-29 11 27 58@2x

The expanders aren't animated here, but revealing the sections is! Not sure exactly how to deal with animating the arrows yet. Have to figure out what the cleanest way to do that is

### DIFF
--- a/src/View/Sidebar/BookmarkRow.vala
+++ b/src/View/Sidebar/BookmarkRow.vala
@@ -102,13 +102,14 @@ public class Sidebar.BookmarkRow : Gtk.ListBoxRow, SidebarItemInterface {
         var label = new Gtk.Label (custom_name) {
             xalign = 0.0f,
             halign = Gtk.Align.START,
-            hexpand = true,
-            margin_start = 6
+            hexpand = true
         };
 
         bind_property ("custom-name", label, "label", BindingFlags.DEFAULT);
 
-        label_stack = new Gtk.Stack ();
+        label_stack = new Gtk.Stack () {
+            homogeneous = false
+        };
         label_stack.add_named (label, "label");
         if (!pinned) {
             editable = new Gtk.Entry ();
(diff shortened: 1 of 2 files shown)


src/View/Sidebar/BookmarkRow.vala
```diff
@@ -126,12 +127,11 @@ public class Sidebar.BookmarkRow : Gtk.ListBoxRow, SidebarItemInterface {
 
         icon = new Gtk.Image.from_gicon (gicon, Gtk.IconSize.MENU) {
             halign = Gtk.Align.START,
-            hexpand = false,
-            margin_start = 12
+            hexpand = false
         };
 
         icon_label_grid = new Gtk.Grid () {
-            orientation = Gtk.Orientation.HORIZONTAL
```
danrabbit, line 134: Gtk.Grid is horizontal by default so this doesn't need to be set

src/View/Sidebar/BookmarkRow.vala
```diff
@@ -102,13 +102,14 @@ public class Sidebar.BookmarkRow : Gtk.ListBoxRow, SidebarItemInterface {
         var label = new Gtk.Label (custom_name) {
             xalign = 0.0f,
             halign = Gtk.Align.START,
-            hexpand = true,
-            margin_start = 6
```
danrabbit, line 106: it's better to use `column_spacing` in the grid so that don't have to add margins to each child

src/View/Sidebar/BookmarkRow.vala
```diff
@@ -124,14 +125,10 @@ public class Sidebar.BookmarkRow : Gtk.ListBoxRow, SidebarItemInterface {
         }
         label_stack.visible_child_name = "label";
 
-        icon = new Gtk.Image.from_gicon (gicon, Gtk.IconSize.MENU) {
-            halign = Gtk.Align.START,
-            hexpand = false,
-            margin_start = 12
```
danrabbit, line 130: Since the label is expanded, and Gtk.Widget.hexpand is false by default, this doesn't really do anything

src/View/Sidebar/BookmarkRow.vala
```diff
@@ -141,12 +138,8 @@ public class Sidebar.BookmarkRow : Gtk.ListBoxRow, SidebarItemInterface {
         };
         event_box.add (icon_label_grid);
 
-        content_grid = new Gtk.Grid () {
-            margin_top = 3,
-            margin_bottom = 3,
```
danrabbit, line 146: This is already taken care of in CSS in the sidebar style

src/View/Sidebar/BookmarkRow.vala
jeremypw: I think that may impact drag and dropping rows (the underlying list box would need to be made a drop target as well as the rows) but that may actually be cleaner

jeremypw (approved):
Thanks for the cleanup and new Expander :-)

PR #1456, opened 2020-08-31
SidebarWindow: Use button in actionbar instead of ActionRow

danrabbit:
Actionbars in Sidebars are not currently styled, but that's an issue with Greenfield

* Since this isn't in a listbox anymore, seems much simpler to use a regular Gtk.Button.
* Added `…` since this opens a new dialog before creating a new item
* Removed redundant text from the tooltip


jeremypw (changes requested):
Makes sense.  My only query is whether the button should remain focused after clicked and whether it should fill the whole row.

If there are more actions in future we might want to change to a Code-like actionbar.

danrabbit:
Yeah that's just default focus handling in Gtk

I don't think we have any precedent for expanding buttons in actionbars to fill the entire space

Can you elaborate on what you mean by there being a difference in Code's actionbar?

jeremypw:
The focussing issue can be fixed (if wished) by the setting `can-focus` property to `false`. There doesn't seem to be a good reason to focus on this button. If the button does not focus it does not matter if it does not hexpand.

The Code Actionbar is thinner, has only icons (with tooltips) and is paler. The focus issue is present but less obtrusive.

Files Actionbar (focused)
FilesActionBar

Code Actionbar (focused)
CodeActionBar (1)

danrabbit:
Ah okay I see. Yes there's a missing style class. I'll fix that

Interactive elements needs to be focusable for keyboard navigation :)

jeremypw:
Oh I forgot that :disappointed:  There is a keyboard shortcut but its not very discoverable with the keyboard I guess.

danrabbit:
Okay, I proposed a branch to style flat actionbars in Greenfield here: https://github.com/elementary/stylesheet/pull/800

I also moved the flat class to the actionbar and made the icon symbolic. This is how it looks in Greenfield:

Screenshot from 2020-09-01 11 42 19@2x

I'm sure it's rather broken in the old stylesheet :sweat_smile:

cassidyjames:
@jeremypw also fwiw we're slowly but surely adding action text to those inline toolbar actions to make them more discoverable and accessible without a tooltip. So what's more likely is for Code to change to look a little more like this ;)

danrabbit:
Updated with the latest Greenfield revision for less vertical padding

Screenshot from 2020-09-01 16 36 15@2x

jeremypw (approved):
Yes, that looks much better - thanks.

PR #1457, opened 2020-08-31
DeviceRow: Make Eject a Gtk.Button

danrabbit:
* Make "Eject" into a Gtk.Button so we can easily set tooltip and support keyboard navigation
* Add `SLIDE_LEFT` transition to the revealer so it won't reserve space when not used such as for the filesystem
* Add `CROSSFADE` transition to the stack so we get a fade between the eject button and the spinner

### DIFF
--- a/src/View/Sidebar/DeviceRow.vala
+++ b/src/View/Sidebar/DeviceRow.vala
@@ -101,39 +101,29 @@ public class Sidebar.DeviceRow : Sidebar.BookmarkRow, SidebarItemInterface {
     }
 
     construct {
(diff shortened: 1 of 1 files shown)


jeremypw (commented):
A couple of things look odd to me
* The eject button appears focused after mounting (the row is already selected so this seems unnecessary)
* I do not like the storage bar changing length depending on presence/absence of eject button
* The eject button slides in but just disappears rather than slides out.

I am not sure whether the storage bar should persist after unmounting a device

PR #1459, opened 2020-09-03
FileOperations: Reduce the use of GIOSchedulerJob for dialogs and remove the io_job field

tintou:
This is a simple step to simplify the port to modern GTask API and to Vala


jeremypw (commented):
LGTM and works as expected (tested conflict dialog and a couple of error situations)

jeremypw (approved):

PR #1461, opened 2020-09-03
Release Action: Update to use odin branch

cassidyjames:
If we're Odin-only, then this shouldn't be releasing to Juno… right?


danrabbit (approved):
Indeed

PR #1462, opened 2020-09-06
FileOperations: finish the port to the GTask API

tintou:


jeremypw (approved):
Works fine afaict.  There is a problem with the Files process getting held indefinitely after some file operations but this is not due to this PR.

PR #1466, opened 2020-09-09
Stop caching pathbar icons

### LINKED ISSUE #1465: Pathbar icons don't use correct text color
Screenshot from 2020-09-08 18 25 56@2x

It appears that the pathbar icon color isn't updated when the theme changes from light to dark.

This is a blocker before Files can support the dark style

jeremypw:
Fixes #1465 

It was simpler to stop caching the icon pixbufs than try to efficiently keep them in sync with the theming.


danrabbit (approved):
Nice simplification! Fixes the linked issue

PR #1470, opened 2020-09-11
Stop hyphenation in TextRenderer

### LINKED ISSUE #1348: Dashes in wrapped filenames
It looks like a change in GTK for focal added dashes for text wrapping. This might make sense in some contexts, but it is strange and confusing when displaying a filename. I ends up looking like you've accidentally added stray characters.

jeremypw:
Partially fixes #1348 

The file names are no longer hyphenated in Icon View but when renaming, the TextView widget used continues to insert hyphens.  There was no obvious way to fix this in Gtk3 but Gtk4 adds an extra TextTag attribute that can turn off insertion of hyphens.

Can anyone suggest a way of stopping auto-hyphenation during renaming in IconView?

### DIFF
--- a/src/TextRenderer.vala
+++ b/src/TextRenderer.vala
@@ -262,6 +262,12 @@ namespace Marlin {
                 connect_widget_signals ();
                 context = widget.get_pango_context ();
                 layout = new Pango.Layout (context);
+
+                /* We do not want hyphens inserted when text wraps */
+                var attr = new Pango.AttrList ();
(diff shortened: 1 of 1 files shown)


jeremypw:
I going to go ahead and mark this partial fix as "Ready for Review" as no simple fixes for the name editing have emerged so far.  That can be done at a later date.

tintou (approved):

tintou:
That actually looks good and minimal

PR #1472, opened 2020-09-12
Update search view markup before each search.

### LINKED ISSUE #1471: Search result header colors are cached on open
Screenshot from 2020-09-11 14 10 27@2x

This is probably a hardcoded style instead of using `Granite.STYLE_CLASS_H4_LABEL`

jeremypw:
Fixes #1471 

The appearance of the text in the search view is controlled using Pango markup derived from the "placeholder_text_color"  from the style_context.  Instead of doing this once on construction, it is now done every time the search view is used.

Probably should be done with custom css in future.

Some related single use functions are also combined.


danrabbit (approved):
Looks good, nice work :+1:

PR #1495, opened 2020-10-06
Removed BountySource

StalkyTalky:
The elementary team no longer uses BountySource, and since the README.md is potentially one of the the first impressions a user (or, in this case, a developer) will have, it may lead to confusion. I have removed BountySource from the README file for this reason.


danrabbit (approved):
Hey thank you for this cleanup PR!

Issue #1496, opened 2020-10-08
Mint and Bubblegum color tags are missing in Odin


fabiozaramella:
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Feature
As the title says, new Mint and Bubblegum colors from the palette are missing from Files color tags section.

![Schermata del 2020-10-08 15 21 05](https://user-images.githubusercontent.com/5459945/95479187-03d6c900-097a-11eb-9d04-ae6fac2d3a1f.png)

**Existing work**
Similar to this: https://github.com/elementary/calendar/pull/606

**Additional context**
On Odin daily.

jeremypw:
Fixed by #1497

PR #1497, opened 2020-10-08
Add Mint and Bubblegum to color tags

fabiozaramella:
Related to #1496


jeremypw (approved):
Does the job - thanks!

4jNsY6fCVqZv:
@fabiozaramella May I ask why you used other hex color values than Tasks or Calendar do?

fabiozaramella:
@4jNsY6fCVqZv Simply because I thought that using the corresponding 100 variant from brand palette was to faded (see banana, lime, blueberry and grape for example where the 300 hex is used) whereas for more darker colors (namely: strawberry, orange, grape, slate) 100 hex is better for legibility reasons.

PR #1500, opened 2020-10-12
Popover breadcrumbmenu

jeremypw:
Related to https://github.com/elementary/applications-menu/issues/423

This avoids a recent problem with using a Gtk.Menu in connection with the headerbar by switching to a Popover built from a GLib.MenuModel.  In order to do this some extra actions were created for Marlin.View.Window for opening paths in new tabs, windows or with an app which could be linked to the GLib.MenuItems.

This results in a basic UI - without icons.  

To get a more complicated UI it will probably be necessary use widgets like ModelButton to construct the menu, but this shows that Gtk.Popover is not affected by the problem of limited height.

### DIFF
--- a/libcore/FileUtils.vala
+++ b/libcore/FileUtils.vala
@@ -181,6 +181,7 @@ namespace PF.FileUtils {
         return result;
     }
 
+	/* Note: For the root directory this returns the root not the parent */
     private string construct_parent_path (string path, bool include_file_protocol) {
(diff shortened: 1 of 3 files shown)


jeremypw:
Not sure why Gtk.Popover.from_model () does not honor the icon attributes in the model - is that expected?  Do we want to keep app icons in the view context menu?

PR #1512, opened 2020-10-27
Remove the GTK module but apply the changes in the filechooser portal

tintou:
This removes the GTK module.
The filechooser portal still has the Files look a feel so that Flatpak applications using the portal are 1st class citizen.


danrabbit:
Kill it with fire

jeremypw:
This does not actually remove the filechooser module from the system if it is already installed - it just does not install it.  If you delete the filechooser module from `/usr/lib/x86_64-linux-gnu/gtk-3.0/modules` then you just get the native Gtk FileChooser, not the Files one.

jeremypw:
Hmm, the PortalTester flatpak is also bringup the normal GtkFileChooser, not the Files one?

tintou:
Yeah the packaging is taking care of removing the old files.
The LegacyFileChooserDialog won't be used if there is already the [pantheon-filechooser-module GTK module enabled](https://github.com/elementary/default-settings/blob/755430290564ac18f07079622cbd13126afa7c23/profile.d/pantheon-filechooser-module.sh) to avoid live-patches clashes.

jeremypw:
@tintou Could you look into issue #1518 before this is merged? Thanks.

jeremypw:
I notice that three xdg-desktop-portals are running:  `io.elementary.files.xdg-desktop-portal`, `xdg-desktop-portal-gtk` and `xdg-desktop-portal` - is this intended?  It is the Files portal that actually processes the request from the Epiphany flatpak mentioned above (as judged by cpu cycles).  However the appearance of the FileChooser dialog is not the modified Files one.

jeremypw:
Also need to fix #1520 before this is merged.

jeremypw:
This should be OK to merge now, but need to dogfood it a while first.

jeremypw:
@tintou So, after merging this, what needs to be done so that the Files filechooser is consistently shown (or is that not possible?).  Or is it intended that only apps using the portal will show the Files filechooser?

danrabbit:
It sounds like FileChooserNative only uses the portal in Flatpak: https://developer.gnome.org/gtk3/stable/gtk3-GtkFileChooserNative.html

I wonder if there's an implementation that always uses the portal in libportal or if we should add a granite util that prefers the portal and only falls back to the Gtk file chooser?

tintou:
It is possible to force the use of the portal outside the container by setting `GTK_USE_PORTAL=1` in the environment. Note that the portal will only be used if the app itself is using the GtkFileChooserNative dialog.

tintou:
See https://github.com/elementary/default-settings/pull/204

jeremypw (approved):
I think it is reasonably safe to merge this now - we can deal with any unexpected consequences before the next release.

PR #1513, opened 2020-10-28
Show properties overlaybar on selection, not on hover.

### LINKED ISSUE #1389: Display overlay bar on selection
As raised in #1387. Opening for thoughts. I think it makes sense, as it makes that data more accessible, especially since it no longer requires a mouse to display it. Thoughts @danrabbit?

jeremypw:
Fixes #1389 

Loses the `item-hovered` signal so overlaybar does not show until a selection is made.  Also fixes an issue where the selection change caused by rubberband selection in ListView not being picked up.


jeremypw:
Closing - will be rewritten

PR #1517, opened 2020-11-09
Scroll to selected file after loading, regardless of sort order.

### LINKED ISSUE #1187: Following hotlinked file reference makes the view jump to the wrong place
When I click on a link that refers to a file somewhere in the system, like this: `file:///home/username/Downloads/filename.txt` (which is used when clicking on downloaded files in the browser for example), the elementary files app is opened and it takes me straight to the file. The file viewer automatically scrolls to the corresponding file which is then auto-selected.

This works as it should as long as the default sorting order is active (sort by name). When this is changed to, let's say, "sort by date modified", the file viewer still scrolls to the location where the file would be if all files were sorted by name and not where it actually is according to the new sorting order.

os: elementaryos 5.1

jeremypw:
Fixes #1187 

AbstractDirectoryView: Set sort order before thawing tree after loading

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -1356,8 +1356,6 @@ namespace FM {
             in_recent = slot.directory.is_recent;
             in_network_root = slot.directory.file.is_root_network_folder ();
 
-            thaw_tree ();
-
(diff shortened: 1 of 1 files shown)


fabiozaramella (approved):
👍🏽️

Issue #1518, opened 2020-11-10
Filechooser Portal can't select files or folders


Marukesu:
## What Happened

When trying to open a file in epiphany or flatpak-ed code, i can't select any file or enter any folder on the portal. Also, if i cancel or close the portal, the application freeze.

## Expected Behavior

That i can open a file in the portal.

## Steps to Reproduce

1. Open flatpak-ed code
2. Try to open a file from the portal.
3. Can't select any file.

## Platform Information

OS: elementary OS next 6.0
Flatpak Version: 1.6.5
xdg-desktop-portal Version: 1.6.0

jeremypw:
Confirmed with the Flathub Epiphany flatpak using import bookmarks dialog.

cassidyjames:
Also confirmed with Firefox Flatpak

Marukesu:
Closing since #1519 fixed it.

PR #1519, opened 2020-11-14
filechooser-portal: Make sure that we are getting the response signal before destroy

tintou:
Also use the GtkDialog before the end of the response block as it will be destroyed.


tintou:
Fixes #1518

jeremypw:
This works when opening a file (using flatpak'd version of code) however, when trying to use it to save a file with Code I get: `Document.vala: File "(null)" saved successfully` and the tab label becomes blank.  No file is saved (where expected).   Not sure how much of this due to an issue with the Code flatpak though.  I'll try with a different flatpak.  Note that this test was performed __after__ removing the pantheon-filechooser module to ensure that the portal was used.
`

tintou:
@jeremypw that might be another bug, I'm fixing the open file method here, not the save one so that might be something to look at afterward

jeremypw:
OK, I'll merge this then and file another issue re saving files.

jeremypw (approved):
Fixes opening files

PR #1521, opened 2020-11-18
fm-list-model: Avoid hitting the wrong part of the memory with color management

tintou:
There is no warranty that the color isn't out of band here, add sanity checks.


jeremypw (approved):
Makes sense

PR #1522, opened 2020-11-20
Remove the -dbg package

tintou:
It is now automatically handled by the packaging process.


davidmhewitt (approved):
Can confirm that debug packages are still built

PR #1523, opened 2020-11-20
fm-list-model: Use a modern GObject definition for FMListModel

tintou:
Also add several sanity checks to all functions and methods.

### DIFF
--- a/libcore/fm-list-model.c
+++ b/libcore/fm-list-model.c
@@ -62,7 +62,7 @@ static void     fm_list_model_tree_model_init (GtkTreeModelIface *iface);
 static void     fm_list_model_drag_dest_init (GtkTreeDragDestIface *iface);
 static void     fm_list_model_sortable_init (GtkTreeSortableIface *iface);
 
-struct FMListModelDetails {
+typedef struct {
     GSequence *files;
     GHashTable *directory_reverse_map; /* map from directory to GSequenceIter's */
     GHashTable *top_reverse_map;       /* map from files in top dir to GSequenceIter's */
@@ -74,7 +74,7 @@ struct FMListModelDetails {
     GtkSortType     order;
 
     gboolean sort_directories_first;
-};
+} FMListModelPrivate;
 
 typedef struct FileEntry FileEntry;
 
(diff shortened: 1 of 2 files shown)


libcore/fm-list-model.c
```diff
@@ -839,25 +860,34 @@ fm_list_model_file_changed (FMListModel *model, GOFFile *file,
 gboolean
 fm_list_model_is_empty (FMListModel *model)
 {
-    return (g_sequence_get_length (model->details->files) == 0);
+    FMListModelPrivate *priv = fm_list_model_get_instance_private (model);
+
+    g_return_val_if_fail (FM_IS_LIST_MODEL (model), TRUE);
```
jeremypw, line 865: Shouldn't this come first?

libcore/fm-list-model.c
```diff
@@ -839,25 +860,34 @@ fm_list_model_file_changed (FMListModel *model, GOFFile *file,
...
+    return (g_sequence_get_length (priv->files) == 0);
 }
 
 guint
 fm_list_model_get_length (FMListModel *model)
 {
-    return g_sequence_get_length (model->details->files);
+    FMListModelPrivate *priv = fm_list_model_get_instance_private (model);
+
+    g_return_val_if_fail (FM_IS_LIST_MODEL (model), 0);
```
jeremypw, line 875: Should come first?

libcore/fm-list-model.c
```diff
@@ -946,45 +981,49 @@ fm_list_model_clear_directory (FMListModel *model, GSequence *files)
...
     }
 }
 
 void
 fm_list_model_clear (FMListModel *model)
 {
-    g_return_if_fail (model != NULL);
+    FMListModelPrivate *priv = fm_list_model_get_instance_private (model);
+
+    g_return_if_fail (FM_IS_LIST_MODEL (model));
```
jeremypw, line 994: Should come first?

libcore/fm-list-model.c
```diff
@@ -1226,17 +1258,21 @@ fm_list_model_sortable_init (GtkTreeSortableIface *iface)
...
     g_return_if_fail (FM_IS_LIST_MODEL (model));
 
-    model->details->has_child = has_child;
+    priv->has_child = has_child;
 }
 
 static void
 fm_list_model_set_icon_size (FMListModel *model, gint size)
 {
+    FMListModelPrivate *priv = fm_list_model_get_instance_private (model);
```
jeremypw, line 1271: Should come after testing model is FMListModel?

tintou:
@jeremypw foo_bar_get_instance_private is just a macro doing pointer offsets, it is safe to do it before testing the object.

jeremypw:
This does not seem to leak any more memory than master when navigating around, creating and destroying models and no regressions found so far so looking good. I'll test a while before approving though.

jeremypw (approved):
AFAICT as a C non-expert the new code is equivalent to the old so not regressions are expected.  No regressions were found on limited testing - I think this is safe to merge.

PR #1527, opened 2020-12-04
filechooser-portal: Make use of native async capabilities and process the dialog before destroy

### LINKED ISSUE #1520: FilechooserPortal cannot save files
## What Happened

When testing the elementary filechooser portal (`io.elementary.files.xdg-desktop-portal`) both with the supplied portal tester and with the Code flatpak, attempts to save a file results in a `null` filename being returned and no actual file being saved (at least not in the expected position).

## Expected Behavior

It should be possible to save files at any permissible position from a flatpak through the portal (or a warning issued for non-permissible locations)

## Steps to Reproduce

On elementary 6.0:
1. Build and install the portal tester flatpak.
2. Run the portal test and click the "Save" button.
3. Enter a filename into the filechooser and save.
4. The portal test reports a file named "null" was saved

Screenshot from 2020-11-16 16 02 50

5. Build and install io.elementary.code flatpak
6. Make a new test document
7. Press "Save At" button
8. Enter a filename into the filechooser and save
9. The document is not saved where expected and the document tab is blank
Screenshot from 2020-11-16 15 50 34

## Platform Information
Elementary 6.0 in VirtualBox with the filechooser-module removed/inactivated
Screenshot from 2020-11-16 16 02 50

tintou:
We were previously using some fields of the GtkFilechooser after it has been destroyed.
Fixes #1520

### DIFF
--- a/filechooser-portal/Main.vala
+++ b/filechooser-portal/Main.vala
@@ -38,7 +38,7 @@ public class Files.FileChooser : GLib.Object {
         this.connection = connection;
     }
 
-    public void open_file (GLib.ObjectPath handle, string app_id, string parent_window, string title, GLib.HashTable<string, GLib.Variant> options, out uint response, out GLib.HashTable<string, GLib.Variant> results) throws GLib.DBusError, GLib.IOError {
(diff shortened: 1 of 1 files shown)


jeremypw (approved):
I can confirm this allows saving of files through the portal

PR #1528, opened 2020-12-05
marlin-file-operations: Move the Dialog utilities to DialogUtils

tintou:
Allows to reduce the amount of code in marlin-file-operations.
Also fix a few C warnings on the go.

### DIFF
--- a/libcore/FileOperations/DialogUtils.vala
+++ b/libcore/FileOperations/DialogUtils.vala
@@ -1,0 +1,182 @@
+namespace PF {
+    private static int run_simple_dialog_va (Gtk.Window? parent_window,
+                                             GLib.Timer time,
+                                             PF.Progress.Info info,
+                                             Gtk.MessageType message_type,
+                                             owned string primary_text,
+                                             owned string secondary_text,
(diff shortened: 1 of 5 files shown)


jeremypw (changes requested):
Thanks for working on this!
Works OK for file conflicts and lack of permissions but seems to silently stop permanent deletion of a file when running as administrator - both with the keyboard and the context menu.  Nothing happens in both cases.

Is there any reason you cannot merge the `get_simple_file_operation_dialog` into FIleOperations/DialogUtils.vala? It is only used once there and no where else.

tintou:
@jeremypw: Oops, indeed it was because I translated it to `string details_text` when the details_text is NULL in this case, fixed.

I haden't looked at `get_simple_file_operation_dialog` at first but I've now removed it completely and integrated in `run_simple_dialog_va` directly which slightly simplifies the code (no struct, less string copy…)

jeremypw (approved):
I have found some instabilities in the file conflict dialog but at least some of them are present in master.  I think it will be easier to merge this and tackle any problems with that dialog afterwards.

PR #1529, opened 2020-12-07
fm-list-model: Add FMListModel creation method

tintou:
Allows Vala to know that the created object type is a Gtk.Widget subclass.


jeremypw (approved):
Makes sense - no regressions found or likely.

PR #1536, opened 2020-12-13
ListModel: Use the native Vala syntax to deal with properties

tintou:
Also make sure that we are using it everywhere.

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -590,7 +590,7 @@ namespace FM {
         }
 
         public void select_gof_file (GOF.File file) {
-            var iter = Gtk.TreeIter ();
+            Gtk.TreeIter iter;
             if (!model.get_first_iter_for_file (file, out iter)) {
                 return; /* file not in model */
             }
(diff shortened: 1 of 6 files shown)


src/View/AbstractDirectoryView.vala
```diff
@@ -3674,8 +3631,8 @@ namespace FM {
 
             var info = new GLib.FileInfo ();
             var dir = slot.directory;
-            string sort_col_s = get_string_from_column_id (sort_column_id);
-            string sort_order_s = (sort_order == Gtk.SortType.DESCENDING ? "true" : "false");
+            unowned string sort_col_s = ((FM.ListModel.ColumnID) sort_column_id).to_string ();
```
jeremypw, line 3634: This is not equivalent code.  Need to check how it affects restoring sort column from existing meta-data.

tintou: It looks to provide really similar results though: https://github.com/elementary/files/blob/master/libcore/fm-list-model.c#L1334

jeremypw: Ah, OK - I thought it was going to change the sort-column-id metadata to "0", "1", "2" but now I see that because of the cast it will not change the metadata.

jeremypw (approved):
Looks good - no regressions.

PR #1547, opened 2021-01-05
ListModel: port to Vala

tintou:
This is a simple rebase of https://github.com/elementary/files/pull/373
The same issues remains.


libcore/ListModel.vala
```diff
@@ -0,0 +1,717 @@
+/* Copyright (c) 2018 elementary LLC (https://elementary.io)
```
jeremypw, line 1: Need to update the date.

libcore/ListModel.vala
```diff
@@ -0,0 +1,717 @@
...
+        FILE_COLUMN,
+        COLOR,
+        PIXBUF,
+        FILENAME,
+        SIZE,
+        TYPE,
+        MODIFIED,
+        NUM_COLUMNS;
+
+        public static ColumnID from_string (string colid) {
```
jeremypw, line 30: No reason to use a weird identifier like `colid` - could use `column_id`

jeremypw (changes requested):
Just a couple of very minor code comments for now.  

Do we need/want to support our own (re)implementation of TreeModel?  Is this faster/more efficient than using the Gtk implementation?

tintou:
@jeremypw You're more familiar with the Files codebase than I am, but if we can use the Gtk model, then we should probably do it. If there is any performance issue I'm pretty confident that it shouldn't be the case anymore when we switch to GTK4

jeremypw:
Well, comparing the old experimental PR I did using Gtk.TreeStore (#824), you seem to have saved almost as many lines of code with this PR. So on reflection, I am not sure whether it worth the effort and risk of regression/performance issues trying to switch.  So I would be happy to merge this once the issues with the expanders and keyboard shortcuts are fixed.

libcore/ListModel.vala
```diff
@@ -0,0 +1,717 @@
...
+        var path = get_path (iter);
+        if (replaced_dummy) {
+            row_changed (path, iter);
+        } else {
+            row_inserted (path, iter);
+        }
+
+        if (file.is_folder ()) {
+            var file_entry_files = new GLib.Sequence<FileEntry?> ();
+            file_entry.files = file_entry_files;
```
jeremypw, line 334: This does not work.  The new `Sequence<FileEntry?>` gets freed as soon as the clause exits.  This is why no expanders appear on folders in ListView. However the obvious port of the C `file_entry.files = new Sequence<FileEntry?>()` gives an error both when file_entry.files is owned (duplicating a sequence) or unowned (assigning owned to unowned).  Not sure what the solution is to this.  Is Vala behaving correctly here?  May need to use a different structure for FileEntry and/or Sequence

jeremypw (changes requested):
See inline

jeremypw:
@tintou I suggest you close this one and propose the `tintou/listmodel-simple` branch.

PR #1548, opened 2021-01-05
build: use define_variable

worldofpeace:
On NixOS all packages are installed into their own immutable prefix.
Because of this systemd_dep.get_pkgconfig_variable will return a
path from within systemd's prefix and we cannot write to it.
By using define_variable we can replace the prefix to be from the
paths from meson. This should have no affect on elementary OS.


tintou (approved):

worldofpeace:
:sparkles:

PR #1563, opened 2021-01-21
Notify user when a removable media is plugged in

### LINKED ISSUE #1562: Notify user when a removable media is plugged in
## Problem
When plugging in a USB/removables, the user doesn't know it's connected until they open Files and see that their drive/volumes are connected.

## Proposal
Send a notification to let the user know that a drive/volume was added instead of silently adding it to Files' sidebar. Perhaps something like (see linked PR)

Screenshot from 2021-01-21 18-47-42

### Windows
Sends a notification and displays an indicator for ejecting removable media.
Screenshot from 2021-01-21 18 50 03

### GNOME
Sends a notification on detection of each new volume
Screenshot from 2021-01-21 19-48-14
Screenshot from 2021-01-21 19-48-21

### Mac
Not entirely sure about Mac but I _think_ it just shows it on the desktop silently (need confirmation).

pongloongyeat:
Closes #1562 

This is a draft but I'm not sure if I should mark it as such as I'd like someone to first review/discuss if showing a notification is even a good way of letting users know a removable usb, etc. was plugged in.

Demo:
image

TODO:
- [ ] Launch files (and mount) volumes on notification click.
- [ ] There are some possibly broken icons (elementary/icons#1000)
https://github.com/elementary/files/blob/676be22bf750b6607789f95a137f30aa52be10e6/src/View/Sidebar.vala#L838-L856
- [x] <strike>Make files run in the background.</strike> Make use of `files-daemon`
- [x] <strike>Make a nice dialog to request to launch in background on startup.</strike>


pongloongyeat:
My stupid self just realised this only works when files is open. Closing this.

pongloongyeat:
Reopening as I think I can instead run files in the background and launch on startup to provide this feature. Feel free to comment if there's a better approach.

jeremypw:
It may be better to start io.elementary.files-daemon in the background rather than the Files app.  This approach was used in a past PR to provide unity quicklists without starting Files althought that never gained traction.  Either way, it requires changing the default elementary startup configuration as well so cannot be fixed with just a Files PR.  It may be easy to use a process that is already started automatically (e.g. a wingpanel indicator) to provide these notifications.  However, it is up to the design team which is the preferred approach.

PR #1572, opened 2021-01-31
Use ngettext properly in OverlayBar

jeremypw:


src/View/Widgets/OverlayBar.vala
```diff
@@ -208,19 +208,19 @@ namespace Marlin.View {
...
                     if (files_count > 0) {
-                        str += ngettext (_(" and %u other item (%s) selected").printf (files_count, fsize),
-                                         _(" and %u other items (%s) selected").printf (files_count, fsize),
-                                         files_count);
+                        str += ngettext (" and %u other item (%s) selected",
+                                         " and %u other items (%s) selected",
+                                         files_count).printf (files_count, fsize);
                     } else {
                         str += _(" selected");
                     }
```
ryonakano, line 216: Honestly I think we may want to avoid this way of constructing a translatable string but would be better to address in another PR.

ryonakano (approved):
Looks good to me.

jeremypw:
Yes, this was just a minimal PR.  There is a possibility OverlayBar will be deprecated in future anyway.

PR #1580, opened 2021-02-15
DialogUtils: Ensure that we are not using Gtk into a different thread

tintou:
The Granite.MessageDialog is using `show_all` which is making the dialog crash
as it is potentially created into a different thread.

### DIFF
--- a/libcore/FileOperations/DialogUtils.vala
+++ b/libcore/FileOperations/DialogUtils.vala
@@ -28,34 +28,32 @@ namespace PF {
                 break;
         }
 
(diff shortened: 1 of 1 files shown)


jeremypw:
Are there any reported issues that this might fix?

tintou:
There aren't but I'm getting lots of crashes because of this on my side

jeremypw (approved):
I did not reproduce any crashes in either master or this PR with the subsection of file operations tried.

Code makes sense so will approve.

PR #1589, opened 2021-02-18
Consistent use of quotes and colon in error messages

jeremypw:
Consistently use single quotes for names in messages - similar to usage in file operations.

Also more consistent form of error messages


src/View/Sidebar/DeviceRow.vala
```diff
@@ -243,7 +243,7 @@ public class Sidebar.DeviceRow : Sidebar.BookmarkRow, SidebarItemInterface {
                         try {
                             mounted = !mount.eject_with_operation.end (res);
                         } catch (GLib.Error error) {
-                            warning ("Error ejecting mount: %s", error.message);
+                            warning ("Error ejecting mount '%s' - %s", mount.get_name (), error.message);
```
ryonakano, line 246: I rarely see this notation that uses `-` to join the error message in other places. Is there any reason for this, e.g. it's grammatical? Maybe unifying the notation to `<error-summary>: <row-error-message>` (for example `Error ejecting mount '%s': %s`) is more common?

jeremypw: This is the form I have used elsewhere in the code.  However, after looking into it, I agree that strict grammar would suggest a colon would be more appropriate.  I would prefer to change all occurrences to be consistent so it will make the PR a bit larger.

ryonakano (approved):

PR #1598, opened 2021-02-24
Avoid unneeded output when attempting sudo.

jeremypw:
Return early when sudo used on commandline.  Only output warning message.

Similar to https://github.com/elementary/code/pull/951.


jeremypw:
Based on wrong branch!

PR #1600, opened 2021-02-25
ConnectServerDialog: Use Granite.ValidatedEntry

danrabbit:
Looks like the only validation we're actually doing is that the entry isn't blank

This uses ValidatedEntry instead of a private widget. We now validate entries when they change, not every time any entry changes, and the logic for only validating shown entries is clearer

### DIFF
--- a/libcore/ConnectServerDialog.vala
+++ b/libcore/ConnectServerDialog.vala
@@ -86,15 +86,15 @@ public class PF.ConnectServerDialog : Granite.Dialog {
     };
 
     private Gtk.InfoBar info_bar;
-    private DetailEntry server_entry;
+    private Granite.ValidatedEntry server_entry;
     private Gtk.SpinButton port_spinbutton;
     private Gtk.Revealer port_revealer;
-    private DetailEntry share_entry;
+    private Gtk.Entry share_entry;
     private Gtk.ComboBox type_combobox;
-    private DetailEntry folder_entry;
-    private DetailEntry domain_entry;
-    private DetailEntry user_entry;
-    private DetailEntry password_entry;
+    private Gtk.Entry folder_entry;
+    private Granite.ValidatedEntry domain_entry;
+    private Granite.ValidatedEntry user_entry;
+    private Granite.ValidatedEntry password_entry;
     private Gtk.CheckButton remember_checkbutton;
     private Gtk.Button connect_button;
     private Gtk.Button continue_button;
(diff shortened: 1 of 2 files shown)


jeremypw (commented):
Code looks good and successful connection can still be made.
A couple of issues in behaviour -
 - The default user name is no entered by default - it appears as a placeholder but still has to be typed in.  In master, it is pre-entered, which is more convenient.
 - When a password is required it must be entered but deleting it again does not render the connect button insensitive. The same applies to the username. In contrast, the default path is used without being entered.

We could probably add some smarter validation but that can be left to later PRs

Nice use of Granite.ValidatedEntry! :smile:

danrabbit:
@jeremypw thanks for catching those mistakes! Ready for re-review :)

jeremypw (changes requested):
This fixes the  previous review issues but I noticed a couple more:
 - It may be good if there were a placeholder in the username and password field to explain that these fields are required to connect.
 - On Ubuntu, the placeholder in the server field is truncated and you cannot see all of it by resizing the window or from a tooltip.  It is OK on elementary tho'. The layout could do with a redesign anyway so this could be left for another PR.
 - Is the regex correct? It seems to require a `.` in the server name. (My server does not have this).

danrabbit:
@jeremypw Reverted the regex since I was trying to be too clever here :) Also made the dialog resizable. I think maybe these placeholder strings could be re-evaluated in another branch to be more concise

Since almost all of the fields are required, I changed the placeholder of the two optional fields instead. Do you think that solves the issue?

libcore/ConnectServerDialog.vala
```diff
@@ -130,8 +127,9 @@ public class PF.ConnectServerDialog : Granite.Dialog {
 
         var server_header_label = new Granite.HeaderLabel (_("Server Details"));
 
-        server_entry = new DetailEntry (_("Server name or IP address")) {
-            hexpand = true
+        server_entry = new Granite.ValidatedEntry () {
+            hexpand = true,
+            placeholder_text = _("Server name or IP address")
         };
```
jeremypw, line 134: You can ensure the placeholder is fully visible by adding 
```
server_entry.width_chars = server_entry.placeholder_text.length;
```

or to ensure long enough to show a full IP address (at least IPv4 - I can't see anyone entering an IPv6 address manually)
```
server_entry.width_chars = int.max (server_entry.placeholder_text.length, 15);
```

danrabbit: I'm not sure if it's a good idea to use the placeholder width chars here since that might end up being a very long string in some more verbose locales and users can now resize the dialog if they need to do that. Setting a min width for an expended IP Address length could work though. I think what might end up being best in the long run is to change this placeholder text to something like: `example.com or 104.21.3.130`

jeremypw (approved):
There is no a large empty space between the fields and the buttons on first opening (because `resizable = false` was removed - is that intended?  Anyway, I think we can bank this now and iterate on design as there are a number of other issues needing addressing in the master version. e.g. I am not sure we should be asking for the password at all here - it is possible that the password has been remembered by the system in which case entering it is unnecessary.  It would therefore be better to attempt to connect without a password and let the system throw a password dialog if required.  This would simplify this dialog.

PR #1603, opened 2021-02-25
ConnectServerDialog: Remove single-use methods

danrabbit:
Remove methods that are only ever called once


jeremypw (commented):
Code good, no regressions found.

PR #1604, opened 2021-02-26
Fix target gettext domain of the policy file

ryonakano:
Because of wrong domain the dialog text is not translatable:

Screenshot from 2021-02-26 12-43-14

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -48,14 +48,12 @@ appdata_in = configure_file(
     install: false,
 )
 
-po_dir = join_paths(meson.source_root(), 'po')
-
 i18n.merge_file(
     'policy',
     input: policy_in,
     output: meson.project_name() + '.policy',
-    po_dir: po_dir,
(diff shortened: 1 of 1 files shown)


data/meson.build
```diff
@@ -48,14 +48,12 @@ appdata_in = configure_file(
...
-po_dir = join_paths(meson.source_root(), 'po')
-
 i18n.merge_file(
     'policy',
     input: policy_in,
     output: meson.project_name() + '.policy',
-    po_dir: po_dir,
-    data_dirs: po_dir,
+    po_dir: join_paths(meson.source_root(), 'po', 'extra'),
+    data_dirs: join_paths(meson.source_root(), 'po', 'extra'),
```
tintou, line 56:
```suggestion
    data_dirs: join_paths(meson.source_root(), 'po'),
```
data_dirs should link to the .its file

tintou (approved):

PR #1607, opened 2021-02-28
Sidebar: Drives as expandable items

### LINKED ISSUE #496: Devices section in sidebar does not show devices but volumes
This is related to #271.

Fixes would include showing volumes as a subtree of each physical device (not simple) or changing the description of this section (simple).

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58576609-devices-section-in-sidebar-does-not-show-devices-but-volumes?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1606: Files showing drive AND volumes
## What Happened

Files is showing the connected drive, it should only show volumes. Caused by #1596 

Before #1596:
Screenshot from 2021-02-27 15-22-26

After:
Screenshot from 2021-02-27 15-18-58

## Expected Behavior

It should only show the volumes.

## Steps to Reproduce

1. Files daily.
2. Connect a drive.

## Platform Information
Files master

### LINKED ISSUE #1667: Connected ssh servers are appearing both under Storage and Network sidebar
I am not sure if this is intentional or a bug. I just want to report it. 

Ekran görüntüsü 2021-04-17 18-00-48

## What Happened
Connected ssh servers are appearing both under Storage and Network sidebar

## Expected Behavior
I think it's better to show just under Network

## Steps to Reproduce
1. Connect to ssh server

## Platform Information
```bash
~ $ head -2 /etc/os-release
NAME="elementary OS"
VERSION="6 Early Access"
```

jeremypw:
Fixes #1606 
Fixes #496
Fixes #1667 

Drives with volumes are shown as expandable items with their volumes as children


jeremypw:
If the UX team are not in favour of this change in design then separate fixes will be required for some of the linked issues.

jeremypw:
ExpandableDriveBookmarks

cassidyjames:
Hm, at a glance I'm not a fan of the nesting, as it seems to be very busy visually even with only one or two connected devices. I would be more in favor of the old behavior of only showing the volumes, but I also see some value in showing the device name itself… I'm just not sure that nesting is the answer. 

Would it make sense to show the device name, then only show the volume name with the device name in parenthesis after it if there are more than one volumes? So you might end up with:

>**Storage**
> :computer: File System
> :floppy_disk: SanDisk Cruzer
> :floppy_disk: boot (Samsung Fit)
> :floppy_disk: data (Samsung Fit)

jeremypw:
@cassidyjames OK, I'll work on an alternative solution. One possible result concatenating the drive and volume names as you suggest could be some pretty long bookmark names which will be ellipsized.  Maybe can add a tooltip to the volume bookmarks showing the additional info (as well).

Leaving as draft in case of further comments.

jeremypw:
Closed in favour of #1752

PR #1612, opened 2021-03-03
ProgressInfoWidget: Use tabular numbers in file transfer dialog

### LINKED ISSUE #1567: Use tabular numbers in file transfer dialog
## Problem

When a file is being transferred, the string showing the progress jitters all over the place due to the differing widths of the numbers.

https://user-images.githubusercontent.com/611168/106035168-04ae7a80-6091-11eb-8eeb-114c455d0554.mp4

## Proposal

We use tabular numbers in the file transfer dialog (or really, anywhere we're showing filesizes, probably). (Idea from [@alatiera@mastodon.social](https://mastodon.social/@alatiera/105624731925706525).)

## Prior Art

GNOME Web seems to use tabular numbers in the download popover to prevent the string from jumping around.

bluesabre:
This branch updates both the status and details labels with tabular numbers. Please let me know if I need to make any other changes. Thanks!

Closes #1567


src/View/Widgets/ProgressInfoWidget.vala
```diff
@@ -83,13 +85,14 @@ public class Marlin.Progress.InfoWidget : Gtk.Grid {
             cancelled (info);
             button.sensitive = false;
         });
-
-        info.bind_property ("status", status, "label");
     }
 
     private void update_data () {
+        status.set_markup (
+            Markup.printf_escaped ("<span font_features='tnum'>%s</span>", info.status)
```
jeremypw, line 92: Shouldn't this be `<span font-features='tnum' ....`?  It doesn't seem to work otherwise.

src/View/Widgets/ProgressInfoWidget.vala
```diff
@@ -83,13 +85,14 @@ public class Marlin.Progress.InfoWidget : Gtk.Grid {
...
+        );
         details.set_markup (
-            Markup.printf_escaped ("<span size='small'>%s</span>", info.details)
+            Markup.printf_escaped ("<span size='small' font_features='tnum'>%s</span>", info.details)
```
jeremypw, line 95: See comment above

jeremypw (changes requested):
Thanks for your first contribution to Files! :grinning: 

I think the attribute should be "font-features", not "font_features"

bluesabre:
Thanks for the review! From my testing, it seems like either dash or underscore was working, and I based it on https://github.com/elementary/wingpanel-indicator-datetime/pull/246/files

Since they both work, I went ahead and swapped the code over to dashes instead. Please let me know if I need to adjust anything else.

jeremypw:
Strange - when I tested it there still seemed to be some movement unless I changed the underscore to hyphen.  I think hyphens are more usual for quoted properties, attributes & signal names and in some other circumstances (e.g. property bindings) mandatory.  Anyway, thanks for changing it - I will approve now.

jeremypw (approved):

cassidyjames (approved):

PR #1627, opened 2021-03-12
Connect to list "row-activated" signal.

### LINKED ISSUE #1626: Entire item in sidebar not clickable
## What Happened

I thought Files required me to click sidebar items several times to open things, but I found out the root cause: the entire item in the sidebar is not clickable.

Screenshot from 2021-03-11 16-43-19

Instead, only the middle area of each sidebar item is clickable, even though the entire item responds to :hover events.

## Expected Behavior

The entire item is clickable, edge-to-edge and top-to-bottom.

## Steps to Reproduce

1. Open Files
2. Hover over the edge of a sidebar item, noting that it's highlighted
3. Click! and nothing happens

## Platform Information

elementary OS 6 early access

jeremypw:
Fixes #1626

It seems (primary) clicks on the edge of a ListBoxRow generate a signal only in the parent ListBox so we have to connect to that as well.  Note this PR only fixes primary clicks.  Middle and secondary clicks are still dealt with by the ListBoxRow so do not work on the edge.  Not sure if there is a solution to this unless the size of the border can be changed using CSS?

### DIFF
--- a/src/View/Sidebar/BookmarkListBox.vala
+++ b/src/View/Sidebar/BookmarkListBox.vala
@@ -41,6 +41,11 @@ public class Sidebar.BookmarkListBox : Gtk.ListBox, Sidebar.SidebarListInterface
         bookmark_list.loaded.connect (() => {
             refresh ();
         });
+
+        //Deal with unintercepted primary click anywhere on row
+        row_activated.connect ((row) => {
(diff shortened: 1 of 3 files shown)


jeremypw:
Note: Even setting `button_press.connect ()` on the listbox row does not intercept the button press - it seems to go to the listbox first and gets swallowed.

PR #1628, opened 2021-03-12
Sidebar: Use the `row-activated` signal in the listbox

tintou:
This is the intended signal to track listbox activations.


jeremypw (approved):
LGTM and works as expected

PR #1634, opened 2021-03-20
Fix mtp mount

### LINKED ISSUE #1625: MTP devices don't open from sidebar
## What Happened

I connected my Pixel 5 Android phone to my computer, enabled data transfer on the phone side, and then clicked the item in the sidebar.

## Expected Behavior

The Pixel 5's internal storage should open via `mtp://`

## Steps to Reproduce

1. Connect an Android phone (or other mtp device) and enable file transfer on the device
2. Open files and click on the device in the sidebar

As a note, right-clicking and opening properties works, and the `mtp://` URI in the properties dialog _does_ actually work and open the folder.

## Platform Information

elementary OS 6 Odin daily build

jeremypw:
Fixes #1625 

Also attempts to fix some problems around mounting devices like mobile phones that can switch protocols between MPT and PTP, do not provide a UUID when queried with `get_uuid ()` and give the mount a different name to the parent volume.   In these circumstances it is hard to match up a `mount_added` signal mount with a pre-existing volume row and duplicate rows - one for the volume and one for the mount can result. 

The simplest solution is just to clear and reload the device listbox but this results in a visual glitch and possible row re-ordering when an unmounted row is clicked to mount. 

The solution used is not elegant but works. Suggestions for a better one welcome.

The handling of the `mount_added` is delayed to ensure that, if it is the result of activating a volume row, then that row has updated its uuid to the mount location if the volume did not already have a uuid.  Then no duplicate row is created.


leonardo-lemos (approved):
Code looks good and the issue in context was fixed. Approved.

PR #1645, opened 2021-04-01
Avoid unexpected delay on first startup

### LINKED ISSUE #1642: Slower to start up with new sidebar
Ever since the new sidebar was merged, I've been noticing Files takes a few seconds to start up across my machines, even on my fast SSDs and relatively high-specced laptops. Interestingly, it seems that subsequent startups are faster, like something has to be initially cached after a cold boot.

Before the new sidebar, Files would open nearly instantaneously.

jeremypw:
Fixes #1642

Do not try to get fileinfo for network:// scheme


davidmhewitt:
Am I right in thinking that the `network://` scheme is just some virtual thing that will never have any real files under it?

I guess if you're on an actual network share with files that you do want to get the info of, then the prefix will be `smb://` or `nfs://` or whatever?

I think a comment explaining that may be useful, because it kind of reads like you're disallowing getting the `FileInfo` for any network files at a glance.

jeremypw:
I just need to check this does not break anything for a selection of network servers. Unfortunately my test server machine needs rebuilding so there will be a slight delay ...

danrabbit (approved):
:rocket: we can always find out if people start reporting errors in beta

PR #1653, opened 2021-04-10
Inline emblems

### LINKED ISSUE #1647: Place emblems in line with labels in list views
bitmap

## Problem

In list views, especially if we go down smaller than the default size, things get really cluttered as we add emblems over the icons

## Proposal

Instead of overlaying emblems on the file icon, it seems like we should make use of the horizontal space in the list view to render emblems after item labels

### LINKED ISSUE #1650: Icons have too much hpadding in smaller list views
## What Happened

I initially thought it was a bug, but it seems that it is caused by the 'more-compact-listview' that was merged.
The listview items should not have so much horizontal padding on their icons, at the smallest mode the padding is wider than even the icon itself:
Screenshot from 2021-04-05 15-15-01

## Expected Behavior

If the icon does not have enough space to have badges without obscuring it then it should use the same style as in other places like settings:
Screenshot from 2021-04-05 15-12-55

## Platform Information

elementary OS 6 Daily

jeremypw:
Fixes #1647 
Fixes #1650 

Needs some design input on some points:

* The emblem cell is on the right of the text cell (which includes all spare space) and the emblems are right justified within the cell.  Is this right or should the emblems be left justifies within the emblem cell.  Note that the emblem cell renderer can only have one width (not vary from row to row) which will be governed by the file with the largest number of emblems in the folder.
* The emblems are not currently styled and appear rather prominent especially in folders where every file has an emblem.  Should the emblems be toned down somehow?
* The space to the right of the icon that was reserved for emblems has been removed so the padding between the icon and the text is now less. OK?
* Extra space has been kept on the right in ListView to make it easier to click on the expander (in ColumnView the icon is only padded enough for the helper. OK?
* Maybe needs some padding to the right of the emblems?


jeremypw:
Screenshot from 2021-04-10 11 57 32
Screenshot from 2021-04-10 11 57 18

hanaral:
Maybe the link emblem should be the same as the one on the cursor when you copy as a link? that way it could be easily badge-ified like the other emblems.

jeremypw:
Yes, it needs some work on the appearance - I just got something working.

hanaral:
I should probably mention that these large coloured flat badges can look a bit like buttons to less 'technically adept' people.

src/EmblemRenderer.vala
```diff
@@ -0,0 +1,108 @@
...
+    License along with this program; see the file COPYING.  If not,
+    write to the Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+    Boston, MA 02110-1335 USA.
+
+    Author(s):  Jeremy Wootten <jeremy@elementaryos.org>
+
+***/
+
+namespace Marlin {
+    public class EmblemRenderer : Gtk.CellRenderer {
```
danrabbit, line 24: For new files, can we use inline namespacing:
```suggestion
public class Marlin.EmblemRenderer : Gtk.CellRenderer {
```

src/EmblemRenderer.vala
```diff
@@ -0,0 +1,108 @@
...
+        private GOF.File? _file;
+        public GOF.File? file {
+            get {
+                return _file;
+            }
+            set {
+                _file = value;
+            }
+        }
```
danrabbit, line 33: Is there any difference between this and:
```suggestion
        public GOF.File? file { get; set; }
```

danrabbit (changes requested):
Yeah I think this will help to have things a bit less cluttered. I wonder if it is maybe too far away though, in the list view especially. Maybe we should try to have it render right after the label and see how that looks.

If we leave the badges right aligned, I definitely feel like we need some margin so they're not rendered right up against the edge of the cell. Currently they are rendered underneath the scroll bar in the column view since there is no gutter at all.

There aren't currently `-symbolic` icons in the icon set, but I wonder if that would look nicer to use `-symbolic` here, especially to avoid the button appearance of the read-only emblem

thoughts?

hanaral:
Firstly, yeah the current files (as in right now) has pretty poor padding in column and list view, p sure I opened an issue for this the other day.
Also I thought of this yesterday, how about having 'wireframe' icons like the `i` in appcenter, which could tell the user what they mean when hovered over (also) in the same way as appcenter

jeremypw:
Left aligned and 50% translucent emblems (with the emblem renderer being the only expanding renderer in the name column so getting all the spare space).

Screenshot from 2021-04-15 18 39 26

jeremypw:
It is not possible to render the emblems right against the text of each name (at least not somehow renderering them in Pango in the TextRenderer.  For left justification to work the `expand` property of the emblem renderer has to be set `true` (otherwise the name renderer takes it all and the emblems appear right justified).  If the same property of the name column is set `false` then the emblem renderer takes all the spare space and long names get truncated unnecessarily.  If the are both set `true` then the spare space is shared and the emblems can appear some distance both from the name and the right margin.

Screenshot from 2021-04-15 19 02 16

jeremypw:
I tried adding `-symbolic` to the emblem names but it makes no difference at the moment.  I used the built Utility function to make the emblems translucent - not sure if there is a way to do it with CSS as they are not widgets just pixbufs.

src/EmblemRenderer.vala
jeremypw: Good point

jeremypw:
With margin to the right of the emblems
Screenshot from 2021-04-15 19 43 16

danrabbit:
@jeremypw I just proposed a branch with symbolic variants here: https://github.com/elementary/icons/pull/1022

Can we reduce that gutter size to 12px? That's our usual margin around dialogs and other containers

danrabbit:
Here's a screenshot with the proposed -symbolic icons. Looks like we need to vertically center as well here:

Screenshot from 2021-04-15 13 08 55

hanaral:
@danrabbit UX question here, would it seriously not be better to just add the link badge and pack all the other badge types in to one of these bad boys: 
infohover
That way you could add some context too like:
`you do not permission to view, read or write to this folder`

jeremypw:
@danrabbit  OK, I will fix the margin and alignment.  What do you think about the left-alignment and translucency (not committed yet)?  I don't like the left alignment tbh because of the free space problems.  You may want to provide custom badges instead of trying to modify the existing ones - I guess there is no hurry to merge this.

jeremypw:
@danrabbit :  Should the symbolic icons be only used for the inline emblems or also in the IconView?

danrabbit:
@hanaral Yes it would be better if we could pack widgets here, but that's not how CellRenderers in Gtk.TreeView work, so we have to use some custom drawing here. In a Gtk4 port of Files we can revisit this and use nicely tooltipped widgets.

@jeremypw yeah I agree that the left alignment doesn't look right. I'm not a big fan of the translucency. The symbolic icons just were merged and should be much less weighty.

jeremypw:
OK, will use the new symbolic emblems only in List and Column Views.

danrabbit (changes requested):
Looks like we aren't updating when switching to dark mode, but other than that, this looks really good!

Screenshot from 2021-04-16 10 19 58

jeremypw:
@danrabbit Is switching to dark mode something that needs to be done in Files code?

danrabbit:
@jeremypw could it be the same issue as in https://github.com/elementary/files/pull/1466 ?

jeremypw:
Maybe a problem with the Files.IconInfo class (which does cache icons but should regenerate when the theme changes) - the pixbuf being rendered comes from here so at that point it is already wrong.

jeremypw:
May need to reduce size of emblems from 16px at small icon sizes?  This would be easy enough.

danrabbit (approved):
Much less cluttered! Far better use of spice. Looks good!

hanaral:
I too appreciate good spice in my food, it adds a certain edge to dishes imo

PR #1654, opened 2021-04-10
Directory.Async: Use vala syntax sugar for variable lock

tintou:
Also add some code cleanup and avoid increasing reference count in loops.


jeremypw (approved):
Looks good - thanks.  I think we can change the class to Files.DirectoryAsync can't we?  But that can be done in a separate PR as it will affect multiple files.  We can also change the variable name `gof` as it has no meaning after losing the `GOF` namespace.

jeremypw:
Oops, pressed the wrong button ...

PR #1659, opened 2021-04-14
Reduce the use of the explicit namespace

tintou:
Only keep the explicit namespace for disambiguity cases.

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -25,6 +25,8 @@
      * ColumnView
 */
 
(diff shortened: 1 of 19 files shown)


src/View/AbstractDirectoryView.vala
```diff
@@ -25,6 +25,8 @@
      * ColumnView
 */
 
+using Marlin;
+
 namespace FM {
```
danrabbit, line 30: I'm guessing this is temporary because of the `FM`?

danrabbit (approved):

PR #1670, opened 2021-04-21
ListModel: simple port to Vala

tintou:
This isn't a 1:1 port, I took the opportunity to subclass Gtk.TreeStore to keep it simple


jeremypw (approved):
I used this for a while without problem so I think its time to merge and give it more exposure.

Issue #1671, opened 2021-04-22
Replace Zeitgeist with Tracker


danrabbit:
## Problem

It looks like we're currently relying on Zeitgeist for some features in Search and Recent documents. Zeitgeist is old and unmaintained and it looks like it isn't always working properly judging by #1183

## Proposal

We should instead use Tracker for providing these features

PR #1674, opened 2021-04-23
Remove Zeitgeist dependency

jeremypw:
Part one of addressing #1671 

Tbh I never use any of the facilities provided by Zeitgeist in Files anyway.  Need input as to which of these is worth re-implementing with Tracker or otherwise.

For me, getting Recent View to work properly would be the main priority.


hanaral:
While I don't use it too much, the recents view is probably used by quite a few people - I recall using it a lot when I was younger and a little less technically adept.
So if it's possible it might be better to swap it for tracker at one fell swoop.

jeremypw:
As far as I can make out, Tracker is, by default, only for indexing the content of media files in the user home directory and hence providing fast search results.  So it can be used instead of Zeitgeist in the searchbar but it doesn't seem to help the Recent Files view (which is provided by gvfs) which is based on usage not content.

We may have to provide our own enhancement of the Recent view to show recently visited folders as well as recently accessed files.

jeremypw:
Closing pending clarification of alternatives.

PR #1676, opened 2021-04-24
Reduce left gutter in column views

### LINKED ISSUE #1578: Better column view spacing
## Problem

Columns take up far too much space and folder and file names and icons are poorly padded, which is a self reproducing problem.

The smallest icon view gives list icons a very generous amount of padding on the left and an absurd amount of padding on the right:
Screenshot from 2021-02-11 12-01-33
I understand that the  padding on the left is for the selection badge, but I personally don't think the badge should overlap the icons when they are so small anyway.

The columns themselves are also far too large, and I can only fit about 2.4 columns in a relatively normal sized window:
Screenshot from 2021-02-11 12-05-22

## Proposal

Item text and icons should have better padding so that columns can be tighter and easier to fit on screen. Possibly the selection badge should be moved to the far end of the list box

## Prior Art

image

jeremypw:
Fixes #1578 

Removes left padding and overlaps helper with icon by half the helper size thereby moving the icon closer to the left margin.

Note that at large icon sizes the helper may not overlap the visible part of the icon because the icon may have a transparent border (e.g. folder icons).

### DIFF
--- a/src/IconRenderer.vala
+++ b/src/IconRenderer.vala
@@ -44,7 +44,7 @@ namespace Files {
                 icon_size = value.to_icon_size ();
                 helper_size = (int) (_zoom_level <= ZoomLevel.NORMAL ?
                                      Files.IconSize.EMBLEM : Files.IconSize.LARGE_EMBLEM);
-                h_overlap = helper_size / 2;
+                h_overlap = int.max (12, helper_size / 2);
                 v_overlap = h_overlap;
             }
         }
--- a/src/IconRenderer.vala
+++ b/src/IconRenderer.vala
@@ -41,8 +41,10 @@ namespace Files {
             set {
                 _zoom_level = value;
                 icon_size = value.to_icon_size ();
-                h_overlap = int.min (icon_size / 8, Files.IconSize.EMBLEM / 2);
-                v_overlap = int.min (icon_size / 8, Files.IconSize.EMBLEM);
+                helper_size = (int) (_zoom_level <= ZoomLevel.NORMAL ?
+                                     Files.IconSize.EMBLEM : Files.IconSize.LARGE_EMBLEM);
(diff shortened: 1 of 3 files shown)


danrabbit:
One major problem I'm experiencing here is that the helper is partially into the resize grab area for the pane handles. The icon positioning here is much much better, but I wonder if we should move the helper location in another 6px to make sure there's not a conflict between the helper click area and the pane handle grab area

jeremypw:
@danrabbit  I assume you mean resizing the individual columns in Column View? There does not seem to be a problem in the other views.

I'll try adding a bit more left padding in Column View.

danrabbit:
@jeremypw There's also the same issue in the list view with the disclosure triangles

jeremypw:
List View (3px start margin)
Screenshot from 2021-04-27 18 25 10

Column VIew (6px start margin)
Screenshot from 2021-04-27 18 25 53

danrabbit:
I don't think there needs to be a different start margin between list and column views, I would just make the selection helper overlap more. I have an overlap of 12px here so we can have a smaller gutter and not worry about that pane grab area:

Screenshot from 2021-04-27 14 44 37

jeremypw:
If you have an overlap of 12px on a 16px icon it will almost complete obscured - does that matter?

danrabbit:
@jeremypw since its a non-default icon size and the label still acts as a click area I think it probably doesn't matter

jeremypw:
ColumnView - AFTER (6px gutter)
PR-ColumnView-6px

ColumnView - BEFORE
Screenshot from 2021-04-28 10 05 15

ListView - AFTER (6px gutter)
PR-ListView-6px

ListView - BEFORE
Master - ListView

danrabbit (approved):

PR #1716, opened 2021-05-17
Fix reclicking a folder in ColumnView

### LINKED ISSUE #1703: Reclicking on the opened folder cut the column in miller view
## What Happened

Reclicking on the same opened folder cut the open column.
My english is not very good so here is a screenshot : 
elementary-miller-view-cut
  Please note that the scrollbar is at the end and we cannot scroll anymore.

## Expected Behavior

Seeing the full column

## Steps to Reproduce

1. Open folders until the scrollbar appear.
2. Reclick on the opened folder.

### Logs
 No log.

## Platform Information

Elementary OS 6 Beta on gnome-boxes. Tested with pantheon-files from the release and the master.

Thanks !

jeremypw:
Fixes #1703 

It was not easy to find a simple fix with the existing rather tangled code - a certain amount of refactoring simplification was carried out.


tintou (approved):
This looks easier to read, I'm concerned by Files.Animation using a static int for `timeout_source_id` as that means that any widget using `smooth_adjustment_to` will share the same animation context.

jeremypw:
@tintou Thanks for review and pointing out that potential issue.  I guess I was only envisaging one widget being animated at any time but maybe it would be safer to make the timout source id non-static.

PR #1721, opened 2021-05-25
AbstractDirectoryView: Add destructive action class to delete menuitem

danrabbit:
minor code style while we're here


jeremypw (approved):
Makes sense and works as expected.

PR #1724, opened 2021-06-03
Stretch "Connect Server..." button

fs3o5:


cassidyjames:
I don't think we want to expand it because it's intentionally at the start of the sidebar; see the design of the Online Accounts settings in System Settings for another instance of this design pattern.

danrabbit:
I'm not super opposed to having the button fill the available space, but this isn't really the right way to do that programatically.

Instead, you can set `hexpand` and `xalign` on the button itself without getting its contents

cassidyjames (commented):
Yeah this works. :+1:

cassidyjames (approved):

PR #1732, opened 2021-06-10
Fix removing file from model after renaming or trashing/deleting

### LINKED ISSUE #1730: Rename is broken
## What Happened

Select a file, rename it and you will see that the file with the previous name is still here. 

## Expected Behavior

The file with the previous name is deleted.

## Steps to Reproduce

1. Select a file.
2. Rename it.

### Logs

No logs.

## Platform Information

Elementary OS 6 Beta on gnome-boxes. Tested with pantheon-files from the master.

Thanks !

jeremypw:
Fixes #1730 

Also fixes moving to trash/deleting file.

Fixes removing file from model when it is at the root level.


tintou (approved):
oops

PR #1741, opened 2021-06-21
Fix empty trash context menu item when already empty

### LINKED ISSUE #1738: You can click "Permanently Delete All Trash" infinitely
## What Happened

You can click `Permanently Delete All Trash` infinitely.

## Expected Behavior

If there is no file on the Trash, it must be grayed out and not clickable.

## Steps to Reproduce

https://user-images.githubusercontent.com/87835/122033006-d085ed00-cdd8-11eb-81ea-9b67d434c980.mp4

## Platform Information

```bash
~ $ dpkg-query --show pantheon-files
pantheon-files	4.5.0+r4513+pkg100~daily~ubuntu6.1
~ $ head -2 /etc/os-release
NAME="elementary OS"
VERSION="6 Early Access"
```

jeremypw:
Fixes #1738 

Rather than patching up the existing code, this PR makes the sidebar Trash Item a plugin item under the control of the Trash plugin which seemed cleaner than the special case code under Sidebar.  However, a drawback of this is that it is not possible to add custom styling to the context menuitems produced so the "Permanently Delete all Trash" item no longer has destructive action styling. 

It could be argued that allowing permanent deletion of trash items "sight unseen" from the sidebar is not a good idea and that that menu option should be removed altogether so that the used must open Trash to see its contents before emptying. That would be simpler.

### DIFF
--- a/libcore/FileOperations/EmptyTrashJob.vala
+++ b/libcore/FileOperations/EmptyTrashJob.vala
@@ -107,9 +107,9 @@ public class Files.FileOperations.EmptyTrashJob : CommonJob {
         }
     }
 
-    public async void empty_trash () {
+    public async void empty_trash (bool confirm = Files.Preferences.get_default ().confirm_trash) {
         inhibit_power_manager (_("Emptying Trash"));
-        if (!Files.Preferences.get_default ().confirm_trash || confirm_empty_trash ()) {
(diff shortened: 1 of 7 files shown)


jeremypw:
Closing in favour of simpler solution.

PR #1750, opened 2021-07-10
Add space before hyphen in contents of file overlay

### LINKED ISSUE #1749: Missing space before hyphen in hover tooltip/toast
## What Happened
When I hover a file, a tooltip/toast appears in the bottom right, showing the filename, filetype and size. Between the filename and the filetype there's a hyphen, but it only has a space to the right. The left-side is glued to the filename. Screenshot for demonstration:

Screenshot from 2021-07-10 15-01-10_

Note: this doesn't happen for folders:

Screenshot from 2021-07-10 15-10-47

## Expected Behavior
I'd expect there to be a space before the hyphen as well.

## Steps to Reproduce

1. Open a Finder window
2. Hover over any regular file, e.g. a `.txt` or executable.
3. Observe the contents of the tooltip/toast message that appears at the bottom right of the window.

## Platform Information
* elementary OS version: elementary OS 5.1.7 Hera

waldyrious:
Fixes #1749.﻿

Note that commit bfcf95a is only about improving the code comments of this file. The actual fix is isolated in commit 97c368f.

### DIFF
--- a/src/View/Widgets/OverlayBar.vala
+++ b/src/View/Widgets/OverlayBar.vala
@@ -37,7 +37,7 @@ namespace Files.View {
         private uint deep_count_timeout_id = 0;
 
         public OverlayBar (Gtk.Overlay overlay) {
-            base (overlay); /* this adds the overlaybar to the overlay (ViewContainer) */
+            base (overlay); /* This adds the overlaybar to the overlay (ViewContainer). */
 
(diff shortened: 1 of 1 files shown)


jeremypw (approved):
Nice catch - thanks.  Thanks for the comment cleanup as well.

Issue #1665, opened 2021-04-17
exFAT formated USBDisks appearing double on the sidebar.


queeup:
I couldn't check all formats but exFAT USBDisks are acting strange. Mounted fine but appearing double on the sidebar and when I try to unmount, one of them disappeared but left one of it visible until I restart files app.

## Steps to Reproduce

1. Put exFAT formated USBDisk and mount it on the files app.
2. Try to unmount

## Platform Information
```bash
~ $ head -2 /etc/os-release
NAME="elementary OS"
VERSION="6 Early Access"
~ $ dpkg-query --showformat='${Package}:\t${Version}\n' --show pantheon-files
pantheon-files:	4.5.0+r4445+pkg100~daily~ubuntu6.1
```

jeremypw:
Probably a regression from recent Sidebar rewrite.

jeremypw:
I created an exfat partition on a USB stick using `mkexfatfs` on the commandline.  Files does not automatically detect this stick when it is plugged in (nor did Nautilus).  If I fuse-mount this partition to a directory in my home folder using `sudo mount.exfat <device> <dir>` on the commandline the USB stick appears under "Devices" (once) and can be written to successfully.  The contents can be viewed either by clicking the device row or by navigating to the folder to which it was mounted. In either case, the pathbar shows the path to the folder to which it was mounted. It could be unmounted with Files after entering the admin password. This was using Files compiled from latest master.

>Mounted fine but appearing double on the sidebar

Please clarify where the double entries appear and which disappeared on unmounting.  Please detail any differences in the way you made the exfat partition and how you mounted/dismounted it.

jeremypw:
I checked devices with FAT16 and ISO9660 partitions and they mounted/unmounted as expected.

queeup:
I used it `gnome-disks`  to format exfat usbdisk. Debug log: https://pastebin.ubuntu.com/p/hmcnMY6JPV/

```bash
~ $ dpkg-query --showformat='${Package}:\t${Version}\n' --show pantheon-files
pantheon-files:	4.5.0+r4453+pkg100~daily~ubuntu6.1

```


https://user-images.githubusercontent.com/87835/116143827-4bd5f700-a6e4-11eb-9d55-1bdffa1f35ad.mp4

selwynorren:
On Elementary OS 6 Stable I am still having this issue. However it only shows duplicated on first mount. If I close files and open it again it then shows a single mounted USB drive

[...]

PR #1752, opened 2021-07-12
Only volume bookmarks where possible

### LINKED ISSUE #496: Devices section in sidebar does not show devices but volumes

### LINKED ISSUE #1606: Files showing drive AND volumes

### LINKED ISSUE #1667: Connected ssh servers are appearing both under Storage and Network sidebar

### LINKED ISSUE #1784: Files does not copy correctly
## What Happened
Files were copied every time with corruption.

## Expected Behavior

I was creating a bootable Ventoy USB drive (program that enables to boot from multiple ISOs without formatting) and each time I copied elementary ISO into the flash drive it refuses to boot and gives me this [booting error ](https://askubuntu.com/questions/5996/unable-to-mount-dev-loop0-during-install). one suggested to perform a checksum on the ISO and it turns out that it was corrupted and i had to open windows and copy elementary ISO and it worked this time!

## Steps to Reproduce
1. [Download ventoy-1.0.49-linux.tar.gz](https://www.ventoy.net/en/download.html) 
2. Run Ventoy web service, [(Useful tutorial for installing Ventoy in linux](https://linoxide.com/how-to-install-ventoy-on-ubuntu/)
3. Now mount the Ventoy USB stick and simply compy elementary ISO and try to boot from it!

### Logs
(in the boot error link)

##Extra Information
1. elementary.files seemed to copy files very fast compared to windows, specifically with 57 MB/s compared to 2.6 MB/s in Windows 10
2. After each time i tried to copy elementary ISO, a service called plymouth prevented the system from poweroff and it remained till eternity until i physically remove the USB from the laptop even if i unmounted it before rebooting, the issue remained the same.
3. this issue did not appear when i copied Windows.ISO as the windows worked with no issues at all!

## Platform Information
OS: elementary OS 6 Odin x86_64 (20210810 built)
Hardware: Lenovo Legion Y530

jeremypw:
Fixes #1606
Fixes #496
Fixes #1667
Fixes #1784 

Storage bookmarks are divided into three sub-classes classes to avoid complex logic in the parent class DeviceRow, which is now merged into AbstractMountableRow.
* VolumeRow - used for most storage entries which exist on mountable volumes (both unmounted and mounted).  The associated drive (if any) is included in the custom name.
* VolumelessMountRow Used for e.g. bind mounts and for the local FileSystem entry (which is a special case where the mount is null). Note certain bind mounts created in `fstab` appear unmountable to GLib so an eject button appears but in fact the operation fails. Not sure if this can be fixed.
* EmptyDriveRow - used for e.g. unformatted USB sticks and empty optical drives.  This row will hide if a volume appears that is linked to the drive (e.g. if media is inserted) so as to avoid duplicate entries. It will reappear if all the associated volumes are removed (e.g. if media is ejected). It is possible to eject/stop an empty drive in order to e.g. open the tray of a DVD drive. A flaw is that there seems no way of detecting that a drive tray is open (`can_eject ()` still returns `true` ) so the eject button does not disappear while the tray is open.
* NetworkRow - derived from VolumelessMountRow - used for remote mountable locations (and "network:///"). If the protocol supports it (e.g. smb) then a storage bar is shown.

Infrastructure for sorting the storage (and network) rows is implemented with a candidate sort order that keeps volumes on the same drive together.

(May also f*x #1665)
Also seems to f*x #1781 although that was not the main aim so any problems in that area should not block this PR.


jeremypw:
2 unmounted local partitions, external disk with 3 partitions (one mounted), MP3 player plugged into USB
Storage1

Tooltip for unmounted volume
StorageTooltipUnmountedVolume

Tooltip for Mounted Volume on external disk
StorageTooltipMountedVolume

Tooltip for mounted MP3 player (note: no associated drive with this volume)
StorageTooltipMP3mounted

Tooltip for DVD drive without media
StorageTooltipEmptyDrive

DVD drive with media (unmounted)
StorageDVDwithMediaUnmounted

DVD drive with mounted media
StorageDVDwithMediaMounted

jeremypw:
SSD, Internal harddrive, Bind mount, SD card, Android phone mounted for photo transfer, 
StorageCameraFAt16
Bind mount tooltip
StorageTooltipBindMount

Android/PTP (not mounted) tooltip
CameraTooltipNotMounted

Android Photo transfer by PTP (mounted) tooltip
StorageCameraTooltip

SD Card (mounted) tooltip
StorageFAT16Tooltip

Android file transfer by MTP (mounted) tooltip
StorageAndroidMTPTooltip

jeremypw:
After connection to remote SSH server (only appears under network)
SidebarSSHConnection

jeremypw:
After connection to remote SMB server - shows storagebar
NetworkSMBMount

jeremypw:
@cassidyjames Some UX input would be appreciated before getting a code review :smile:

cassidyjames:
@jeremypw I haven't dived super into the details, but I think I like where this is going. It seems to solve some issues and be clear about the devices/volumes mounted in the sidebar.

jeremypw:
@cassidyjames Great! I'll request a code review in that case.

hanaral:
Would it be possible to get this merged? I can't unplug my USB HDD without potentially damaging it/loosing data on Odin unless I install gnome-disks from the terminal :(

marbetschar:
@jeremypw not looked at the code yet, but did some functional testing. Congrats so far, this is a huge improvement!

For the system drive I noticed the progress bar is missing now. In the current version of Files it looks like this for me (notice the progress bar below "File System"). Personally I don't think I would miss the bar much - but it definitely caught my eye first thing I opened Files; so this will be noticed by others for sure :)

Screenshot from 2021-09-05 11-23-29

Second thing I noticed is for my USB stick: When I plug it in, its mounted without a hitch. However, on eject it is added to the "Devices" section now. Previously it was completely removed from the sidebar. The new behavior is fine by me - but if it is still shown in the sidebar, I'd expect to be able to re-mount it. But when I click on it, I only get the following error message - and the only way to re-mount it is to physically unplug and plug it back in. Unless I miss something, it does not feel useful to display it under devices at all:

Screenshot from 2021-09-05 11-27-39

jeremypw:
@marbetschar Many thanks for the review!  I had recently noticed the USB issue but haven't found a fix yet.  It is due to the need to handle (builtin) optical drives where unmounting physically ejects the media, but the drive is still present so it did not feel right to remove it from the sidebar.  At least some USB sticks behave the same way - when unmounted it is as if the media is removed so they cannot be remounted.

I will work some more on it. 

Do you think it would be OK to not show optical drives and similar with no media? In that case unmounted/ejected USB sticks and empty optical drives would no be shown.  I'll also investigate showing separate unmount and eject options - the button only unmounting but keeping the media available for remounting.

Not sure why "File System" is not showing a space used bar - I'll look into it.

marbetschar:
> I had recently noticed the USB issue but haven't found a fix yet. It is due to the need to handle (builtin) optical drives where unmounting physically ejects the media, but the drive is still present so it did not feel right to remove it from the sidebar. At least some USB sticks behave the same way - when unmounted it is as if the media is removed so they cannot be remounted.

Currently I do have a bootable iso flashed on this USB stick (elementary OS 6 to be exact), maybe that's the reason why it behaves similar to an optical drive?

> Do you think it would be OK to not show optical drives and similar with no media? In that case unmounted/ejected USB sticks and empty optical drives would no be shown.

Personally I would even expect Files to behave exactly like this. I don't think displaying a drive w/o media does make sense from my perspective.

> I'll also investigate showing separate unmount and eject options - the button only unmounting but keeping the media available for remounting.

I'd be totally fine if I had to physically unplug the USB stick after unmounting and plug it in again to be able to access its media. As long as it also disappears from the sidebar in Files. This would then also be consistent with optical drives, right?

> Not sure why "File System" is not showing a space used bar - I'll look into it.

Thank you so much for working on this!! :+1:

jeremypw:
One reason why you might want Files to show drives without media would be if you want to implement formatting unformatted drives from Files.  Unformatted drives look the same as empty (media ejected) drives.  According to the YAGNI principle maybe shouldn't too much about that though.

jeremypw:
@marbetschar I have tried to address your comments.  The "File System" bookmark now shows the storage bar.  I am still showing empty or unformatted drives but additional information is given and there is no unusable eject button shown. Nor is any unusable context menu shown.

Also if you press the "unmount/eject" button it first tries to unmount the volume.  Unmounted volumes are still shown and can be unmounted.  However, if the volume is ejectable, then the button remains visible even for unmounted volumes and pressing it ejects the volume (after which it cannot be remounted without reinserting media or unplug/replugging the drive).  Only once all the volumes on a drive have been ejected does the drive bookmark show.

It would not be a big deal for empty/unformatted drives to be hidden if that is what the design team prefer - and it would be easily reversible if a formatting function were to be added.

jeremypw:
Empty Drive Bookmarks now look like this (for a USB drive).
EmptyDriveBookmark

For a builtin optical drive the message about unplugging will not show (only shows if `drive.is_removable ()` returns `true`.

marbetschar:
@jeremypw thanks for the fix! The storage bar is there again and the warning dialogs message is much clearer now. However, I'm still getting an additional eject icon for my USB stick - but from the description below the drive, I'd say ejecting the second time seems not necessary, right?

I made a short video which shows the behavior:

https://user-images.githubusercontent.com/392542/132140396-1c69027c-5de8-4480-ab02-a073eff729f1.mp4

marbetschar:
PS: Can we also show the message in the tooltip? Since its quite long its potentially hard to read. Something like this maybe?

```
Samsung Type-C: Media ejected. This device can be unplugged safely.
```

jeremypw:
@marbetschar  I have made sure (I hope) that the eject button is hidden when there is no media that can be ejected.  The only drawback of this is that empty but closed optical drives cannot be opened through the app interface. As users would generally do this manually with a physical button on the drive I don't think this matters.

I have just copied the whole bookmark label to the tooltip for now.  If the UX team want a different layout or text I will change it.

marbetschar:
@jeremypw works flawlessly - thank you so much!!

src/View/Sidebar/AbstractMountableRow.vala
```diff
@@ -140,295 +127,195 @@ public abstract class Sidebar.AbstractMountableRow : Sidebar.BookmarkRow, Sideba
...
+        working = true;
+        try {
+            yield Files.FileOperations.eject_stop_drive (drive, Files.get_active_window ());
+            return true;
+        } catch (Error e) {
+            // MountUtil has already shown error dialog.
+            return false;
+        } finally {
+            working = false;
+            update_visibilities ();
```
marbetschar, line 232: Nicely solved resetting the `working = false;` here :+1:

src/View/Sidebar/BookmarkRow.vala
```diff
@@ -243,7 +243,7 @@ public class Sidebar.BookmarkRow : Gtk.ListBoxRow, SidebarItemInterface {
         }
     }
 
-    protected void popup_context_menu (Gdk.EventButton event) {
+    protected virtual void popup_context_menu (Gdk.EventButton event) {
```
marbetschar, line 246: Why do we need this to be `virtual` now?

src/View/Sidebar/DeviceListBox.vala
```diff
@@ -58,38 +48,71 @@ public class Sidebar.DeviceListBox : Gtk.ListBox, Sidebar.SidebarListInterface {
...
             add (new_bm);
 
-            return new_bm;
+            show_all ();
+            bm = new_bm;
+            bm.update_free_space ();
         }
 
-        return bm;
+        return bm; // Should not be null (either an existing bookmark or a new one)
```
marbetschar, line 115: I guess adding an `assert (bm != null);` before the return statement would be good to spot potential flaws in the future?

src/View/Sidebar/DeviceListBox.vala
```diff
@@ -130,87 +152,62 @@ public class Sidebar.DeviceListBox : Gtk.ListBox, Sidebar.SidebarListInterface {
             );
         }
 
-        foreach (unowned GLib.Drive drive in volume_monitor.get_connected_drives ()) {
-            bookmark_stoppable_or_removeable_drive_if_without_volumes (drive);
-        } // Add drives not otherwise bookmarked
+        foreach (Volume volume in volume_monitor.get_volumes ()) {
+            bookmark_volume (volume);
```
marbetschar, line 156: `foreach (var volume in ...` should do it

src/View/Sidebar/DeviceListBox.vala
```diff
@@ -130,87 +152,62 @@ public class Sidebar.DeviceListBox : Gtk.ListBox, Sidebar.SidebarListInterface {
...
+        }
 
-        foreach (unowned Volume volume in volume_monitor.get_volumes ()) {
-            bookmark_volume_if_without_mount (volume);
-        } // Add volumes not otherwise bookmarked
+        foreach (GLib.Drive drive in volume_monitor.get_connected_drives ()) {
```
marbetschar, line 159: `foreach (var drive in ...`

src/View/Sidebar/DeviceListBox.vala
```diff
@@ -130,87 +152,62 @@ public class Sidebar.DeviceListBox : Gtk.ListBox, Sidebar.SidebarListInterface {
...
+            bookmark_drive (drive);
+        }
 
-        foreach (unowned Mount mount in volume_monitor.get_mounts ()) {
-            bookmark_mount_if_not_shadowed (mount);
-        } // Bookmark all native mount points ();
+        foreach (Mount mount in volume_monitor.get_mounts ()) {
```
marbetschar, line 163: `foreach (var mount in ...`

src/View/Sidebar/DeviceListBox.vala
```diff
@@ -221,29 +218,29 @@ public class Sidebar.DeviceListBox : Gtk.ListBox, Sidebar.SidebarListInterface {
...
-        if (searched_uuid == null) {
-            return null;
-        }
-
-        foreach (unowned Gtk.Widget child in get_children ()) {
-            if (child is DeviceRow) {
-                if (((DeviceRow)child).uuid == searched_uuid) {
-                    return (DeviceRow)child;
+        if (searched_uuid != null) {
+            foreach (unowned Gtk.Widget child in get_children ()) {
```
marbetschar, line 231: `foreach (unowned var child in ...`

src/View/Sidebar/DeviceListBox.vala
```diff
@@ -252,12 +249,16 @@ public class Sidebar.DeviceListBox : Gtk.ListBox, Sidebar.SidebarListInterface {
     }
 
     public void unselect_all_items () {
-        unselect_all ();
+        foreach (unowned Gtk.Widget child in get_children ()) {
```
marbetschar, line 252: `foreach (unowned var child ...`

src/View/Sidebar/NetworkListBox.vala
```diff
@@ -41,21 +41,49 @@ public class Sidebar.NetworkListBox : Gtk.ListBox, Sidebar.SidebarListInterface
                 select_item ((SidebarItemInterface) row);
             }
         });
+
+        set_sort_func (network_sort_func);
+    }
+
+    private int network_sort_func (Gtk.ListBoxRow? row1, Gtk.ListBoxRow? row2) {
+        string key1 = row1 != null ? ((AbstractMountableRow)row1).sort_key : "";
+        string key2 = row2 != null ? ((AbstractMountableRow)row2).sort_key : "";
```
marbetschar, line 50: Shouldn't we also check for the type here? Also, we can simply use `var`:

```
var key 1 = row1 != null && (row1 is AbstractMountableRow) ? ...
...
```

src/View/Sidebar/DeviceListBox.vala
```diff
@@ -58,38 +48,71 @@ public class Sidebar.DeviceListBox : Gtk.ListBox, Sidebar.SidebarListInterface {
...
             }
         });
+
+        set_sort_func (device_sort_func);
     }
 
-    private DeviceRow? add_bookmark (string label, string uri, Icon gicon,
+    private int device_sort_func (Gtk.ListBoxRow? row1, Gtk.ListBoxRow? row2) {
+        string key1 = row1 != null ? ((AbstractMountableRow)row1).sort_key : "";
+        string key2 = row2 != null ? ((AbstractMountableRow)row2).sort_key : "";
```
marbetschar, line 57: Shouldn't we also check for the type here? Also, we can simply use `var`:

```
var key 1 = row1 != null && (row1 is AbstractMountableRow) ? ...
...
```

src/View/Sidebar/NetworkListBox.vala
```diff
@@ -70,12 +98,15 @@ public class Sidebar.NetworkListBox : Gtk.ListBox, Sidebar.SidebarListInterface
 
         string scheme = Uri.parse_scheme (mount.get_root ().get_uri ());
 
-        /* Some non-native schemes are still local e.g. mtp, ptp, gphoto2 */
+        /* Some non-native schemes are still local e.g. mtp, ptp, gphoto2.  These are shown in the Device ListBox */
         if ("smb ftp sftp afp dav davs".contains (scheme)) {
```
marbetschar, line 102: `var scheme =` should do it. Watch out, `Uri.parse_scheme` may return `null`, so we should probably check this in the if below:

```
if (scheme != null && ..
```

src/View/Sidebar/NetworkRow.vala
```diff
@@ -20,16 +20,47 @@
...
-            null, null, null, //No uuid, drive or volume
-            mount);
+        if (mount != null) {
+            string scheme, hostname;
+            try {
+                var connectable = NetworkAddress.parse_uri (uri, 0);
+                scheme = connectable.scheme;
+                hostname = connectable.hostname;
+            } catch (Error e) {
+                scheme = Uri.parse_scheme (uri);
```
marbetschar, line 45: this may set `scheme` to `null`, which would lead to a weird custom name

src/View/Sidebar/SidebarWindow.vala
```diff
@@ -210,22 +210,10 @@ public class Sidebar.SidebarWindow : Gtk.Grid, Files.SidebarInterface {
...
-                } else if (network_listbox.has_uri (location, out row)) {
-                    network_listbox.select_item (row);
-                }
-
-                return Source.REMOVE;
-            });
+            /* select_uri () will unselect other uris in each listbox */
+            bookmark_listbox.select_uri (location);
+            device_listbox.select_uri (location);
+             network_listbox.select_uri (location);
```
marbetschar, line 216: One space before ` network_listbox` can be removed.

src/View/Sidebar/VolumeRow.vala
```diff
@@ -0,0 +1,206 @@
...
+            /* Mount the device if possible, defer showing the dialog after
+             * we're done */
+            working = true;
+            Files.FileOperations.mount_volume_full.begin (volume, null, (obj, res) => {
+                try {
+                    Files.FileOperations.mount_volume_full.end (res);
+                } catch (Error e) {
+                } finally {
+                    working = false;
+                }
```
marbetschar, line 196: What happens, if `working` was already set to true before we are coming to this?

Also, we probably want to print a warning of the error to the console with something like:

```
warning ("Error while showing mount info: %s", e.message);
```

Otherwise it might get hard to spot potential errors around here in the future.

marbetschar (changes requested):
Well done! :+1: 

Although I'm not familiar with the inner workings of the GLib mount stuff, the overall code looks good to me.

I just left some comments on stuff which may cause (potential) issues in the future, some minor code style thingies as well as some questions where I don't know whats happening.

cassidyjames (dismissed):
So, I found a pretty serious issue presumably with this branch; my USB Blu-ray drive constantly physically ejects itself when connected. :sweat_smile: When reverting to master and rebooting, I no longer get this behavior. This makes the drive unusable.

jeremypw:
@cassidyjames Did that happen when you tried it previously do you know or has it been introduced by recent changes? I do not have an optical drive (either builtin or USB) on my current laptop but I have an old one currently running Hera that does. I'll try to build this PR on the old machine.

jeremypw:
@marbetschar Thanks for your detailed code review!

src/View/Sidebar/DriveRow.vala
```diff
@@ -0,0 +1,126 @@
...
+
+    private void set_visibility () {
+        visible = !drive.has_volumes ();
+        if (!drive.has_media () || !drive.has_volumes ()) {
+            var details = !drive.has_media () ? _("Media ejected") : _("Unformatted");
+            custom_name = drive.get_name () +
+                          "\n" + details + " " +
+                         (drive.is_removable () ? _("This device can be safely unplugged.") : "");
+
+            add_mountable_tooltip (); // Change tooltip to match new custom name.
```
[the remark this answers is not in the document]
jeremypw: Done

libcore/FileOperations/MountUtils.vala
```diff
@@ -17,6 +17,86 @@
...
+                return false;
+            }
+        } else {
+            return false;
+        }
+    }
+
+    public static async bool eject_stop_drive (Drive drive, Gtk.Window? parent) throws Error {
+        // First unmount any mounted volumes
+        foreach (Volume vol in drive.get_volumes ()) {
```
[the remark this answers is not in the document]
jeremypw: Done

libcore/FileOperations/MountUtils.vala
```diff
@@ -17,6 +17,86 @@
...
+                    null
+                );
+                return true;
+            } catch (Error e) {
+                PF.Dialogs.show_error_dialog (_("Unable to stop drive '%s'").printf (drive.get_name ()),
+                                              e.message,
+                                              null);
+                throw e;
+            }
+        } else if (drive.can_eject ()) {
```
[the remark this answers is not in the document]
jeremypw: I think that stopping a drive ejects it first so we do not need to do both.  However, we need some design input anyway on when to unmount, eject or stop when the "eject" button is pressed.  Or do we need different buttons/icons? Or should the button always just unmount and other options (Eject, Stop) be put on the context menu?

libcore/FileOperations/MountUtils.vala
```diff
@@ -17,6 +17,86 @@
...
+                        null
+                );
+                return true;
+            } catch (GLib.Error e) {
+                PF.Dialogs.show_error_dialog (_("Unable to unmount '%s'").printf (mount.get_name ()),
+                                              e.message,
+                                              null);
+                return false;
+            }
+        } else if (mount.can_eject ()) {
```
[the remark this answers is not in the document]
jeremypw: See above reply

src/View/Sidebar/DeviceListBox.vala
jeremypw: Done

src/View/Sidebar/NetworkListBox.vala
jeremypw: Done

src/View/Sidebar/NetworkRow.vala
jeremypw: Fixed

src/View/Sidebar/SidebarWindow.vala
jeremypw: Done

src/View/Sidebar/VolumeRow.vala
jeremypw: This function is called as a callback from a menuitem from `add_extra_menu_items_for_mount ()` in `AbstractMountableRow`, which does check that `working` is not `true`.  So no menuitems are added unless the volume has finished mounting. I can add a `requires` to document this.

jeremypw:
@cassidyjames I did not have any similar issues with a builtin optical drive. I do not have a USB optical drive to test with unfortunately.  There is nothing obvious in the code that would eject a volume or drive without user input. Could you give some more details:

 - Is there a disc in the tray when it ejects itself? Does it also eject itself when empty?
 - If there is a disc in the tray does it spin up and does a Volume bookmark appear before it ejects?
 - After ejecting, is there a Drive bookmark saying "Media ejected ..."?

I have added a couple of debug statements to code that ejects the drive so is there any relevant output from running `G_MESSAGES_DEBUG=all io.elementary.files`?

jeremypw:
@cassidyjames There are some design issues that need UX team input:

 As it stands, if you unmount a USB stick with the button it just unmounts and can be remounted by activating the row.  However it is not obvious that it could be unsafe to unplug it in this state.  To safely unplug you have to eject it using the "Eject Media" option on the context menu.  

How can this be improved?  The difficulty is that it is not straightforward to distinguish a USB memory stick and an optical drive - they both return `true` for `is_removable` and `is_media_removable`  (even my non-removable builtin optical drive gives `is_removable = true` :shrug:).  If a way can be found to do this then the text of the menu option and tooltips could be adjusted to be more meaningful and different behaviours implemented.

The simplest option would be to always eject after unmounting if it is possible to do so (as in master).  The drawbacks of these are that you cannot then remount a plugged in USB stick and the tray of the optical drive would always open.

More complicated would be to have another control for ejecting which appears for an unmounted but not ejected ejectable device and having some visible warning not to unplug the device.  A difficulty would be arranging that devices that cannot in fact be unplugged do not have the warning.

Another possibility would be for safely removable devices to appear in a separate list (would need another PR).

marbetschar:
@jeremypw I'm not very knowledgeable about the API's involved here, so this is a lucky punch - feel free to ignore if its not applicable:

Can we distinguish an optical drive from a USB stick by using [`GLib.Drive.get_start_stop_type`](https://valadoc.org/gio-2.0/GLib.Drive.get_start_stop_type.html)? Asking because [the documentation of `GLib.Drive`](https://valadoc.org/gio-2.0/GLib.Drive.html) states:

> Note that the actual semantics and side-effects of starting/stopping a Drive may vary according to implementation. To choose the correct verbs in e.g. a file manager, use get_start_stop_type.

cassidyjames:
> The simplest option would be to always eject after unmounting if it is possible to do so (as in master). The drawbacks of these are that you cannot then remount a plugged in USB stick and the tray of the optical drive would always open.

Honestly I'm a fan of this. If you hit the "eject" icon on a USB drive, I expect that you should be able to safely unplug it. And if you "eject" e.g. a flash drive, I would also expect that you'd need to unplug and replug it to get it to show up again. For the optical disc case, I feel similarly; if you hit the "eject" button to unmount it, I expect the tray to open. Iirc that's similar to what happens in Windows, as well.

jeremypw:
The problem I am having is that it is difficult to ensure that a USB stick and an optical drive are dealt with differently.  According to @cassidyjames, the USB stick should be ejected and disappear from the sidebar when ejected (which seems reasonable since it cannot be remounted if it has been ejected) however, an optical drive would normally be expected to remain visible in the sidebar even when the media has been ejected.  The values returned by GLib.Drive for USB memory sticks is not always what you might expect (and may depend on the make of the stick).

I'll continue to work on this using the basis of always ejecting the drive if possible.

jeremypw:
@marbetschar Thanks for the suggestion!  I'll give it a try.

jeremypw:
@marbetschar The only difference in properties reported by GLib.Drive that I have found for my internal optical disc drive and a USB memory stick (with Ventoy installed) was for the `sort-key`.  The internal drive had a key including "coldplug" whereas the USB memory stick contained "hotplug".  Presumably this would not help for an external USB optical drive but some improvement is possible.  

I also found that devices often report misleading properties through GLib.Drive anyway.  I have an external USB hard disk that reports `false` for `can-stop` yet if run code to stop it, it switches off as expected.  So the latest version of this PR attempts to eject and stop all drives as a `safely remove` and just ignores any resulting error.  This has the benefit of removing the drive from the sidebar whereas if you just eject it, it remains visible (but currently unusable).

Maybe in future we can use UDisks or another service to provide more detailed/accurate information but I do not want to do that in this PR.

src/View/Sidebar/DeviceListBox.vala
jeremypw: Done

src/View/Sidebar/BookmarkRow.vala
jeremypw: It is overridden in `DriveRow.vala`

src/View/Sidebar/DeviceListBox.vala
jeremypw: Done

src/View/Sidebar/NetworkListBox.vala
jeremypw: As we are in full control of what rows are added to this listbox I am not sure a check is necessary but I will put one in just to be on the safe side.

cassidyjames:
@jeremypw am I right in that this just requires code re-review at this point? Have the UX questions been addressed to your satisfaction? I don't have access to an internal disc drive, but will continue using this branch and note any strange behavior I see, but otherwise I think once we're happy code-wise, we're good to merge.

jeremypw:
@cassidyjames Yes, just awaiting a code review now.  It would be good to get it merged well before the next stable release so that any issues arising from exposure to a wider range of hardware can be fixed.

jeremypw:
One small thing I noticed using this is that it shows an (unusable) entry for a broken internal SSD that I have since spilling tea on the keyboard :man_facepalming: called "Unnamed Drive (/dev/sda) Unformatted or no media".  I am not sure it worth trying to address such an exceptional situation.  It might be considered to not show such entries except when they relate to e.g. optical drives with removable media.

cassidyjames:
@jeremypw I have a feeling that would be fine to ignore for now. Eventually it might make sense to, since there's no action that can be taken on the drive from within Files. If we either added disk formatting or a way to open a disk utility from in Files, then it might make sense to show. 

Either way, I think I'd be fine dealing with that later.

danrabbit:
With this branch I'm also seeing a "Unformatted or no media" device, even though I only have a single drive in this computer:

Screenshot from 2021-11-16 15 04 40

jeremypw:
@danrabbit Thanks for trying the PR.  That's strange if you only have one SSD and that is formatted.  I assume the partition(s) on that SSD are displayed properly as well?   Does gparted show any unformatted partitions on the SSD (or a device you didn't know you had :stuck_out_tongue_winking_eye: )

jeremypw:
I am inclined to hide the unformatted/no media drives for now as there is not much you can do with them at the moment and they could be confusing.  I do not think that an optical drive with no media not appearing is particularly problematic as users tend to know whether they have an optical drive or not.  If it is desired to see optical drives with no media displayed I could special case them.

jeremypw:
If required I could add UI to toggle visibility of unformatted/unpartitioned/stopped drives

danrabbit:
Hey @jeremypw sorry for taking so long to reply! I think I agree that hiding makes sense here since like you said we can't really do anything with them. Any objections @cassidyjames ?

Gparted is only showing the one device. It's unencrypted:
Screenshot from 2021-12-07 14 19 35

cassidyjames:
@danrabbit no objections!

jeremypw:
@danrabbit Interesting.  Although on my system `gparted` does not show my dead SSD but Files was making a Sidebar entry for it so :shrug:.  Will need to get to the bottom of it before making unformatted drives visible but I guess it can be left for now.

danrabbit (approved):
Thank you for your attention to detail here Jeremy!

PR #1760, opened 2021-07-16
Allow dropping uris onto Storage and Network rows.

### LINKED ISSUE #1747: elementary OS 6 Early Access file drag and drop doesnt work on Storage sidebar
Hi.

When I use a USB drive, I'm always dragging files with the mouse to my drive's row in Storage part of the sidebar.
In Odin beta 2, with all the updates as I'm writing this, this cannot be achieved. Hovering with the file over USB drive does not change the pointer and it does not "drop" the files when I release the button.

## Platform Information
Pantheon-files version: 4.5.0+r4551+pkg100~daily~ubuntu6.1

I couldn't say if that was the case with the older version of pantheon-files

Cheers

jeremypw:
Fixes #1747 (regression)

Dropping files onto (but not between) storage and network rows is restored.


danrabbit (approved):
Works as expected :rocket:

PR #1764, opened 2021-07-19
Move Miller animation into Miller class

jeremypw:
Following a comment from @tintou and noting that only the Miller class uses the animation from libwidgets, I have moved the animation functions into the Miller class instance itself.


danrabbit (approved):
Makes sense!

tintou:
Don't forget the POTFILES

PR #1790, opened 2021-08-17
Improve drop onto/between sidebar bookmarks

### LINKED ISSUE #1769: Drag and drop to sidebar is needing to be easier and better visual explanations.
## Problem

Drag and drop to sidebar is very difficult to do and to understand what is going to happen. You have to be really precise to move to trash or bookmark on the sidebar. Also there is no hover around sidebar items while moving files on top of them.

https://user-images.githubusercontent.com/87835/126909867-a8d2cd4d-1afb-40ae-955f-0de9075ab3b5.mp4

## Proposal

 - Make hover on the sidebar directories while drag and drop
 - Bigger action icons
 - Little bit less sensitive between move to and bookmarked

jeremypw:
Fixes #1769 

* Larger target for drop onto bookmark
* Additional visual cue for drop between

There is an unavoidable (?) difficulty when moving down the list in that when the drop-between revealer closes, the following bookmark moves relative to the pointer so the pointer is near to the lower edge and the target is apparently smaller.


tomaswarynyca (dismissed):

jeremypw:
@tomaswarynyca Thanks for the review!

JoseExposito (dismissed):
Thanks for the PR! Code looks fine :tada: 

There is a small visual issue when the separator is being hidden. It is pushed up to the row above it instead of keeping centered. It is not a big deal, but I guess it's worth to hear the opinion of @elementary/ux

jeremypw:
@JoseExposito Thanks for the review! I'll try a different revealer transitions to see if there is a better.

jeremypw:
@JoseExposito Using a SLIDE_UP transition seems to stop the separator appearing to move - what do you think?

cassidyjames (approved):
Looks good to me now!

JoseExposito:
Looking great, thanks! :rocket:

PR #1809, opened 2021-08-23
FileUtils: Remove unnecessary trailing whitespace

ryonakano:
Find while revisiting translations of Files on Weblate


jeremypw (approved):
Nice catch! Thanks.

PR #1812, opened 2021-08-24
Release 6.0.2

danrabbit:
Drafting in case we want to squeeze in other fixes this week


jeremypw:
Commit a4038b9af23b83703895612253a9cc1fabf414b8 probably justifies a new release alone. 

It would be nice to get #1752 in before too long as it fixes some important issues like ejecting USB sticks safely but it needs careful independent testing for regressions so it probably won't make this week.

cassidyjames:
@decathorpe no, we just missed it. Sorry about that!

PR #1822, opened 2021-09-01
Re-Adding "Send by Email"

marbetschar:
[Using the `org.freedesktop.portal.Email` portal](https://flatpak.github.io/xdg-desktop-portal/portal-docs.html#gdbus-org.freedesktop.portal.Email) as suggested by @Marukesu here: https://github.com/elementary/mail/pull/674.

~~Unfortunately I can't get this to work. The context menu item is shown and I'm able to click on it - but the portal itself is never shown. What am I missing?~~


plugins/send-by-email/Portal.vala
```diff
@@ -0,0 +1,45 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+namespace Portal {
+    const string DBUS_DESKTOP_PATH = "/org/freedesktop/portal/desktop";
+    const string DBUS_DESKTOP_NAME = "org.freedesktop.portal.Desktop";
```
jeremypw, line 20: Shouldn't these end in "Email" not "Desktop"?

jeremypw:
Just testing this: At the moment it is failing because the file_descriptors are not being created (returning "-1").  `dbus-monitor` shows that the method call is being made and the contents otherwise look OK.

plugins/send-by-email/plugin.vala
```diff
@@ -0,0 +1,125 @@
...
+        try {
+            var portal = Portal.Email.get ();
+
+            window_export.begin ((obj, res) => {
+                var options = new HashTable<string, Variant> (str_hash, str_equal);
+                options["handle_token"] = Portal.generate_token ();
+
+                int[] file_descriptors = {};
+                foreach (var file in files) {
+                    file_descriptors += Posix.open (file.get_uri (), Posix.O_RDONLY);
```
jeremypw, line 39: Should use `file.get_path ()`

jeremypw:
This is what I get from `dbus-monitor` now:
```
method call time=1630519639.938736 sender=:1.596 -> destination=org.freedesktop.portal.Desktop serial=142 path=/org/freedesktop/portal/desktop; interface=org.freedesktop.portal.Email; member=ComposeEmail
   string "x11:38005c2"
   array [
      dict entry(
         string "handle_token"
         variant             string "io_elementary_files_1493378944"
      )
      dict entry(
         string "attachment_fds"
         variant             array [
               int32 28
            ]
      )
   ]

```

However, the email program is not launched nor is any error or response given from the portal.  Not sure how to check whether the portal itself is actually working ...

Pasting a "mailto" URI into Firefox does launch the email program but I am not sure whether FireFox is using the portal.

Marukesu:
So, there's two issues:

One is that the portal segfault when there's no address on the options dict, looking at the portal code, it seems that the portal needs at least one address (see flatpak/xdg-desktop-portal-gtk#343).

The second is that vala only support sending a single file descriptor in a DBus method (see [GNOME/vala#338](https://gitlab.gnome.org/GNOME/vala/-/issues/338)). So we need to use the underlying [`DBusProxy.call_with_unix_fd_list`](https://valadoc.org/gio-2.0/GLib.DBusProxy.call_with_unix_fd_list.html). Looking at the code on libportal, we need to add the file descriptor to the `UnixFdList` and the index returned by `UnixFdList.append` to the **attachment_fds** in options.

you can run `/usr/libexec/xdg-desktop-portal-gtk -vr` to check if the mailto URI is sending the attachments correctly.

jeremypw:
@Marukesu  Thanks for the diagnosis. According to https://flatpak.github.io/xdg-desktop-portal/portal-docs.html#gdbus-org.freedesktop.portal.Email "All the keys in the options are are optional".  Is that not the case if an address is required?

Marukesu:
I would say it's a bug on the GTK backend of the portal. if it was really required the frontend would throw a error.

marbetschar:
> The second is that vala only support sending a single file descriptor in a DBus method (see [GNOME/vala#338](https://gitlab.gnome.org/GNOME/vala/-/issues/338)). So we need to use the underlying [`DBusProxy.call_with_unix_fd_list`](https://valadoc.org/gio-2.0/GLib.DBusProxy.call_with_unix_fd_list.html). Looking at the code on libportal, we need to add the file descriptor to the `UnixFdList` and the index returned by `UnixFdList.append` to the **attachment_fds** in options.

@Marukesu can you please provide some more info how exactly this would be implemented in Vala? I'm not too familiar with DBus and file descriptors yet, so I can't follow unfortunately.

plugins/send-by-email/Portal.vala
```diff
@@ -0,0 +1,45 @@
...
+
+    public static string generate_token () {
+        return "%s_%i".printf (
+            GLib.Application.get_default ().application_id.replace (".", "_"),
+            Random.int_range (0, int32.MAX)
+        );
+    }
+
+    [DBus (name = "org.freedesktop.portal.Email")]
+    interface Email : Object {
```
Marukesu, line 31:
```suggestion
    interface Email : DBusProxy {
```

plugins/send-by-email/Portal.vala
```diff
@@ -0,0 +1,45 @@
...
+        public abstract uint version { get; }
```
Marukesu, line 32:
```suggestion
        [DBus (name = "version")]
        public abstract uint version { get; }
```

DBus properties needs the `[DBus (name)]` annotation to work.

plugins/send-by-email/Portal.vala
```diff
@@ -0,0 +1,45 @@
...
+        public static Email @get () throws IOError, DBusError {
+            if (email == null) {
+                var connection = GLib.Application.get_default ().get_dbus_connection ();
+                email = connection.get_proxy_sync<Email> (DBUS_DESKTOP_NAME, DBUS_DESKTOP_PATH);
+            }
+
+            return email;
+        }
+
+        public abstract ObjectPath compose_email (string window_handle, HashTable<string, Variant> options) throws IOError, DBusError;
```
Marukesu, line 43:
```suggestion
        [DBus (visible = false)]
        public ObjectPath compose_email (string window_handle, HashTable<string, Variant> options, UnixFDList? attachements) throws IOError, DBusError {
            var options_builder = new VariantBuilder (VariantType.VARDICT);
            object.foreach ((key, val) => {
                options_builder.add ("{sv}", key, val);
            });
            
            var response = call_with_unix_fd_list_sync (
                "ComposeEmail",
                new Variant ("(sa{sv})", window_handle, options_builder),
                DBusCallFlags.NONE,
                -1,
                attachements
            );
            
            return (ObjectPath) response.get_child_value (0).get_string ();
        }
```

Since we can't use the vala generated method, i passed the UnixFDList as a parameter (the method needs to be non-visible for that) and did the right call inside.

plugins/send-by-email/plugin.vala
```diff
@@ -0,0 +1,132 @@
...
+
+            window_export.begin ((obj, res) => {
+                var options = new HashTable<string, Variant> (str_hash, str_equal);
+                options["handle_token"] = Portal.generate_token ();
+
+                int[] file_descriptors = {};
+                foreach (var file in files) {
+                    file_descriptors += Posix.open (file.get_path (), Posix.O_RDONLY);
+                }
+                options["attachment_fds"] = file_descriptors;
```
Marukesu, line 41:
```suggestion
                var files_builder = new VariantBuilder (new VariantType ("ah"));
                var file_descriptors = new UnixFDList ();
                foreach (var file in files) {
                    var fd = Posix.open (file.get_path (), Posix.O_RDONLY | Posix.O_CLOEXEC);
                    if (fd == -1) {
                        warning ("send-by-mail: cannot open file: '%s'", file.get_path ());
                        continue;
                    }

                    files_builder.add ("h", file_descriptors.append (fd));
                }
                
                options["attachment_fds"] = files_builder.end ();
```
Here we create a UnixFDList and a VariantBuilder for the handles, we insert the fd in the list, and the returned index on the builder. 

Also, skip non-valid file descriptors.

plugins/send-by-email/plugin.vala
```diff
@@ -0,0 +1,132 @@
...
+
+                /** Even though the org.freedesktop.portal.Email portal specs
+                * claims that "all the keys in the options are are optional",
+                * the portal does not work if no "addresses" key is passed.
+                * This is probably a bug in the Gtk backend of the portal.
+                */
+                options["addresses"] = new Variant ("as", null);
```
Marukesu, line 48:
```suggestion
                if (portal.version > 2) {
                    options["addresses"] = new Variant ("as", null);
                }
```

The `addresses` key was added on the version 3 of the portal.

plugins/send-by-email/plugin.vala
```diff
@@ -0,0 +1,132 @@
...
+
+                try {
+                    var handle = window_export.end (res);
+                    portal.compose_email (handle, options);
```
Marukesu, line 52:
```suggestion
                    portal.compose_email (handle, options, file_descriptors);
```

Marukesu (commented):
I played a bit with the code, with the suggestions, the portal generates a valid mailto uri with the  attachments as expected.

However, i don't see the attachments on mail composer, and it failed to send with ` mail_backend_session_send_email: assertion 'from != NULL' failed`, but that are issues in mail.

marbetschar:
@Marukesu thank you so much for breaking this down - that helped tremendously! Thinks are working now on my end as expected together with https://github.com/elementary/mail/pull/682.

So I'll guess we are ready for a review here :)

jeremypw:
I can confim that this now brings up the Mail program with a new mail with the selected files attached as expected :tada: 

I cannot fully confirm that the mail gets sent as expected since I am having trouble setting up Mail to access my GMail account (even with 2FA off and less secure app access enabled).  Is there an idiot's guide for doing this available?

plugins/send-by-email/plugin.vala
```diff
@@ -0,0 +1,146 @@
...
+                handle += h;
+                window_export.callback ();
+            });
+            yield;
+
+            if (handle != "wayland:") {
+                return handle;
+            }
+            return "";
+
```
jeremypw, line 97: Blank line should be moved after end of code block, before `return`.

jeremypw (changes requested):
Just a few minor code style nitpicks.  Otherwise seems to work as far as I can test it.

marbetschar:
@jeremypw I'm able to add Gmail with the following configuration:

1. Go to your webmail settings into the tab Forwarding & POP/IMAP: https://mail.google.com/mail/u/0/#settings/fwdandpop
2. Enable IMAP there
3. Enable "Less Secure Apps" here: https://support.google.com/accounts/answer/6010255
4. Add your Gmail account with the default settings via Online Accounts

At least on my end this procedure works w/o any issue.

jeremypw:
@marbetschar Thanks, that sounds pretty much like what I was was doing but I'll try again. I am trying to add my elementaryos.org account.

marbetschar:
@jeremypw oh, ok then you might need https://github.com/elementary/switchboard-plug-onlineaccounts/pull/225 by installing Online Accounts from current master. Not sure though, lost a bit track which change exactly fixed what :)

Marukesu:
@jeremypw When the dialog opens did the sender combobox is setted and active? when mail is closed, if i use this PR, it's empty and inactive, but if mail is already open, the sender is setted, and the combobox is active.

marbetschar:
@jeremypw mea culpa! I was so focused to get this fixed that I totally missed to test this scenario as well 😞

marbetschar:
@jeremypw yes, that's correct. Mail is distributed as deb for the time beeing because of https://github.com/elementary/mail/issues/591

jeremypw:
@marbetschar OK, thanks.  As a stable release of Files has just been made I think it safe enough to merge this now.

jeremypw (approved):
Works as expected provided Mail is not Flatpak'd.

PR #1829, opened 2021-09-05
Paste into background folder by default

### LINKED ISSUE #1827: Paste files doesn't work when multiple items are currently selected
## What Happened

When I try to paste files into a folder, but multiple items in that folder are selected (have the accent color highlight), it doesn't paste and just sends an error sound. This seems to only happen if 2 or more files are selected. If only one file is selected in the destination, the paste proceeds as usual. However, if I select another, it fails. This appears to be independent of the number of copied files.

## Expected Behavior

Paste should still occur.

## Steps to Reproduce

1. Copy a file from any directory
2. Open a new directory
3. Select at least 2 files in that directory (using shift-click, drag-select, etc.)
4. Try to paste the copied files using the ctrl+v shortcut
5. Error beep sound will play, and the files will not paste.
6. Deselect files in the destination directory and re-select only 1 file
7. Try to paste the copied files using ctrl+v again
8. The paste will proceed

### Logs

I imagine this is it:
```
** (io.elementary.files:190997): WARNING **: 09:04:47.747: AbstractDirectoryView.vala:3054: Cannot paste into a multiple selection
```

## Platform Information

Odin 6.0, up to date as of 9/2/2021.

jeremypw:
Fixes #1827 

* `<Ctrl>v` always pastes into background (in line with other file managers (e.g. Nautilus, Dolphin etc)
* `<Shift><Ctrl>v` pastes into selected folder

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -82,7 +82,8 @@ namespace Files {
 
         const GLib.ActionEntry [] COMMON_ENTRIES = {
             {"copy", on_common_action_copy},
-            {"paste-into", on_common_action_paste_into},
+            {"paste-into", on_common_action_paste_into}, // Paste into selected folder
+            {"paste", on_common_action_paste}, // Paste into background folder
             {"open-in", on_common_action_open_in, "s"},
             {"bookmark", on_common_action_bookmark},
             {"properties", on_common_action_properties},
@@ -223,6 +224,7 @@ namespace Files {
                         action_set_enabled (selection_actions, "cut", false);
                         action_set_enabled (common_actions, "copy", false);
                         action_set_enabled (common_actions, "paste-into", false);
+                        action_set_enabled (common_actions, "paste", false);
 
                         /* Fix problems when navigating away from directory with large number
                          * of selected files (e.g. OverlayBar critical errors)
@@ -1262,6 +1264,19 @@ namespace Files {
             clipboard.copy_files (get_selected_files_for_transfer (get_files_for_action ()));
         }
 
+        private void on_common_action_paste (GLib.SimpleAction action, GLib.Variant? param) {
+            if (clipboard.can_paste && !(clipboard.files_linked && in_trash)) {
+                var target = slot.location;
+                clipboard.paste_files.begin (target, this as Gtk.Widget, (obj, res) => {
+                    clipboard.paste_files.end (res);
+                    if (target.has_uri_scheme ("trash")) {
+                        /* Pasting files into trash is equivalent to trash or delete action */
+                        after_trash_or_delete ();
+                    }
+                });
+            }
+        }
+
         private void on_common_action_paste_into (GLib.SimpleAction action, GLib.Variant? param) {
             var file = get_files_for_action ().nth_data (0);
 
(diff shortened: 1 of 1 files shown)


src/View/AbstractDirectoryView.vala
```diff
@@ -2212,6 +2242,10 @@ namespace Files {
 
                 if (in_trash) {
                     if (clipboard != null && clipboard.has_cutted_file (null)) {
+                        paste_menuitem.add (new Granite.AccelLabel (
+                            _("Paste into Trash"),
+                            "<Ctrl>v"
+                        ));
```
danrabbit, line 2248: I don't see where this is used. It doesn't look like we get these menuitems when right clicking in Trash or right clicking on the Trash bookmark item

danrabbit (commented):
Makes sense with the change to the shortcut behavior, I think.

I'm not sure about showing both "Paste" and "Paste Into" when right clicking on a folder. That seems a bit confusing that you'd have an action on a folder which doesn't effect the folder

jeremypw:
@danrabbit Thanks for the review.  I'll have another look at this.

src/View/AbstractDirectoryView.vala
```diff
@@ -3155,8 +3198,23 @@ namespace Files {
...
+                                                            _("You do not have permission to change this location"),
+                                                            window as Gtk.Window);
+                        }
+
+                        res = true;
+                    }
+
+                    break;
+
+                case Gdk.Key.V: // Alternative paste shortcut action - paste into selected folder if there is one
```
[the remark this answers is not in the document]
jeremypw: Agreed, but that is for another day. I have now combined the two cases to make it slightly clearer.

src/View/AbstractDirectoryView.vala
danrabbit: Ah okay I understand. In that case, I'm not sure we need to add a new string to be localized here. I agree this is probably not a typical workflow and when it's in the context menu I don't think there's any ambiguity about what "Paste" will do since it's the same as any other folder

danrabbit:
I think we're almost there. Left a couple follow up comments on the inline discussions. AFAICT the menu items all make sense and keyboard shortcuts work as expected :+1:

src/View/AbstractDirectoryView.vala
jeremypw: Done

src/View/AbstractDirectoryView.vala
```diff
@@ -2152,16 +2165,33 @@ namespace Files {
...
-                                paste_menuitem.label = _("Paste Link into Folder");
+                                paste_into_menuitem.add (new Granite.AccelLabel (
+                                    _("Paste Link into Selected Folder"),
+                                    "<Shift><Ctrl>v"
+                                ));
                             } else {
-                                paste_menuitem.label = _("Paste into Folder");
+                                paste_into_menuitem.add (new Granite.AccelLabel (
+                                    _("Paste into Selected Folder"),
+                                    "<Shift><Ctrl>v"
```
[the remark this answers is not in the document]
jeremypw: Reverted changes to translatables

danrabbit (approved):
Nice work!

PR #1832, opened 2021-09-10
Allow blank password attempt when connecting with a user name

### LINKED ISSUE #1666: There is no way to connect ssh server just with key (without password)
## What Happened
All keys are exist for connecting to my ssh server but when I try to connect through files app, it's not allowing me to connect. Connect button is greyed out. Oblige to entering password.

## Expected Behavior
Allow me to connect ssh servers without password.

## Steps to Reproduce

1. Try to connect ssh server from files app (don't enter password)

## Platform Information
```bash
~ $ head -2 /etc/os-release
NAME="elementary OS"
VERSION="6 Early Access"
```

jeremypw:
Fixes #1666 

Allow connection attempt with blank password (connection with a key may be enabled) - if that fails then require a non-blank password.

Rather than try and determine in advance whether connection without a password will succeed, just try and adjust validation requirements if it fails due to no password.

Do not show the remember password checkbox for blank passwords, do not  try to save blank passwords.


cassidyjames (approved):
This makes sense and works for me! I see a couple of UI improvements that could be made (e.g. using a revealer for the remember_checkbutton), but that's not unique to this branch.

PR #1835, opened 2021-09-14
Fix secondary-click behaviour wrt selection/deselection

### LINKED ISSUE #1221: Right-clicking in a blank area in the folder when files/folders are selected shows options for those files/folders rather than the folder
## Prerequisites

## Describe the bug
Selecting any amount of files or folders and then right-clicking on a blank space in the folder shows the context menu for those files rather than the parent folder.

I don't actually know if this is a bug or intended behaviour. The problem is that when you have a grid of lots of files/folders in Grid view, you have to find a blank part of the folder that you can right-click. With the current right-click behaviour you often find you have to first deselect the items by left clicking on blank space (usually you have to find it again as the pointer has probably moved onto the context menu to some extent by then) and then right clicking there again.

### To Reproduce
1. Open Files
2. Go to a folder with at least one file in it
3. Select any amount of files
4. Right click on a blank area within the folder

### Expected behaviour
The context menu for the parent folder shows.

image

## Platform Information

It's elementary OS 5.1.

## Additional context
I think if this is fixed, it may be a good idea to retain the selection of files/folders so that your selection isn't lost if you right clicked in the wrong place (e.g. if it's a very specific, complex selection of files and folders).

### LINKED ISSUE #1834: IconView: Secondary-click on selected file deselects it (sometimes)
## What Happened

In IconView secondary-clicking on a selected file sometimes deselects whereas in ListView it does not. Deselection seems to occur if you click below the name but not if the name occupies four lines. If you click on the text of the name or on the icon then deselection does not occur.

## Expected Behavior

Secondary-clicking anywhere on a selected item should not deselect it.

## Steps to Reproduce

1. Open IconView
2. Select an item with a short name
3. Secondary-click on the icon - does not deselect. Press Esc to close menu
4. Secondary click on the text - does not deselect. Press Esc to close menu
5. Secondary click on the blank below the text (within selected area) - item is deselected and the background menu shows

## Platform Information

Odin 6 stable

jeremypw:
Fixes #1834 
Fixes #1221 

* Handle all selection/deselection within switch clauses
* Handle clicks on empty space within switch clauses
* Replace if/else if with switch clause
* Drop unneeded code
* No rubberband with secondary click when determining ClickZone
* Fix ClickZone returned by IconView when not rubberband

In view of the relative intricacy of click handling, this PR needs substantial testing for regressions.

### DIFF
--- a/data/io.elementary.files.appdata.xml.in.in
+++ b/data/io.elementary.files.appdata.xml.in.in
@@ -63,6 +63,8 @@
           <li>Open bookmarks in a new tab with Ctrl + Click</li>
           <li>Fix dropping uris onto storage devices and network locations in sidebar</li>
           <li>Fix restoring tabs after a system restart</li>
+          <li>Don't unselect when secondary clicking blank space around a file or folder</li>
+          <li>Show the folder context menu when secondary clicking outside of a selection</li>
         </ul>
         <p>Other updates:</p>
         <ul>
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -3379,13 +3379,7 @@ namespace Files {
             drag_x = (int)(event.x);
             drag_y = (int)(event.y);
 
-            click_zone = get_event_position_info (event, out path, true);
-            /* certain positions fake a no path blank zone */
-            if (click_zone == ClickZone.BLANK_NO_PATH && path != null) {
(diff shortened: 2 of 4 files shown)


danrabbit (approved):
I can confirm this fixes the linked issues. Let's get this in now so we have a good couple weeks of testing in daily :)

Nice job cleaning up this code as well!

PR #1838, opened 2021-09-19
Fix middle-click handling in pathbar

### LINKED ISSUE #1837: "Middle click paste" option doesn't work with text from files path bar
## What Happened
Middle click paste doesn't work when the selected text is from the path bar of the file.

## Expected Behavior
The selected text from the path bar of files should be pasted if middle-click mouse is pressed.

## Steps to Reproduce
1. Go to system settings -> Mouse & Touchpad -> Clicking behaviour and activate the "Middle click paste" option
2. Go to files and select any text from the path bar
3. Use the middle click to paste the selected text, it fails

## Additional information
This feature works with any other app I had tested, including chrome, firefox, Inkscape, etc...

## Platform Information
image

jeremypw:
Fixes #1837 

Also small cleanups/simplifications of existing relevant code.

* Only grab focus if not focussed
* Remove side-effect (calling base event handler) from testing for button event on entry icon
* More appropriate name for boolean function
* Only handle primary button release and pass to Gtk handler if breadcrumb not activated

### DIFF
--- a/libwidgets/Chrome/BasicBreadcrumbsEntry.vala
+++ b/libwidgets/Chrome/BasicBreadcrumbsEntry.vala
@@ -233,8 +233,10 @@ namespace Files.View.Chrome {
         }
 
          protected virtual bool on_button_release_event (Gdk.EventButton event) {
-            /* Only activate breadcrumbs with primary click when they are showing and not hidden by placeholder */
-            if (event.button == Gdk.BUTTON_PRIMARY && placeholder == "" && !is_icon_event (event)) {
+            /* Only activate breadcrumbs with primary click when pathbar does not have focus and breadcrumbs showing.
+             * Note that in home directory, the breadcrumbs are hidden and a placeholder shown even when pathbar does
+             * not have focus. */
+            if (event.button == Gdk.BUTTON_PRIMARY && !has_focus && !hide_breadcrumbs && !is_icon_event (event)) {
                 reset_elements_states ();
                 var el = get_element_from_coordinates ((int) event.x, (int) event.y);
                 if (el != null) {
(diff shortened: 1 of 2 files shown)


danrabbit (changes requested):
Hm I'm experiencing a regression in this branch where the pathbar always loses focus after selecting some text in it while in home. It looks like switching to another directory and then back to home fix it for some reason

jeremypw:
The home directory is special-cased in that the placeholder shows instead of the breadcrumbs - that maybe causing the problem. I'll look into it.

danrabbit (approved):
This code seems much better, but I'm not sure it fixes #1837. It sounds like the reporter is saying they've highlighting text from the files pathbar and want to paste it elsewhere. But it looks like Files' pathbar is completed cleared when the window loses focus.

So, merging this because it seems better than before, but not closing the linked issue since I think it can still be reproduced

jeremypw:
Hmm, yeah the issue report is not exactly clear.  I tested both with selecting text from the pathbar (in location entry mode) and selecting text from e.g. Code - in both cases pasting into the pathbar (in location entry mode).  Is middle-click supposed to both copy *and* paste the selected text?  I assumed that it was only intended to paste text already in the clipboard (this seems to be the native behaviour of Gtk.Entry) but I notice that Terminal does not require you to pre-copy the text.  This PR just restores the native behaviour of the underlying Gtk.Entry.  A further PR will be required if both copy and paste the selected text is required.  It will have to be handled in Files.

jeremypw:
Well - that's weird. This did not seem to be working after installing the merged master - even with Gtk.Entrys in other apps. But after some fiddling around it started working - even after resetting all the changes I made and after rebooting :shrug:  It seems the system can get into a state where middle-click paste stops working but it is probably not due to Files.

I'll now get on with looking at pasting text *from* the Files pathbar elsewhere ...

PR #1839, opened 2021-09-19
Discontinue long press support

jeremypw:
Discontinuue showing breadcrumb context menu if primary button held down.  This conflicts with the system accessibility setting for long press secondary click.  If this setting is on then the context menu gets shown twice after a long primary button click- once before the button is released and once afterwards.


jeremypw:
I am unaware of any other elementary app providing similar behaviour independent of the accessibility setting so there seems no good reason to keep it.

danrabbit (approved):

danrabbit:
Nice little cleanup

PR #1843, opened 2021-09-24
Do not try to thumbnail audio files without album art

### LINKED ISSUE #1842: Audio file icons disappear when scrolling or changing view.
## What Happened

Hi! I found this bug. Audio file icons disappear when scrolling or changing view (I tried with mp3, ogg, and wav).
Documents, videos and images icons work ok.

## Expected Behavior

Icons should not disappear.

## Steps to Reproduce

https://user-images.githubusercontent.com/52709674/134596433-d0b3b9e5-b63b-44c8-a8fa-d1f3e9f509f5.mp4

### Logs

--

## Platform Information

elementary OS 6

jeremypw:
Fixes #1842


danrabbit:
Not sure this is right, since I do get thumbnails for audio files here:
Screenshot from 2021-09-24 10 54 07

jeremypw:
@danrabbit  Is that with master?  What is the thumbnail - the album cover?  Looks like I need to look deeper into this.

danrabbit:
Yes that's correct it's the album cover. I'm AFK right now, but I'll test the scrolling and view switching Monday :)

jeremypw:
So, these seems to be due to a bug (feature) of the tumbler thumbnailer whereby if an audio file contains no album art it does not get reported either as an "error" or "finished" by the proxy - so its thumbstate remains as "LOADING".  Unfortunately the "finished" signal does not include a list of uris sent in the original request so we will have to keep our own list so that we can check whether any files have been silently ignored.

danrabbit (approved):
Looks good on my end! Probably want to update appdata here and remove that commented line, but otherwise :+1:

PR #1846, opened 2021-09-25
Removing annoying message to report bugs

QWxleA:
Files works great from the command-line, but the standard message it prints to "Report any issues/bugs you might find to https://github.com/elementary/files/issues" starts to get old after 417 times ;-)

This message is more useful for the devs than the user. As a user, I would this line to go away. There is no "quiet" option, so this is the best next thing.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -130,7 +130,7 @@ public class Files.Application : Gtk.Application {
             return 1;
         };
 
-        message ("Report any issues/bugs you might find to https://github.com/elementary/files/issues");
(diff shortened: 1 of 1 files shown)


ChildishGiant:
I feel like adding a quiet flag would be a much better solution

BAProductions:
> I feel like adding a quiet flag would be a much better solution

I agree

tintou:
Oh I thought this was an issue and created #1851, just remove the line as we don't gain anything keeping it aroung, git is here for this :slightly_smiling_face:

PR #1847, opened 2021-09-25
Do not lose pathbar entry text if another window is focused

### LINKED ISSUE #1837: "Middle click paste" option doesn't work with text from files path bar

jeremypw:
Fixes #1837  (part 2)

Allows middle-click pasting of selected text in pathbar into another window.


danrabbit (changes requested):
It looks like with this branch you get both the path and the placeholder at the same time on focus out:

Screenshot from 2021-11-09 11 21 18

jeremypw:
Ah, this only happens in the Home directory - I must have only tested it elsewhere. Will fix.

danrabbit (approved):
Nice fix!

PR #1851, opened 2021-10-01
Application: Remove startup message about issue reports

tintou:
There is an app in the system to know where to report bugs.

Closes #1846


tintou:
Oops, #1846 is already a PR

PR #1857, opened 2021-10-07
Unmount instead of ejecting drives with multiple volumes

vjr:
If a removable drive has multiple partitions and we mount one or more, then clicking the unmount/eject button ejects the drive including unmounting any other partitions that are still mounted, preventing the ability to only unmount the intended partition while not affecting the rest, as a user might expect.

This PR sets the eject button and tooltip to say "unmount" instead of "eject" and also fixes the blank custom name (label) in the tooltip due to the virtual property `can_eject` being called too early in the object construction.

If the drive has only a single volume/partition then the behaviour is unchanged - the button ejects and the tooltip shows "eject...".


vjr:
@jeremypw  here's another diff for your PR https://github.com/elementary/files/pull/1752 if you think it's useful.

Issue #1821, opened 2021-08-31
DnD target reveals below "Trash" bookmark, but dropping on it fails


jeremypw:
## What Happened

Dragging a folder over the sidebar causes bookmarks to move apart to allow creation of a new bookmark by dropping. However, if the drag goes below "Trash" a target is revealed but dropping onto it fails with an error dialog.

## Expected Behavior

No target should be revealed below "Trash" as that is fixed as the last bookmark.

## Platform Information

Odin daily

PR #1848, opened 2021-09-26
Allow inserting bookmark after Recent

jeremypw:
Fixes #1819 
Fixes #1821 

This PR implements better control of where dragged items can be dropped in the sidebar.

* Add `can_insert_before` and `can_insert_after` properties
* Use new properties to determine whether to reveal drop between target
* Drop some unneeded code and variables
* Clarify meaning of `pinned` and `permanent`
*

PR #1863, opened 2021-10-17
check for drop below trash icon

a01wielo:
Here it is checked if the lower element is still of type BookmarkRow. 
If not, the drop is not allowed, so the error message does not appear anymore. 
So the trash can icon always remains the last bookmark in the list.

This is aimed at issue #1821


jeremypw:
Thanks for working on Files!  This issue might also be fixed in #1848, which reworks the drag and drop logic in the sidebar.

jeremypw:
Closing as #1848 fixes the issue and has been merged.  Thank you for the contribution though!

PR #1882, opened 2021-11-02
Focus more results

### LINKED ISSUE #1880: "more results" not keyboard accessible
When searching for files the bottom option to show more results can't be focused using the keyboard.

jeremypw:
Fixes #1880 

Remove checks preventing cursor landing on an ellipsis entry. There seems no good reason to prevent this - accepting such an entry either with mouse or keyboard repeats the search with greater depth and greater max results.

Directory view now ignores `cursor_changed ()` signal when on ellipsis.

We can, at the moment, assume that all category children are either a valid match (with an associated file) or an ellipsis so some checks can be lost.

Only clicks on matches or ellipses now calls the `accept ()` function so  reduce checking there.


danrabbit (approved):
Looks good, nice job simplifying!

PR #1889, opened 2021-11-09
.desktop: Add XDG dirs to actions

### LINKED ISSUE #1888: Add XDG dirs as .desktop actions
### Problem

I find myself hitting Super+Space and starting to type "Downloads" to open the downloads folder, which obviously doesn't do anything today.

### Proposal

We could add the standard XDG directories to the .desktop actions to have these in a quicklist and in search in the Applications Menu.

A potential disconnect I see is that these would _not_ mirror the user's added bookmarks in the sidebar, as it would be static. However, I think making it quicker to open the standard XDG dirs would still be a net benefit.

### Prior Art (Optional)

I believe we used to add bookmarks to the dynamic quicklist (e.g. using the Unity API), but that only worked when the app was open.

cassidyjames:
Screenshot from 2021-11-09 11 03 52

Fixes #1888. Open questions:

- [ ] Is this the right way to access XDG user dirs in a .desktop file?
- [ ] Is this the write naming for actions? I tried e.g. "Open Documents folder" but that looked really messy in the dock menu:

  Screenshot from 2021-11-09 10 56 12

- [x] Can/should we add Recent here?
- [x] Trash?


danrabbit:
Actually thinking about this for a second, I wonder if this should be a plugin for the applications menu so that we can get all the bookmarks and it's sorted into its own category instead of being mixed in with app actions, which this kind of isn't an app action. That would make it work regardless of your default file manager as well

cassidyjames:
@danrabbit yeah @jhaygood86 mentioned something similar on the issue report. These actions in this PR do use Files explicitly, though, and I do see value in having quick access from the dock, but I'm not really super hung up on it.

danrabbit:
Ah okay I think I might have misread it.

Generally I think a plug-in for applications menu might be more useful here, but I'm not gonna block an interim feature 🤷🏻‍♀️

cassidyjames:
Yeah having `@exec_name@` in there probably made it less clear it's still Files doing the opening. :smile:

cassidyjames:
@danrabbit thoughts on the order? Should the window actions stay up top, or [move to the bottom](https://twitter.com/Matt_Fantinel/status/1458138174329286657)?

danrabbit:
I would probably put the actual actions closer to the mouse

cassidyjames:
@danrabbit yeah that makes sense. Done.

Screenshot from 2021-11-09 11 28 05

jeremypw:
I do not see a lot of value in adding these actions to the dock but I agree it would be useful for them to come up when typing e.g. Downloads at the Applications Menu.

jhaygood86:
@jeremypw https://github.com/elementary/applications-menu/pull/514 for an implementation of that.

jeremypw:
@jhaygood86 I tend to prefer your approach (although it is more code).

cassidyjames:
Marking as draft since in its current state, this would conflict with the Applications Menu search.

jeremypw:
Closing as design conflict.

PR #1895, opened 2021-11-18
filechooser-portal: Check for flatpak-builder before providing a build target

tintou:
This allows to build without flatpak-builder in newer meson version as the command
line arguments were actually checked.


jeremypw (approved):
Makes sense

PR #1899, opened 2021-11-19
filechooser-portal: Rewrite Dialog over Gtk.FileChooserWidget

### LINKED ISSUE #1845: File Chooser Portal always spawns at 1,1 for maximized parents
## What Happened

If the parent of the file portal is a maximized window, the portal window always spawns at 1,1  (top left)

## Expected Behavior

We should never spawn new windows at 1,1 and instead center the portal window on the display if we can't center it on the parent

## Steps to Reproduce

1. Maximize a window (like web)
2. Active the file chooser (like attaching a file in GitHub)
3. Notice that the portal is at 1,1

### LINKED ISSUE #1865: "Save as" dialog doesn't offer 'New Folder' option
### What Happened?

_[I'm not sure if this falls under Files, or Browser, as it occurs across different browsers]_
Save As dialog doesn't offer a folder creation option, nor as a button, nor through a shortcut (`Ctrl+N`, or was it `Ctrl+Shift+N`) in default Web Browser, but also in Chrome and Brave

### Steps to Reproduce

1. Open any website in a web browser (built-in default, or Chrome, or Brave - happens in all)
2. Hit 'Save' (Ctrl+S)
3. Save dialog opens - offers location selection, but doesn't offer 'Create New Folder' option, nor does a shortcut work

View Issue

### Expected Behavior

In step 3, I expect Save As dialog to offer a New Folder button, and a shortcut to create a new folder at a selected location to work (as it used to work in Elementary OS 5.x).

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

### LINKED ISSUE #1878: Files Portal dialog should support file format filter from portal
### Problem

Currently, the Files Portal doesn't show the file format filter. 

### Proposal

Currently, the Files Portal should show the file format filter. The standard approach is a combobox with the filter name.

### Prior Art (Optional)

The GTK and KDE dialogs do

Marukesu:
Fixes #1878.
Fixes #1845 
Fixes #1865.

The old LegacyFileChooserDialog expected that the dialog used was already configurated, however, with portal we are the one configuring it. using a FileChooserWidget as base seems to be a better alternative in this case.

Besides the issues above, the rewrite also make the choices show on the UI, and added a option to open the files in read-only mode (similar to the Gtk Portal). it also fixed entering in search mode and the mixed mode implementation on the TreeView.

this PR also update the portal removing unneeded checks since the frontend already do them, and adding the "current_filter" key to the results HashTable on open_file and save_file.
Dialog in Save Mode
DIalog in Open Mode


filechooser-portal/Main.vala
```diff
@@ -19,84 +19,104 @@
...
-            multiple = multiple_variant.get_boolean ();
+        if ("multiple" in options) {
+            dialog.select_multiple = options["multiple"].get_boolean ();
         }
 
-        dialog.select_multiple = multiple;
-        unowned GLib.Variant? accept_label = options["accept_label"];
-        if (accept_label != null && accept_label.is_of_type (GLib.VariantType.STRING)) {
-            dialog.add_button (accept_label.get_string (), Gtk.ResponseType.OK);
+        if ("acccept_label" in options) {
```
jeremypw, line 64:
```suggestion
        if ("accept_label" in options) {
```

filechooser-portal/FileChooserDialog.vala
```diff
@@ -1,121 +1,441 @@
...
+
+        /* get a reference of the tree view, so we can grab focus later */
+        Gtk.Stack view_stack = find_child_by_name (revealer.parent, "list_and_preview_box/browse_files_stack");
+
+        tree_view = find_child_by_name (
+            view_stack.get_child_by_name ("list"),
+            "browse_files_swin/browse_files_tree_view"
+        );
+
+        /* remove extra unneded widgets */
```
jeremypw, line 382:
```suggestion
        /* remove extra unneeded widgets */
```

filechooser-portal/FileChooserDialog.vala
```diff
@@ -1,121 +1,441 @@
...
+        );
+
+        /* move the new folder button to HeaderBar and remove the chooser header */
+        Gtk.Stack stack = find_child_by_name (revealer.get_child (), "browse_header_stack");
+        Gtk.MenuButton new_folder_button = find_child_by_name (stack, "<GtkBox>/browse_new_folder_button");
+        new_folder_button.image = new Gtk.Image.from_icon_name ("folder-new", Gtk.IconSize.LARGE_TOOLBAR);
+        new_folder_button.parent.remove (new_folder_button);
+        header.pack_end (new_folder_button);
+
+        /* hide the revealer when not searching, for this work:
```
jeremypw, line 340:
```suggestion
        /* hide the revealer when not searching, for this to work:
```

jeremypw:
This seems to work pretty well!  One thing I noticed is that the "Save As" dialog (when launched from Code) appears in a different position and underneath the parent window compare to the the "Open" dialog (again launched from Code) which appears centered and above the parent.

Marukesu:
> This seems to work pretty well! One thing I noticed is that the "Save As" dialog (when launched from Code) appears in a different position and underneath the parent window compare to the the "Open" dialog (again launched from Code) which appears centered and above the parent.

That's a bug in Code. it isn't setting the "Save As" dialog as transient for the main window, with causes #1845.

danrabbit:
I can confirm that this branch fixes issues with parenting in Code and in Web, and probably other places

jeremypw:
There is a small visual issue - the "Modified" column header is sometimes slightly truncated:
FileChooserTruncatedHeader

Not necessarily blocking though as this is an important fix.

jeremypw:
There is also an issue with the pathbar not fully expanding so the icons in the first (and second) breadcrumb overlay the text.  I cannot take a screenshot of this as the act of taking the screenshot causes the pathbar to complete its animation.  Probably best to suppress the animation of the pathbar when the filechooser opens.

jeremypw:
Hopefully at some stage we can completely remove and replace all the widgets from the Gtk.FileChooser right? Having to search specific widget paths is not ideal.

jeremypw:
@danrabbit Do you think we should still have single-click to navigate behaviour when the purpose of the filechooser is set to "SELECT_FOLDER"?  I am not too sure about this.  Is it obvious that the opened folder is the one selected?

danrabbit:
@jeremypw It looks like the pathbar thing is more an issue with receiving focus. Once the window is focused things are calculated correctly, so probably not necessarily an issue with this branch, but making sure the dialog gets focus would be a better fix that disabling animations, imo

For the modified thing, I can only reproduce this when the folder the filechooser opens to is empty. I wonder if this is more of a treeview bug? I'm not sure how much power we have over that since presumably this view comes from Gtk.FileChooser

Yeah I would assume that the opened folder is the one that will be chosen, especially since there's no immediate selection in the list view. Probably best to be consistent here with how many clicks things take

tintou (approved):
The code looks good!

jeremypw (approved):
Lets go for it!

jeremypw:
@zappan Yes, I did notice that (see comment above) and meant to fix it before release but somehow it didn't happen. I'll fix as soon as I can.

zappan:
Ups, I haven't read that far, three posts up, sorry! 😳

jeremypw:
Thanks for the reminder anyway :smile:

jeremypw:
Fixed with #1931

PR #1902, opened 2021-11-23
Inactivate browser button if dragged.

### LINKED ISSUE #1898: If the arrow button is pressed and the window is dragged, the button will remain shaded.
### What Happened?

Hi! I found this little bug. 
If I press the arrow button and at the same time drag the window, the button will remain shaded until it is pressed again.

### Steps to Reproduce

https://user-images.githubusercontent.com/52709674/142517993-42d2d394-b853-4559-b81e-c527b480bd24.mp4

### Expected Behavior

The button should not remain shaded after being pressed.

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1898 

If the button is dragged it does not receive any button-release event and so does not go back to an inactive state.

As a conservative solution to the issue, the button becomes inactive as soon as the pointer moves with the button pressed.

Other solutions include not activating the button at all on button press (since it is a custom signal that actually causes navigation on button release) but this changes the UI.


libwidgets/Chrome/ButtonWithMenu.vala
```diff
@@ -136,10 +136,20 @@ namespace Files.View.Chrome {
...
             button_press_event.connect (on_button_press_event);
             button_release_event.connect (on_button_release_event);
+            motion_notify_event.connect (() => {
+                if (timeout > 0) {
+                    Source.remove (timeout);
+                    timeout = 0;
+                    active = false;
+                }
+
+                return false;
```
cassidyjames, line 151: Might make the intent more clear:

```suggestion
                return Gdk.EVENT_PROPAGATE;
```

cassidyjames (approved):

PR #1904, opened 2021-11-23
Add selection actions to context menu

### LINKED ISSUE #1903: Add select/deselect all to context menu
### Problem

Reading the massive thread at https://github.com/elementary/files/discussions/1729, I noticed a recurring theme of it being difficult for some to deselect items in certain views. At the same time, I realized Select All (Ctrl+A) action is not surfaced in the UI making it non-discoverable for anyone who doesn't already know the Ctrl+A shortcut.

### Proposal

We could encourage learnability for those using keyboards and usability for those who aren't by exposing "Select all" in the context menu of the view. I imagine these would make sense grouped in the menu with the existing clipboard actions of cut/copy/paste. If all the items are selected, it should change to "Deselect all".

### Prior Art (Optional)

We have a whole project board for [Keyboard Shortcut Discoverability](https://github.com/orgs/elementary/projects/32) with similar issues. :)

jeremypw:
Fixes #1903 

## With all selected:
DeselectAll (1)

## With some selected:
InvertSelection

## With none selected
SelectAll

At the moment it is not possible to show both "Select All" and "Deselect All" when only some are shown because both actions have the same accelerator in order to continue the existing toggle action of the accelerator.  Since the shortcut `<Ctrl>A` currently acts to select all items when only some are selected this is what is shown in the menu.


src/View/AbstractDirectoryView.vala
```diff
@@ -1977,7 +1989,36 @@ namespace Files {
...
+                select_all_menuitem.add (new Granite.AccelLabel.from_action_name (
+                    _("Select All"),
+                    select_all_menuitem.action_name
+                ));
+            } else {
+                unselect_all_menuitem = new Gtk.MenuItem () {
+                    action_name = "common.select-all"
+                };
+                unselect_all_menuitem.add (new Granite.AccelLabel.from_action_name (
+                    _("Unselect All"),
```
cassidyjames, line 2008: From my brief research, it seems "deselect" is the more common term in US English, at least. Not sure if you'd like to update the method/variable names as well, to match.

```suggestion
                    _("Deselect All"),
```

jeremypw: OK, I assumed US English preferred "Unselect" as that is what is used for Gtk and other library functions.  Probably would have translated it to "Deselect" for UK English anyway.

cassidyjames:
@jeremypw this is close but I think I need to discuss/define which items should be available when:

- Nothing selected:
  - Select All
- One or more items selected:
  - Select All
  - Invert Selection
  - Deselect All
- All items selected:
  - Deselect All

Right now it seems like "Deselect All" isn't showing up when I'd expect.

jeremypw:
@cassidyjames Yes, that is because there is no corresponding action at that point (without refactoring) because `<Ctrl>A` currently has a toggle action in master - it can either cause "Select all" or "Deselect all". `<Shift><Ctrl>A` is used for "Invert selection".  I note that Dolphin and Nautilus use `<Ctrl>A` for "Select All" and `<Shift><Ctrl>A` for "Deselect All" - they do not appear to offer an "invert selection" action.

Would you like me to refactor to use the same shortcuts as those other filemanagers?  It would actually simplify things a little,  What accelerator would be appropriate for "Invert Selection"?  `<Ctrl>I` seems to be available.

danrabbit:
@cassidyjames Can we get a follow up here on if this needs refactoring or it's enough of an improvement as is for just the current state of the world and maybe we can refactor later?

cassidyjames:
@danrabbit @jeremypw I'm mixed. This is probably an improvement over master, but I'm realizing the way the shortcuts work in master is pretty confusing. If zero or one thing is selected, Ctrl+A selects all, but if _all but_ one thing is selected, Ctrl+A deselcts all, but if everything is selected, Ctrl+A doesn't do anything… etc.

But yeah, getting the shortcuts in the menu sounds good to me and then we should follow up and sort out how those shortcuts actually work.

danrabbit (approved):
In that case, this works as expected for me. Nice way to surface these functions. Nice little actions cleanup with keypress handling. And yeah we can follow up on the shortcuts later and make it even cleaner :+1

PR #1910, opened 2021-11-26
Action match pointer

### LINKED ISSUE #1909: Click action does not always correspond to appearance of mouse pointer
### What Happened?

When hovering  over file items, the pointer changes appearance between a hand and an arrow.  It is intended that when the pointer shows a hand a click will navigate and when it shows an arrow it will select.  However, when clicking on the whitespace of the filename column of a folder the pointer is an arrow but a click nevertheless navigates.

### Steps to Reproduce

Open a folder in  any view
However over the whitespace after a folder's filename
The pointer shows an arrow but a click causes the folder to open

### Expected Behavior

The click action should be as indicated by the pointer.

### OS Version

6.x (Odin)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1909 

This makes IconView behave more similarly to TreeView and ensures that when the "activatable" cursor shows the item will activate (open) on single unmodified click, whereas when the default cursor shows it will be selected.   

The blank area of the filename column is now selectable/rubberbandable in both IconView and TreeView.

The actual text is activatable for folders in all views.

I was unsure whether the text should be activatable or selectable so UX team advise appreciated on that.

Annoying differences between the behaviour of the Gtk.IconView and Gtk.TreeView widgets made it necessary to select on button release when clicking on the blank are but this is hardly noticeable in practice.


danrabbit:
@jeremypw can you resolve conflicts here?

danrabbit:
I can't confirm that text is activatable on icon views with this branch

I think it makes sense that text is activatable. This is how we treat other labels attached to actionable widgets like Checkbutton

jeremypw:
It seems there is a vertical offset of about 1 fontheight between the tip of the cursor arrow/finger and the point at which the cursor changes from arrow to pointer so that, when you have a single line name,  the cursor can appear to be on the text but is still an arrow. This is less noticeable for multi-line names - it only affects the last line.  Probably why I didnt notice it.  I'll fix.

danrabbit (approved):
Works as expected. Nice job!

PR #1914, opened 2021-11-29
Sidebar.BookmarkRow: Explicitly highlight drop target

### LINKED ISSUE #1915: Drop targets in Sidebar not visibly highlighted
### What Happened?

Dragging a file item onto the sidebar and hovering over a destination folder shows an emblem on the drag-icon if the file can be accepted but this is difficult to see.  The entire listbox row should be highlighted but there is no visible change.

### Steps to Reproduce

1. Drag a file item from the view over the sidebar
2. Hover over a folder bookmark
3. Observe appearance of bookmark.

### Expected Behavior

When a folder bookmark will accept a dropped file then its appearance should clearly changed when hovered during drag and drop.

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

### LINKED ISSUE #1969: Highlight sidebar places while drag and drop.
### Problem

Small indicator icons is not enough while drag and drop to sidebar. (I added this line because of "Problem" box is mandatory.)

### Proposal

If I hover sidebar places (Downloads, Videos, etc.) with my cursor, it is highlighting. It's a good addition to have this same behavier while drag and drop.

### Prior Art (Optional)

Ekran görüntüsü 2021-12-31 09-39-33

jeremypw:
Fixes #1915 
Fixes #1969 

The bookmark row being hovered as a drop target does not have any visible highlight despite having Gtk.DEST_DEFAULT_HIGHLIGHT set on it.  Not sure why this is - stylesheet issue?

This PR adds an explicit way of changing the appearance of an acceptable drop target.  At the moment it just adds the Gtk.STYLE_CLASS_HIGHLIGHT but more customised changes could be added in the same way such as color change or emboldening the text.


danrabbit:
Alright, so it looks like there's a selector `row:drop(active)` that we can use to style this already. The only problem is that since each row currently contains a revealer with a separator to mark the drop area between rows, the drop style applies to the whole row:

Screenshot from 2022-04-08 14 16 26

So what we need to rectify styles here is probably to have the in-between drop revealers be their own rows in between each real row

jeremypw:
I'll see if having revealers as their own rows can be done - it will need a fair of refactoring.

jeremypw:
@danrabbit I have pushed a draft PR #2021 putting revealers in separate rows but a seemingly insuperable issue is that this causes the bookmarks to be spaced out compared with the storage and network lists.  The revealers have to have a finite height in order to be drop targets.  I have to hack around this but so far without success.

danrabbit:
Thanks for the quick follow up! Yeah this is a challenging problem

jeremypw:
Is this simple fix mergeable pending a more elegant solution?

jeremypw:
Or is there a CSS solution that does not involve refactoring?

danrabbit (approved):
This is definitely better than nothing! We can come back to it later to use built-in DND styles

PR #1928, opened 2021-12-02
Merge libwidgets and libcore.

### LINKED ISSUE #1656: Merge libcore and libwidgets
Do we really need two libraries ?

jeremypw:
Fixes #1656 

Widgets from libwidgets are moved into either `libcore/Widgets` or `src/Widgets` as appropriate, interfaces are moved into `libcore/Interfaces`. `Resources.vala` is moved into `libcore`.  To further facilitate development of the filechooser portal dialog, some Sidebar code is also moved from `/src` into `/libcore`.


danrabbit (dismissed):
Let's go :rocket:

Marukesu:
@jeremypw can we move the `SidebarListInterface` and `SidebarItemInterface` to libcore too? that would be needed for writing an files based sidebar for the FileChooser, so doing it now would be one step less after.

danrabbit:
@jeremypw before merging, I'd like to get an approve on https://github.com/elementary/files/pull/1930 so we can merge at the same time.

Do we also need to make any POTFILES changes here?

jeremypw:
@danrabbit Thanks for the reminder - I'm always forgeting POTFILES.  I have audited and updated it now.  There seem to be a lot of files in there which do not contain any translatable strings but I have left them for now.  I have taken the opportunity to tidy up the order of the files lists in POTFILES, `src/meson.build` and `libcore/meson.build` for easier comparison.

danrabbit (approved):

PR #1940, opened 2021-12-13
Conflict old libwidgets packages

danrabbit:


danrabbit:
Confirmed with Corentin in Slack

PR #1945, opened 2021-12-17
Remove deprecation: button.xalign; bump min gtk, disable deprecated

jeremypw:
First step towards GTK4 - remove remaining GTK3  deprecations.  Note that needs https://github.com/elementary/granite/pull/545 to suppress warning about a deprecation in DynamicNotebook

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -20,6 +20,8 @@ terminal_name = 'pantheon-terminal'
 add_project_arguments(
     ['-DGETTEXT_PACKAGE="' + meson.project_name() + '"',
     '-DHANDY_USE_UNSTABLE_API',
+    '-DGDK_DISABLE_DEPRECATED',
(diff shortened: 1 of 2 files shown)


danrabbit:
Hm, so I think the reason this was done was to make it so clicking anywhere along the actionbar will activate the button. If we want to preserve that behavior, it looks like according to the docs that the child of the button should be a Gtk.Alignment and we can set `xalign` there instead of on the button itself

jeremypw:
@danrabbit OK, I can do that.  I left it like this because Code does not hexpand the button - although in that case there is a menu button on the right.

jeremypw:
@danrabbit Slight problem - `Gtk.Alignment` is deprecated and does not appear in Gtk4. I'll see if there is another solution.

jeremypw:
Closing - fix xalign better in another PR.

PR #1946, opened 2021-12-17
Do not use deprecated style property

jeremypw:
Based on #1945
The style property "vertical-separator" is deprecated.  It is not clear how to obtain the same information otherwise so it is estimated at 10% of the cell height, which visually gives the same result.


jeremypw:
New PR based on master will be submitted.

PR #1949, opened 2021-12-18
Avoid using a style property

jeremypw:
In line with Gtk4 migration guidance, use of a style property is avoided.
It is not clear how to obtain the same information otherwise so it is estimated at 10% of the cell height, which visually gives the same result.


jeremypw:
Closing as a Gtk4 port is not imminent.

PR #1950, opened 2021-12-18
Avoid event properties

jeremypw:
Reduce work required for Gtk4 migration by by providing and using similar functions in EventUtils instead of using event properties.  For convenience the `get_coords` function returns `int` not `double` values.

TODO:  Find a way of not modifying event properties in IconView in order for rubberbanding to work more easily (larger target area).  This will not be possible in Gtk4.


jeremypw:
Closing as this was addressed in another PR.

PR #1951, opened 2021-12-18
Fix linear selection with mixed-mode

### LINKED ISSUE #1948: Selecting files with shift + arrow keys has imprecise behavior
### What Happened?

Selecting files with shift + arrow keys has imprecise behavior

### Steps to Reproduce

1. Select a file by clicking on it.
2. Press "shift" + "left or right" keys.
Result: Sometimes it does nothing. In others it selects more than one file at a time (sometimes it does it in reverse order).

### Expected Behavior

It should work fine.

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1948


danrabbit:
Is this still relevant now that clicking labels activates?

jeremypw:
Hmm, there is only one small material difference which makes the cursor follow the last selection.  Not sure how essential that was now.  I'll check.

jeremypw:
I can confirm the specific issue described at https://github.com/elementary/files/issues/1948#issuecomment-997227394 still exists in master (when clicking on the name to select) and this PR fixes it.

danrabbit:
Ah yep, I'm not sure what I missed when I was looking at this before, but I can confirm this fixes the issue in all views. Nice work!

danrabbit (approved):

PR #1962, opened 2021-12-27
Disable view actions when renaming

### LINKED ISSUE #1958: Ctrl + A after hitting rename does not select the text
### What Happened?

Ctrl + A after hitting rename does not select the text, but it selects all the folders and files in the directory. 

### Steps to Reproduce

1. Open File Manager.
2. Open any folder.
3. Try to rename a file. (Either by F2 or by Right Click > Rename)
4. Press Ctrl + A.

### Expected Behavior

All the text in the rename should get selected instead of all the folders.
 

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1958 

Most view actions, including changing the selected items are now disabled when renaming starts so that the entry actions work as expected.  Note that window actions are not disabled so it is still possible to e.g. navigate away from the view but no issues have yet been raised due to this.

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -2431,23 +2431,24 @@ namespace Files {
             can_open = can_open_file (file);
             can_show_properties = !(in_recent && more_than_one_selected);
 
-            action_set_enabled (common_actions, "paste-into", can_paste_into);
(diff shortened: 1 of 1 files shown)


danrabbit (approved):
Makes sense and I can confirm it fixes the issue

Issue #1960, opened 2021-12-26
Allow disabling single click to open on folder


undergroundwires:
### Problem

Single click on folder opens it.

`gsettings set io.elementary.files.preferences single-click false` does not work anymore. There seems to be no way of going around this while using elementary OS/files.

**Background**
Sometimes I need rename many folders one by one. This is simply too challenging with what current UX offers.

It is frustrating to click on a folder to rename, then go back, then rename it. Trying to click on the blue tick is not so productive when you need to be fast to get things done but also be pixel perfect.

### Proposal

Introduce a setting to disable single click to open folder.

jeremypw:
There are couple of discussions at https://github.com/elementary/files/discussions covering the disadvantages of having only one selection mode.  Certain operations need a different way of working but can still be done efficiently - if they cannot we will iterate on the selection mode until they can.  In this case it seems to be the problem is that after renaming the renamed item does not remain selected after naming - which I think is a regression.  When that is fixed, you can just press the right-arrow to move the selection to the next item and then F2 to rename it.  I agree you do not want to have to keep switching between mouse and keyboard.

jeremypw:
You could also consider installing a bulk renamer program if you are doing a lot of renaming. There is at least one free one in the AppCenter.  I started work on a bulk renamer plugin for Files (see https://github.com/jeremypw/bulk-renamer-plugin) but haven't got round to submitting it yet.

jeremypw:
Another possibility would be for Files to modify the rename mode when there is a multiple selection so that after renaming the first item the next selected item enters rename mode.

rdanter:
I know it was a design decision to remove the option to have double-clicking to open files/folders, but for someone who often switches between eOS, macOS and various other Linux distros (for both work and personal), I find it really frustrating to have just eOS work this way.

I don't have a problem with this being the default, but for usability it would be really nice to have it optional.

jeremypw:
I understand but it is unlikely that design decision will be reversed any time soon. You could consider using list view as standard as the area where folders can be selected with a single (unmodified) click is much larger than in grid view (just avoid clicking on the filename column).  It would be possible to increase the size of the helper to make it much easier to select folders I guess.  I'll push a PR some time and see what the design team think.

I am not sure what macOS does with modified clicks but you might be able to use `<Alt>Click` or `<Ctrl>Click` on both to select folders.

rdanter:
> I understand but it is unlikely that design decision will be reversed any time soon.

[...]

PR #1964, opened 2021-12-28
Implement sequential manual renaming of a selection of files

jeremypw:
Suggested by #1960 

The linked issue complained of difficulty experienced renaming several files as a result of the change to a single mixed-mode selection model.  

This PR is an attempt to provide a reasonably efficient way of manually renaming several files in sequence without having to start renaming over again for each file. 

It is assumed that the number of files selected is modest and that the new names do not follow a pattern that could be better applied with a bulk renamer app or plugin.

At present, only pressing `Esc` cancels the entire renaming sequence. Ending renaming of an item any other way just starts renaming the next selected item.


undergroundwires:
This is an elegant solution to the problem. Very minimal but very useful solution boosting the productivity. It's not bloating anything and very intrusive almost an easter-egg level feature. Is there anything blocker to it? Can we see it merged? Or get some reasoning why not?

PR #1965, opened 2021-12-28
Cancel current selection on button press to block dragging text

### LINKED ISSUE #1957: Allow starting a new selection when clicking on and dragging a selected text
### What Happened?

Files tries to move the folder.

### Steps to Reproduce

1. Choose a folder
2. F2 to rename
3. Hold down left mouse button to start selection action

### Expected Behavior

Start selecting text from where I start

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1957 

This overrides the native behaviour of the widgets used for renaming so that instead of dragging selected text, a new selection is started.  This emulates the behaviour of the Nemo filemanager (and contrasts with Nautilus and Dolphin)

### DIFF
--- a/src/View/Widgets/MultiLineEditableLabel.vala
+++ b/src/View/Widgets/MultiLineEditableLabel.vala
@@ -26,9 +26,16 @@ namespace Files {
 
         public override Gtk.Widget create_editable_widget () {
             textview = new Gtk.TextView ();
+            /* Cancel selection under cursor to avoid dragging selected text */
+            textview.button_press_event.connect (() => {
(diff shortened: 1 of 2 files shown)


jeremypw:
Blocking pending merging of #2030 since the issue only occurs when text is pre-selected.

danrabbit:
I'm -1 on this since it would make this one particular text selection in Files inconsistent with the behavior of every other text selection in every other app

jeremypw:
OK, no problem.  Should the issue be marked "Design Conflict"?

PR #1966, opened 2021-12-30
FileOperations: Move the datastructures to Vala

tintou:
Allows a smoother transition to Vala of the last remaining piece of C.


libcore/FileOperations/CopyMoveJob.vala
```diff
@@ -0,0 +1,82 @@
...
+        undo_redo_data = new UndoActionData (Files.UndoActionType.COPY, (int) this.files.length ());
+        undo_redo_data.src_dir = this.files.nth_data (0).get_parent ();
+        undo_redo_data.dest_dir = target_dir;
+    }
+
+    public CopyMoveJob.move (Gtk.Window? parent_window = null, owned GLib.List<GLib.File> files, GLib.File target_dir) {
+        base (parent_window);
+        this.files = (owned) files;
+        this.destination = target_dir;
+        debuting_files = new GLib.HashTable<GLib.File, unowned GLib.File> (GLib.File.hash, GLib.File.equal);
```
jeremypw, line 45: I do not think `debuting_files` has any actual function anymore.  Need to look into whether we need to port it.

tintou: Indeed, by looking at the C code, the `debuting_files` field is only  populated but never read, I've removed it

jeremypw (changes requested):
A couple of regressions/behaviour changes noticed so far:

1. Dragging and dropping a file into the same folder does not create a link as before, but either gives the conflict dialog or crashes.  (However, copy + paste into same folder works as expected and creates a copy)
2. Restore selected item from trash restores the item, but also leaves it in the trash (i.e. copies instead of moves)

Will continue testing.

jeremypw:
@tintou Is this PR still relevant in view of your other recent merges?

PR #1970, opened 2021-12-31
Do not activate subfolder if double-click on folder

### LINKED ISSUE #1968: Double-clicking a folder may open a subfolder of the clicked folder
### What Happened?

Double clicking a folder resulted in FIles opening a subfolder of the clicked folder rather than the folder itself.

### Steps to Reproduce

 1. Open a folder that has as its first item a folder that itself contains subfolders
 2. Double click on the first item

### Expected Behavior

The folder double clicked on should open.

### OS Version

6.x (Odin)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1968 

Not an elegant solution as it uses a timeout and another boolean and introduces a slight delay before the folder opens.  However, no other solution was obvious and ColumnView already uses a similar method to detect double-clicks on folders.

Need opinion on whether the issue is serious enough to need fixing.


ChildishGiant:
Mixed opinions on this since ideally nobody would be double clicking a folder but equally when I double click an element I expect it to effect that one, not another. I think as long as properly implemented a fix would be good

jeremypw:
This is really for the benefit of people frequently swapping between machines with a different OS that uses double-click to activate. They could doubly click on a folder out of habit.

danrabbit:
Hrm, I'm really not a fan of adding a timeout here. This seems like designing for the minority case at the expense of the majority case

jeremypw:
OK, no problem.  Should the issue be marked "Design conflict"?

PR #1972, opened 2022-01-06
Fix Public Share symbolic icon missing in breadcrumb

ETDV-TheVoid:
Public Share is the only default User Directory to not have its symbolic icon displayed in the breadcrumb. This fixes that issue.


jeremypw (approved):
Thanks for providing that overlooked icon!  Works as expected.

PR #1973, opened 2022-01-07
Meson: remove positional arguments from i18n.merge_file

Xeonacid:
`i18n.merge_file` has been ignoring positional arguments for a time and explicitly rejects with error since meson 0.60.0.

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -49,7 +49,6 @@ appdata_in = configure_file(
 )
 
 i18n.merge_file(
-    'policy',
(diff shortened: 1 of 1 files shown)


jeremypw:
https://github.com/mesonbuild/meson/issues/9441 explains the situation.
This looks good to me.

jeremypw:
Confirmed that this builds and installs with Meson 0.53

jeremypw (approved):

PR #1974, opened 2022-01-07
Bulk renamer plugin

### LINKED ISSUE #1040: Bulk rename
Hello there,
I was wondering if there was any work in progress for implementing bulk rename functionality.

It would be nice to have it.

Kind regards, and Keep up the good work

jeremypw:
Fixes #1040 

Functional but needs UX of dialog sorting out.  Suggestions welcome.


plugins/bulk-renamer/RenamerDialog.vala
```diff
@@ -0,0 +1,171 @@
...
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ *  Authors:
+ *  Vartan Belavejian <https://github.com/VartanBelavejian>
+ *  Jeremy Wootten <jeremywootten@gmail.com>
+ *
+*/
+
+public class Files.RenamerDialog : Gtk.Dialog {
```
danrabbit, line 24: Using Granite.Dialog here would save some styling work

plugins/bulk-renamer/RenamerDialog.vala
```diff
@@ -0,0 +1,171 @@
...
+    private Files.Renamer renamer;
+    private Gtk.Widget rename_button;
+    public RenamerDialog (string _basename, GLib.File[] _files) {
+        renamer.add_files (_files);
+        renamer.set_base_name (_basename);
+    }
+
+    construct {
+        deletable = true;
```
danrabbit, line 33: Dialogs shouldn't have close window controls since it's ambiguous what happens when the window is closed. 
```suggestion
```

plugins/bulk-renamer/RenamerDialog.vala
```diff
@@ -0,0 +1,171 @@
...
+                    break;
+            }
+
+
+            return false;
+        });
+
+        realize.connect (() => {
+            resize (500, 300);  //Stops the window being larger than necessary
+        });
```
danrabbit, line 102: Probably should find the actual cause of this, this seems pretty janky

plugins/bulk-renamer/RenamerDialog.vala
```diff
@@ -0,0 +1,171 @@
...
+
+            switch (keyval) {
+                case Gdk.Key.Escape:
+                    if (mods == 0) {
+                        response (Gtk.ResponseType.REJECT);
+                    }
+                    break;
+                case Gdk.Key.Return:
+                    if (mods == 0 && renamer.can_rename) {
+                        response (Gtk.ResponseType.APPLY);
```
danrabbit, line 90: This definitely isn't the right way to do this. the "Rename" button should have `has_default = true` and entry widgets in the dialog should have `activates_default = true`

plugins/bulk-renamer/BulkRenamer.vala
```diff
@@ -0,0 +1,568 @@
...
+        can_rename = false;
+        orientation = Gtk.Orientation.VERTICAL;
+        directory = "";
+
+        file_map = new Gee.HashMap<string, GLib.File> ();
+        file_info_map = new Gee.HashMap<string, FileInfo> ();
+        modifier_chain = new Gee.ArrayList<Modifier> ();
+
+        var base_name_label = new Granite.HeaderLabel (_("Base"));
+        base_name_label.get_style_context ().add_class (Granite.STYLE_CLASS_H2_LABEL);
```
danrabbit, line 69: Making these header labels this big seems pretty excessive

plugins/bulk-renamer/BulkRenamer.vala
```diff
@@ -0,0 +1,568 @@
...
+
+        base_name_combo = new Gtk.ComboBoxText () {
+            valign = Gtk.Align.CENTER
+        };
+        base_name_combo.insert (RenameBase.ORIGINAL, "ORIGINAL", RenameBase.ORIGINAL.to_string ());
+        base_name_combo.insert (RenameBase.CUSTOM, "CUSTOM", RenameBase.CUSTOM.to_string ());
+
+        base_name_entry = new Gtk.Entry () {
+            placeholder_text = _("Enter naming scheme"),
```
danrabbit, line 78: Placeholder text should be an example of valid text for this entry. Descriptions of entries should be labels outside the entry. I'm not sure what good example text here would be, so we should probably remove this

plugins/bulk-renamer/BulkRenamer.vala
```diff
@@ -0,0 +1,568 @@
...
+            hexpand = false,
+            max_width_chars = 64,
```
danrabbit, line 80: Is there a reason for restricting the character limit here?

plugins/bulk-renamer/BulkRenamer.vala
```diff
@@ -0,0 +1,568 @@
...
+
+        base_name_combo = new Gtk.ComboBoxText () {
+            valign = Gtk.Align.CENTER
+        };
+        base_name_combo.insert (RenameBase.ORIGINAL, "ORIGINAL", RenameBase.ORIGINAL.to_string ());
+        base_name_combo.insert (RenameBase.CUSTOM, "CUSTOM", RenameBase.CUSTOM.to_string ());
+
+        base_name_entry = new Gtk.Entry () {
+            placeholder_text = _("Enter naming scheme"),
+            hexpand = false,
```
danrabbit, line 79: expand is `false` by default, so we don't need to specify it
```suggestion
```

plugins/bulk-renamer/BulkRenamer.vala
```diff
@@ -0,0 +1,568 @@
...
+
+        var sort_type_grid = new Gtk.Grid () {
+            orientation = Gtk.Orientation.HORIZONTAL,
+            column_spacing = 6,
+            halign = Gtk.Align.END,
+            valign = Gtk.Align.CENTER,
+            margin = 3
+        };
+        sort_type_grid.add (sort_type_switch);
+        sort_type_grid.add (sort_type_label);
```
danrabbit, line 123: This seems kind of excessive. Do we really need sort controls here?

danrabbit (changes requested):
Wow this is a lot! First thing, instead of a separate plugin/menuitem, this should probably just be the behavior for the `Rename` menuitem when multiple files are selected

I kind of feel like this is probably overwhelming for most people to the point of not longer being useful. Perhaps we should scope down here to just simple renaming schemes and leave some of the more advanced stuff for specialized apps if you really have a wild use case? I'd guess the vast majority of bulk renaming operations would be to suffix with either numbers or a date, but probably not multiple combinations of prefixes, suffixes, and dates.

The GNOME approach here is also interesting. It uses a single entry where you can add items from a menu to make a regex-like template:

nautilus-batch-rename

This seems like it would allow for really flexible renaming while keeping things simple for someone to specify a format like `[Original file name]-[1, 2, 3]`.

jeremypw:
Thanks for the review.  This was originally written as a separate app but I never got round to submitting it to AppCenter as I had difficulty coming up with a suitable icon/name.  I submitted it here in response to renewed interest in the related issue.

I'll have a think about simplifying the UX so only the most commonly used options appear initially and perhaps hide less used options under an "Advanced..." button or something.  The screenshot does show more modifiers than most people would use just to illustrate that you can chain an arbitrary number of them.  The dialog does need to be made smaller.

Regarding having only one option "Rename",  I have also submitted an alternative way the existing option could handle a multiple selection, i.e. by sequentially renaming them arbitrarily (which this PR cannot do at present).  I guess in most cases where a user wants to rename multiple files it is in order to apply some pattern though.

jeremypw:
The code is quite old so definitely needs updating to best practices!

jeremypw:
I can drop the sorting of the original files and have it follow the current sort order of the view.

jeremypw:
Closed in favour of #1999, which is not a plugin and has a simpler interface.

PR #1976, opened 2022-01-08
Reference bookmark when iterating list to avoid destruction

### LINKED ISSUE #1975: Dragging to reorder bookmarks causes Files to crash and some bookmarks to be removed from the list.
### What Happened?

Hi!

Dragging to reorder bookmarks causes Files to crash and some bookmarks to be removed from the list.

### Steps to Reproduce

https://user-images.githubusercontent.com/52709674/148648907-9da48dc0-5b11-4162-a4e9-5f628c98242f.mp4

### Expected Behavior

Files shouldn't crash and bookmarks shouldn't be removed when doing this.

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1975 

Should not use unowned reference when removing bookmark from list - causes destruction of bookmark object and to a  bookmark row with null bookmark, leading to a crash if moved again.

### DIFF
--- a/libcore/BookmarkList.vala
+++ b/libcore/BookmarkList.vala
@@ -177,7 +177,7 @@ namespace Files {
          * only permits one bookmark per uri.
          */
         public void rename_item_with_uri (string uri, string new_name) {
-            foreach (var bookmark in list) {
+            foreach (unowned var bookmark in list) {
                 if (uri == bookmark.uri) {
                     if (new_name == Path.get_basename (uri)) {
                         // Do not use basename as a custom name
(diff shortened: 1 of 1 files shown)


libcore/BookmarkList.vala
```diff
@@ -177,14 +177,15 @@ namespace Files {
          * only permits one bookmark per uri.
          */
         public void rename_item_with_uri (string uri, string new_name) {
-            foreach (unowned Files.Bookmark bookmark in list) {
+            foreach (var bookmark in list) {
```
tintou, line 180: there is no real reason to use a full reference here though

jeremypw: There is no significant cost though is there?  I did this just to be on the safe side ...

jeremypw: Reverted to unowned var

tintou (approved):

PR #1978, opened 2022-01-09
Extend File.compare_by_size to use folder item count where appropriate

### LINKED ISSUE #1971: Sort by size has no effect on folders.
### What Happened?

When I sort by size this has no effect on folders.

### Steps to Reproduce

Screenshot from 2022-01-03 12 27 46

### Expected Behavior

Files should be able to sort folders by size, just like other file managers.

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

### LINKED ISSUE #2108: Files no longer shows size of folders in list view
### What Happened?

In list view, Files used to show the size of folders as either "N items" or "Empty". Around the time of the update to Files mid-November, it seems to no longer display the size of folders in list mode. Instead it shows "----" for all of the folders.

### Steps to Reproduce

1. Apply all updates
2. Open Files and change to "View in list" mode
3. See what it says under the size column for any folder

### Expected Behavior

Files used to show the number of items in the folder, or "Empty".

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1971 
Fixes #2108


jeremypw:
Blocking PR was closed without merging.

jeremypw:
Its hard to see how this PR, which does not touch code related to icons causes a failure in the IconInfo test suite :disappointed: (after merging master).

zeebok (dismissed):
Code looks good. Seems straightforward enough to not test but if you'd like me to, let me know.

jeremypw:
I would consider this complex enough to require at least some testing especially as it was written a while ago and has had master merged several times.  I'll try it out again with some tests from the wiki.

jeremypw:
One bug I notice in this is that the item count does not update when files are added (or removed) externally until the view is refreshed.  I guess I should fix that.

jeremypw:
I don't think there is a simple way of getting the number of items to update automatically - to do that we would have to put a monitor on every folder item.

jeremypw:
I have confirmed Nautilus has the same issue.

jeremypw:
Another slight quirk is that if a folder contains e.g. just one hidden file then the item count is "1" regardless of the "Show hidden" setting.  So opening such a folder can result in the message "Empty" even though the item count is >0.  However Nautilus behaves the same way.

jeremypw:
Now only visible items are counted matching Nautilus's behaviour (without the bug).

jeremypw:
@zeebok I have made some changes to this PR.  I hope you do not mind having another look :smile:

danirabbit (commented):
Just one questions from me :)

libcore/File.vala
```diff
@@ -1094,36 +1095,60 @@ public class Files.File : GLib.Object {
...
-                }
+    private string format_item_count () {
+        ensure_item_count (true); // Re-count file items
+
+        if (count < 0) {
+            return (_("—"));
+        } else if (count == 0) {
+            return _("Empty");
+        } else {
+            return ngettext ("%'d item", "%'d items", count).printf (count);
```
zeebok, line 1116: Does this text need to be translated? I am not too familiar with how ngettext works with it.

jeremypw: `ngettext` does the translation.

libcore/File.vala
```diff
@@ -1094,36 +1095,60 @@ public class Files.File : GLib.Object {
...
-                var f_enum = location.enumerate_children ("", FileQueryInfoFlags.NONE, null);
-                var count = 0;
-                while (f_enum.next_file () != null) {
-                    count++;
-                }
+    private string format_item_count () {
+        ensure_item_count (true); // Re-count file items
+
+        if (count < 0) {
+            return (_("—"));
```
ryonakano, line 1112: Does this string really need to be translatable?

libcore/File.vala
```diff
@@ -1094,36 +1095,60 @@ public class Files.File : GLib.Object {
...
+                    }
                 }
             } catch (Error e) {
-                return _("Inaccessible");
+                count = -1;
             }
         }
 
-        return _("----");
+
```
ryonakano, line 1151:
```suggestion
```

Extra newline here.

ryonakano:
Also, it feels strange for me that the item count won't be updated on file creation/deletion inside a folder unless we press Ctrl + R to trigger refresh.

libcore/File.vala
jeremypw: I'll take your advice on this.  Does that character have the required meaning in all scripts?

jeremypw:
> Also, it feels strange for me that the item count won't be updated on file creation/deletion inside a folder unless we press Ctrl + R to trigger refresh.

I can probably get the count to update when the file operation is internal but it still will not if the operation is external.

jeremypw:
After the last commit, the count updates after an internal change.  It will also update after an external change if the changed folder is open in a tab, window or expanded in list view. It will not update if only being shown as an unexpanded item in a view because there is no Directory object to monitor for changes.

jeremypw:
We should perhaps consider whether we need to throttle the count updates when adding/removing large numbers of files in one operation else it will slow things up.

jeremypw:
Converting to draft while investigating performance

jeremypw:
The latest commits improve performance by updating the count only once per batch of file changes.  This assumes that each batch refer to the same folder - this needs testing thoroughly.  It is certainly true for internally generated changes as the user can only interact with one background folder at a time but may not be true for external changes.  Performance is also improved by only signalling the model(s) to add a file if it is not already in the directory file hash.

Also, folder item counts now update their count even if there is no Directory object for them. The Directory handling the change now also checks for a Directory object for the grandparent of the file and if it exists notifies that that its item count should update.

jeremypw:
TESTING:  Try to generate a set of file changes where the files in the list do not have the same parent folder.

zeebok (approved):
This seems to work when I did my testing and the code seems reasonable

Issue #1956, opened 2021-12-24
Right clicking on new tab icon could provide an option to open new window


heidefinnischen:
### Problem

Up until now to open a new window I would either right click on a random folder or right click on the files icon in the dock. Another option would be to create a new tab and then move it out of the window.

### Proposal

I think it would be great if you could just right click the tab "plus" icon as it would make sense and it's easier to do than having to move the mouse all the way to the dock for a clean new instance. Its a bit as if you're creating a new tab and moving into a new window. A feature for the power users if you would say so.

jeremypw:
Note that `<Ctrl>N` opens a new window.  For some reason this shortcut is not yet exposed in the menus.

jeremypw:
This would need the Granite.Widgets.DynamicNotebook to be amended - the context menuitems for the tab bar are hard wired (except for the accelerators) so this issue cannot be solved by Files.

Note that you can get the equivalent effect by right clicking on the "Home" bookmark and choosing "Open in New Window"

Closing as "Out of Scope".

PR #1980, opened 2022-01-09
Show New Tab and New Window accelerators

jeremypw:
Suggested by #1956 

 * Add new tab and new window window actions with accelerators
 * Show accelerators in tab menu
 * Also amend view context menu to show accels and use window actions
 * Add view accelerator for open selected in new window

Also fixes an issue where the "Open in New Tab" view context menu for the background folder showed an inactive accelerator.


danrabbit (approved):
Nice work!

PR #1981, opened 2022-01-10
Double click option

### LINKED ISSUE #1960: Allow disabling single click to open on folder

jeremypw:
Fixes #1960 

Providing this option only requires relatively trivial changes to the click handling code.  Most of the diff is providing the UX to change the option the remember it state.

For simplicity the option accompanies other similar settings toggles in the view context menu, although it would be perhaps better placed in an AppMenu.

It is not necessary to install the PR to test the UX and functionality but in that case the setting will not be persisted.


ibrahimrahman:
Any possibility to have this at least reviewed before OS7?

JoseExposito (approved):
Thanks for this change @jeremypw, the code looks good, and after a quick test it works as expected.

There is a little conflict with `master` though.

I'm absolutely in favor of adding this option, but I know some people preferred to not include it. It'd be nice to have more feedback about it.

jeremypw:
@JoseExposito Thanks for the review! What was the conflict with master you mentioned? Github is not showing anything.  I'll not merge this before more comment as it is likely to be controversial.

JoseExposito:
@jeremypw I got a conflict rebasing on top of master, but it looks like it was on my end, GitHub didn't have any issues merging master into you branch.

> I'll not merge this before more comment as it is likely to be controversial.

Agrees. I truly think that it is great to provide this option, but it could be controversial.

zeebok:
I think if it is an option, I see no issue with it being added. For me the vision of the project should be the defaults, not always the *only* way. Particularly something like this that can really mess with people's workflow.

PR #1982, opened 2022-01-11
Rewrite ViewSwitcher.vala without Granite.ModeButton

jeremypw:
Suggested by https://github.com/elementary/granite/pull/558

Some simplifications were made, including removing the ability to switch by scrolling on the buttons.
Scrolling caused a bug requiring over-rapid switching to be suppressed, removing it enabled removal of throttling as it seems unlikely that operation by clicking could trigger the bug.  If necessary throttling can be restored.  It is still possible to crash Files by rapidly pressing the view mode accelerator keys but that bug pre-exists in Files so will be dealt with in another PR.

### DIFF
--- a/libcore/Widgets/ViewSwitcher.vala
+++ b/libcore/Widgets/ViewSwitcher.vala
@@ -22,56 +22,79 @@
 ***/
 
 namespace Files.View.Chrome {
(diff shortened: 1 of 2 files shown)


danrabbit (approved):
Looks good :+1:

PR #1989, opened 2022-01-22
Add dialog to ask for overwrite

### LINKED ISSUE #1955: FileChooser portal doesn't prompt on overwriting
### What Happened?

If you're saving a file with the FileChooser portal and give it the same name as an existing file, the portal happily overwrites the existing file without warning.

### Steps to Reproduce

1. Open Code and save a new file `foo.txt`
2. Close the file, then create a new, different file
3. Save the new file as `foo.txt` in the same folder

### Expected Behavior

the FileChooser should not allow the same name, or more likely should throw a MessageDialog asking if you want to overwrite the existing file.

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

mcclurgm:
Fixes #1955

Adds a new second-layer modal dialog that asks whether to overwrite a file when the name clashes with an existing file.

Bildschirmfoto von 2022-01-21 16 14 05

Known issues:

- [ ] When the selected file is renamed, deleted, or otherwise removed from the current directory, the dialog goes to the next selectable file and picks that. We need to preserve the user's filename choice in such cases, so the saved file remains consistent.
    - To demonstrate:
        1. Open a "Save As" dialog box
        2. Select an existing file to overwrite in a folder with multiple files (make sure you don't care about any of them).
        3. The conflict dialog will appear. Do not take action.
        4. In the file manager, delete the file that you selected.
        5. Back in the save dialog, the selection will move to a different file.
        6. Click the "Replace" button. The newly selected file, not the file that you originally picked or was displayed in the conflict dialog, will be overwritten.
    - Expected behavior: if the chosen file is removed, a new file should be saved with the name of the file originally chosen. (This may be up for debate, but that's what I expect.)
    - [ ] Related: if the file is deleted, is there any reason to change the behavior? I don't imagine we should dismiss the conflict dialog automatically, and I can't think of any other behavior change to make.
- [ ] When the dialog is closed by confirming and saving the file, the parent window remains dimmed.
- [ ] The save dialog is not dimmed when the overwrite dialog is presented.

Things to check:

- [ ] I'm not sure if the way I'm checking for existing files is preferable. The docs seem to suggest you don't use file tests for this functionality, but I'm not sure what else to do.
- [ ] Better wording could be nice. I did my best to explain (largely taken from Apple's built in wording), but there's always room for improvement.
- [ ] It does not handle filed URI→filename conversions. I'm not sure what situation would cause them, so I don't know what would be a sensible fallback.
- [ ] Do I need to handle different file types, like symlinks, differently? If you choose a symlink then the linked file would be overwritten, not the link file. I'm not sure if I should use different wording in that case.
- [ ] I'm not sure if it's possible to select multiple files in a replace dialog. If so, then that would almost certainly need special handling.
- [ ] Likewise, I have not checked the behavior with other FileActions, like creating a new folder. I assume no selection actions would be able to trigger an overwrite.

Any other testing, code, or design input would be hugely appreciated.

### DIFF
--- a/filechooser-portal/FileChooserDialog.vala
+++ b/filechooser-portal/FileChooserDialog.vala
@@ -489,7 +489,7 @@ public class Files.FileChooserDialog : Hdy.Window, Xdp.Request {
 
             if (GLib.FileUtils.test (filename, GLib.FileTest.EXISTS) && !GLib.FileUtils.test (filename, GLib.FileTest.IS_SYMLINK)) {
                 var display_filename = GLib.Filename.display_basename (filename);
-                var primary = _("Replace \"%s\"?".printf (display_filename));
+                var primary = _("Replace “%s”?".printf (display_filename));
                 var secondary = _("The file already exists. Replacing it will permanently overwrite its current contents.");
                 var replace_dialog = new Granite.MessageDialog.with_image_from_icon_name (primary, secondary, "dialog-warning", Gtk.ButtonsType.CANCEL) {
                     transient_for = this
--- a/filechooser-portal/FileChooserDialog.vala
+++ b/filechooser-portal/FileChooserDialog.vala
@@ -272,12 +272,12 @@ public class Files.FileChooserDialog : Hdy.Window, Xdp.Request {
 
         chooser.file_activated.connect (() => {
              if (!GLib.FileUtils.test (chooser.get_filename (), FileTest.IS_DIR)) {
-                 response (Gtk.ResponseType.OK);
+                 accept ();
              }
         });
 
         cancel_button.clicked.connect (() => response (Gtk.ResponseType.CANCEL));
-        accept_button.clicked.connect (() => response (Gtk.ResponseType.OK));
+        accept_button.clicked.connect (accept);
 
         // save the dialog size and close after selection
         response.connect_after (() => {
@@ -381,7 +381,7 @@ public class Files.FileChooserDialog : Hdy.Window, Xdp.Request {
             entry.bind_property ("text-length", accept_button, "sensitive", BindingFlags.SYNC_CREATE);
             entry.activate.connect (() => {
                 if (accept_button.sensitive) {
-                    response (Gtk.ResponseType.OK);
+                    accept ();
                 }
             });
 
@@ -469,6 +469,45 @@ public class Files.FileChooserDialog : Hdy.Window, Xdp.Request {
         if (chooser.list_filters ().search<string> (name, (a, b) => strcmp (a.get_filter_name (), b)) == null) {
             chooser.add_filter (filter);
             filter_box.append (name, name);
+        }
+    }
+
+    private void accept () {
+        if (action == Gtk.FileChooserAction.SAVE || action == Gtk.FileChooserAction.CREATE_FOLDER) {
+            // If an existing file would be overwritten, ask for permission first
+
+            assert (select_multiple == false);
+
+            // TODO handle failed URI conversions
+            var filename = "";
+            try {
+                filename = GLib.Filename.from_uri (get_uri ());
+            } catch (Error e) {
+                warning ("Could not convert URI of selected item to filename");
+                warning (e.message);
+            }
+
+            if (GLib.FileUtils.test (filename, GLib.FileTest.EXISTS) && !GLib.FileUtils.test (filename, GLib.FileTest.IS_SYMLINK)) {
+                var display_filename = GLib.Filename.display_basename (filename);
+                var primary = _("Replace \"%s\"?".printf (display_filename));
+                var secondary = _("The file already exists. Replacing it will permanently overwrite its current contents.");
+                var replace_dialog = new Granite.MessageDialog.with_image_from_icon_name (primary, secondary, "dialog-warning", Gtk.ButtonsType.CANCEL) {
+                    transient_for = this
+                };
+                var replace_button = replace_dialog.add_button ("Replace", Gtk.ResponseType.YES);
+                replace_button.get_style_context ().add_class (Gtk.STYLE_CLASS_DESTRUCTIVE_ACTION);
+
+                var replace_response = replace_dialog.run ();
+                if (replace_response == Gtk.ResponseType.YES) {
+                    response (Gtk.ResponseType.OK);
+                }
+                replace_dialog.destroy ();
+            } else {
+                response (Gtk.ResponseType.OK);
+            }
+        } else {
+            // Just selecting an existing item to open: no need to check for overwrite
+            response (Gtk.ResponseType.OK);
         }
     }
 


jeremypw:
I havent tested yet but presumably this is intended to replace any dialog the calling app throws up when a file conflict occurs? Otherwise the user will be presented with two dialogs if the user confirms overwriting is intended.

jeremypw:
I see that Gtk.FileChooser interface has "do_overwrite_confirmation" property to deal with this but does the freedesktop portal interface support this?

jeremypw:
The documentation for the portal linked to above says
 > Supported keys in the options vardict include:

Does this mean other options can also be supported - in which case we need to support and honour the "do-overwrite-confirmation" property in the portal and default that to "false".

mcclurgm:
Hmm. I don't see any such option in the freedesktop portal, do you mean that it could be possible to pass it in the dictionary anyway out of spec? If so, then do we need to reimplement the whole GtkFileChooser overwrite interface?

jeremypw:
Using the word "include" does not exclude the possibility of other options - to do that you would say "The supported keys in the options vardict are:".  However it is possible that it is an unintentional miswording.

Would need to add some code to `filechooser-portal/Main.vala` like 
```
if ("do-overwrite-confirmation" in options) {
    dialog.do_overwrite_confirmation = options["do-overwrite-confirmation"].get_boolean ();
}
```
and add  `do_overwrite_confirmation` as a property of the Files FileChooserDialog implementation (defaulting to `false`), I would guess.

mcclurgm:
Yeah, adding a simple key like that would be relatively easy. I guess I may be misunderstanding. I was referring to the `Gtk.FileChooser` interface, which is much more complex, involving signal callbacks. I don't understand the API from the docs, but it also looks like something that would not be possible to pass through the fd.o portal?

So are you thinking of just including a single key that turns on/off the overwrite dialog, or should we work on reaching more feature completeness with the Gtk interface? The second would be much harder.

jeremypw:
I am assuming that the portal interface will expect a dictionary Variant for the options but not be concerned with what it contains.  I would be very wary of changing the spec of the portal.  Just need to try it with one extra boolean option first.

Marukesu:
about the code, IMHO, it's better to do the check inside the `response` callback in the `save_file` and `save_files` methods, since in one we are selecting an file, and in the other we are selecting an folder, the code for handle this for both cases are different.

> Does this mean other options can also be supported - in which case we need to support and honour the "do-overwrite-confirmation" property in the portal and default that to "false".

The frontend [filter out unsupported options](https://github.com/flatpak/xdg-desktop-portal/blob/master/src/file-chooser.c#L584-L586). so we cannot add an extra `do-overwrite-confirmation` option.

> So are you thinking of just including a single key that turns on/off the overwrite dialog, or should we work on reaching more feature completeness with the Gtk interface? The second would be much harder.

Fature completeness with the Gtk.FileChooser shouldn't be an goal, we should only provide what's needed for the org.freedesktop.portal.impl.FileChooser interface.

jeremypw:
> The frontend filter out unsupported options. so we cannot add an extra do-overwrite-confirmation option.

That's a pity. I guess the client app will have to continue do the check then.  So does this make this PR a non-starter then?

mcclurgm:
Is your concern that apps calling the portal will check for duplicates, making a dialog in the portal redundant? I can't think of any app that does that. At least for a selected file (as opposed to a directory into which it places multiple files, like Audacity's bulk export). That's what led me to this issue.

I guess we could see what GNOME's portal does, but if we don't handle it in the file chooser, then my experience is that files will be silently overwritten.

jeremypw:
It depends on whether the portal expects the filechooser to *always* check for and confirm overwriting duplicates or *never* to check - since the option is not supported.   You may be right that most apps leave it to the filechooser, but since, currently, that option defaults to `false`, always doing it in the filechooser would be a change in default behaviour.  I don't know what other platforms do.

I notice that Code sets the `do_overwrite_confirmation` option (which will not work with the current portal) so that needs resolving at least.  My app `Gnonograms` however does not set that option and confirms overwrite itself.

 Either way some apps are going to have to change.

jeremypw:
>  the portal may prompt or take some other action to construct a unique file name and return that instead

This does not sound like the portal must *always* take an action.  However, for the elementary portal, I guess we can specify that it does and that apps do not have to do any checking

danrabbit:
Yeah I think making things the responsibility of the portal makes sense anyways so developers don't have to worry about it

mcclurgm:
So are there any design changes I should be making, or do you think this is good to continue how I have it broadly set up here? I'm not quite understanding where this conversation ended up, sorry.

@Marukesu I agree that in principle handling this in `save_file` makes sense. The way to implement it isn't very obvious to me though. If the user chooses to cancel in the overwrite dialog, is there a reasonable way to keep the file chooser open and continue as before? My first guess is that if we avoid the `save_file.callback ()` line (say by returning early from the `response` callback before it gets there) it won't continue through the `save_file` code and close the window. But in my experience that's not working.

Marukesu:
> If the user chooses to cancel in the overwrite dialog, is there a reasonable way to keep the file chooser open and continue as before? My first guess is that if we avoid the `save_file.callback ()` line (say by returning early from the `response` callback before it gets there) it won't continue through the `save_file` code and close the window. But in my experience that's not working.

@mcclurgm the FileChooser calls `destroy ()` in `response.connect_after ()` what closes the dialog anyway, that need to be moved to the portal methods, and the state saving code to be in `dispose ()` instead.

However, i think having the dialog be like the FileConflictDialog offering to replace/rename (and a extra skip option, for the `save_files` case) with cancel beginning the same of clicking cancel in the FileChooser better, and more like the Files Application.

mcclurgm:
So if I understand correctly, you're saying the code that's currently in `response.connect_after` needs to be moved to a different function. I'm not sure what you mean by `dispose`, could you be referring to `destroy`? If that's done, then maybe we could avoid destroying the window when a cancel button is clicked.

I see what you're saying about offering to rename the file in the dialog. Offering to rename does seem like a nice idea. But I'm wary of having a new dialog with no way to get back to the file chooser in its current state, for example if someone wanted to change the directory when they realized that there's a conflict. This could probably use input from UX.

Sorry for letting this PR languish. Since I started working on it things have come up and I don't really have time right now, so I'm going to have to leave it be for a while. I may be able to come back to it, but it may be a while, since there's quite a bit of work left to do. If someone else wants to pick up from my work here I'd welcome it.

danrabbit:
Since the save file portal already has the ability to rename, we probably don't need to add that to this confirmation since "Cancel" for the confirmation should just take the user back to the save file portal where they can rename or change locations.

filechooser-portal/FileChooserDialog.vala
```diff
@@ -473,6 +473,45 @@ public class Files.FileChooserDialog : Hdy.Window, Xdp.Request {
         }
     }
 
+    private void accept () {
+        if (action == Gtk.FileChooserAction.SAVE || action == Gtk.FileChooserAction.CREATE_FOLDER) {
+            // If an existing file would be overwritten, ask for permission first
+
+            assert (select_multiple == false);
```
jeremypw, line 480: Can this be guaranteed as it is set externally?  Would be better to handle this case?

jeremypw:
If the file to be overwritten is read-only then the operation seems to fail silently.

jeremypw:
@Marukesu
> The frontend [filter out unsupported options](https://github.com/flatpak/xdg-desktop-portal/blob/master/src/file-chooser.c#L584-L586).

 Am I right in thinking that this means that there is no need to check that `select_multiple` is false when saving (and the assertion can be omitted)?  In which case I will remove the dialog dealing with that.

filechooser-portal/FileChooserDialog.vala
jeremypw: Testing shows that the portal code disallows select_multiple in this case.

jeremypw:
Attempting to create a new folder with the filechooser that has the same name as an existing folder is not possible - the "Create" button remains insensitive.

The alternative to testing for conflicting file is to attempt the save then handle the resulting "IOError.EXISTS" error, I think.  However, as the FileChooser does not actually save the file I think you have to test.

jeremypw:
I have now eliminated the filename conversion so no need to deal with possible error there.

jeremypw:
@mcclurgm If you have time to try out the changes I made that would be helpful.  I have tried to address most of the outstanding issues you mentioned.

Marukesu:
> Am I right in thinking that this means that there is no need to check that `select_multiple` is false when saving (and the assertion can be omitted)? In which case I will remove the dialog dealing with that.

i see that you already found this out, but yes, the `multiple` option only exists on the OpenFile method.

i still believe that the overwrite check shouldn't be in the FileChooser, but in the portal methods, and the FileChooser shouldn't be deleting itself after emitting `response ()` anymore.

filechooser-portal/FileChooserDialog.vala
```diff
@@ -473,6 +473,56 @@ public class Files.FileChooserDialog : Hdy.Window, Xdp.Request {
...
+        if (action == Gtk.FileChooserAction.SAVE || action == Gtk.FileChooserAction.CREATE_FOLDER) {
+            var file = chooser.get_file ();
+            if (file == null) {
+                response (Gtk.ResponseType.CANCEL);
+                return;
+            }
+
+            if (file.query_exists ()) {
+                var display_name = file.get_basename ();
+                var primary = _("Replace “%s”?".printf (display_name));
```
tintou, line 486:
```suggestion
                var primary = _("Replace “%s”?").printf (display_name);
```
For the i18n to work you have to first translate the raw string then replace the placeholders

filechooser-portal/FileChooserDialog.vala
```diff
@@ -473,6 +473,56 @@ public class Files.FileChooserDialog : Hdy.Window, Xdp.Request {
...
+                var secondary = _("Replacing this file will overwrite its current contents");
```
tintou, line 487:
```suggestion
                unowned var secondary = _("Replacing this file will overwrite its current contents");
```

mcclurgm:
Thank you for your help! I'll try to get to this soon.

I think I agree about moving into the portal handler code that creates the FileChooser dialog. To do so I think we would move the handling from `response.connect_after` to a Gtk destroy signal. That makes sense to me.

Thinking on it some more I'm not sure how I feel about the symlink copy. I'll try to think about it a bit more, but if @elementary/ux has any thoughts that would be great. I'm not sure if we need to special case the link in the UI. I assume that what users expect is that the contents the file points to would be overwritten just like a regular file, which is basically what the default copy describes. We could perhaps say that it will "also overwrite the original file" or something like that.
We may also want to add something to deal with folder links, but that may be separate from this PR scope.

danrabbit:
@mcclurgm the current copy seems to make sense to me :+1:

Marukesu:
> Since the save file portal already has the ability to rename, we probably don't need to add that to this confirmation since "Cancel" for the confirmation should just take the user back to the save file portal where they can rename or change locations.

in `save_files`, the chooser is for selecting an folder, and the filenames are defined by the application in the method call. going back to the chooser will only allow the user to change the directory where the files will be saved.

we need to support renames in the overwrite dialog at least in this case.

> I think I agree about moving into the portal handler code that creates the FileChooser dialog. To do so I think we would move the handling from response.connect_after to a Gtk destroy signal. That makes sense to me.

`destroy ()` isn't an signal anymore in Gtk4, we need to override `Object.dispose ()`.

jeremypw:
@Marukesu 
> i still believe that the overwrite check shouldn't be in the FileChooser, but in the portal methods, and the FileChooser shouldn't be deleting itself after emitting response () anymore.

By "portal methods" you mean in `/filechooser-portal/Main.vala`? Yes, that should not be a problem - I'll do that.

jeremypw:
@Marukesu I've moved the check into `Main.vala` (only for `save_file ()`) and tried to deal with the `destroy ()` issue.  I am not sure I have done it right but it seems to work.

I guess we should really check for overwrites in `save_files ()` as well?  But maybe that can be left for separate PR as it is more complicated.

mcclurgm:
Unfortunately yes, you are expected to rename batch files. From the portal spec:
> An array of strings containing the uri corresponding to each file given by options, in the same order. Note that the file names may have changed, for example if a file with the same name in the selected folder already exists.

I'd personally be fine with punting that for another PR soon, since the dialog itself (and maybe the checking?) will not overlap much with the single file case.

jeremypw:
Yes, we need to be sure this PR is correct before starting on the save_files () case.

filechooser-portal/Main.vala
```diff
@@ -343,10 +339,58 @@ public class Files.FileChooserPortal : Object {
...
+            }
+
+            var replace_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+                primary, secondary, "dialog-warning", Gtk.ButtonsType.CANCEL
+            );
+
+            var replace_button = replace_dialog.add_button ("Replace", Gtk.ResponseType.YES);
+            replace_button.get_style_context ().add_class (Gtk.STYLE_CLASS_DESTRUCTIVE_ACTION);
+
+            var replace_response = replace_dialog.run ();
```
Marukesu, line 382: `Gtk.Dialog.run()` was also removed in Gtk4, to block until the response, you need an `new MainLoop(null, true)` and call `MainLoop.quit()` in the `Gtk.Dialog.response()` signal.

filechooser-portal/Main.vala
```diff
@@ -218,22 +217,20 @@ public class Files.FileChooserPortal : Object {
...
-
         dialog.response.connect ((id) => {
             switch (id) {
                 case Gtk.ResponseType.OK:
-                    _results["uris"] = dialog.get_uris ();
-                    _results["choices"] = dialog.get_choices ();
-                    if (dialog.filter != null) {
-                        _results["current_filter"] = dialog.filter.to_gvariant ();
-                    }
+                    if (check_overwrite (dialog.get_file ()) == Gtk.ResponseType.OK) {
```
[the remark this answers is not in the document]
jeremypw: Yes, it would be annoying for a warning to keep being shown when the application has already granted permission.  So the warning should only be shown when the uri has been chosen *within the filechooser*.  I'll pass the uri instead.

filechooser-portal/FileChooserDialog.vala
```diff
@@ -288,8 +288,6 @@ public class Files.FileChooserDialog : Hdy.Window, Xdp.Request {
 
             settings.set_string ("last-folder-uri", chooser.get_current_folder_uri ());
             settings.set ("window-size", "(ii)", w, h);
```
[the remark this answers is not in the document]
jeremypw: I was concerned the window dimensions might not be valid in `Object.dispose()` but I'll try it.

filechooser-portal/Main.vala
```diff
@@ -252,6 +249,11 @@ public class Files.FileChooserPortal : Object {
         yield;
 
         dialogs.remove (parent_window);
+        if (dialog.register_id != 0) {
+            connection.unregister_object (dialog.register_id);
+        }
```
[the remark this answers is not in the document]
jeremypw: Done

filechooser-portal/Main.vala
```diff
@@ -252,6 +249,11 @@ public class Files.FileChooserPortal : Object {
...
+
+        dialog.dispose ();
```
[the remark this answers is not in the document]
jeremypw: Done

filechooser-portal/Main.vala
jeremypw: Rather than messing with MainLoop I have made the overwrite check async with a modal dialog.  Seems to work - let me know if there are flaws.

filechooser-portal/Main.vala
```diff
@@ -343,10 +339,58 @@ public class Files.FileChooserPortal : Object {
...
         response = _response;
         results = _results;
     }
 
+    private Gtk.ResponseType check_overwrite (GLib.File? file) {
```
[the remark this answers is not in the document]
jeremypw: Done

filechooser-portal/Main.vala
```diff
@@ -209,32 +203,34 @@ public class Files.FileChooserPortal : Object {
...
+                        check_overwrite_uri.begin (chosen_uri, (obj, res) => {
+                            if (check_overwrite_uri.end (res)) {
+                                _results["uris"] = dialog.get_uris ();
+                                _results["choices"] = dialog.get_choices ();
+                                if (dialog.filter != null) {
+                                    _results["current_filter"] = dialog.filter.to_gvariant ();
+                                }
+
+                                _response = 0;
+                                warning ("Save File callback");
```
Marukesu, line 225:
```suggestion
```
no need for this here, or should be downgraded to debug.

filechooser-portal/Main.vala
```diff
@@ -343,10 +332,61 @@ public class Files.FileChooserPortal : Object {
...
+            var replace_button = replace_dialog.add_button ("Replace", Gtk.ResponseType.YES);
+            replace_button.get_style_context ().add_class (Gtk.STYLE_CLASS_DESTRUCTIVE_ACTION);
+
+            var response_ok = false;
+            replace_dialog.response.connect ((id) => {
+                response_ok = id == Gtk.ResponseType.YES;
+                check_overwrite_uri.callback ();
+            });
+
+            replace_dialog.show_all ();
```
Marukesu, line 381:
```suggestion
            replace_dialog.present ();
```

`Gtk.Widget.show_all()` was removed in Gtk4 too, since non-toplevels Widgets are shown by default, Granite.MessageDialog already make they childs visible in Gtk3, so calling `Gtk.Window.present()` works here.

filechooser-portal/Main.vala
```diff
@@ -343,10 +332,61 @@ public class Files.FileChooserPortal : Object {
...
+                } catch (Error e) {
+                    warning ("Could not get info for %s", file.get_uri ());
+                }
+            }
+
+            var replace_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+                primary, secondary, "dialog-warning", Gtk.ButtonsType.CANCEL
+            );
+
+            replace_dialog.set_modal (true);
```
Marukesu, line 370:
```suggestion
            replace_dialog.modal = true;
```
Code Style

filechooser-portal/Main.vala
jeremypw: Yes - that was just for testing.

jeremypw:
@Marukesu Thanks for the suggestion,which I have committed :smile:

filechooser-portal/Main.vala
```diff
@@ -209,32 +203,33 @@ public class Files.FileChooserPortal : Object {
...
+                                }
+
+                                _response = 0;
+                                save_file.callback ();
+                            }
+                        });
+
+                        return; // Continue showing dialog until check completes
+                    } else {
+                        break;
```
Marukesu, line 231: This is only handling success if we show the overwrite dialog.

Marukesu (changes requested):

filechooser-portal/Main.vala
jeremypw: Oops. In fixing I have assumed that it is OK to pre-emptively set `_results[]` because if the response is eventually `CANCEL` it will be ignored.

filechooser-portal/Main.vala
```diff
@@ -178,11 +170,13 @@ public class Files.FileChooserPortal : Object {
...
 
+        var supplied_uri = "";
         if ("current_file" in options) {
             dialog.set_uri (FileUtils.sanitize_path (
-                options["current_file"].get_bytestring (),
-                Environment.get_home_dir ()
+                options["current_file"].get_bytestring (), Environment.get_home_dir ()
             ));
+
+            supplied_uri = dialog.get_uri ();
```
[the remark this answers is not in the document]
jeremypw: Not sure I understand. Doesn't `dialog.get_uri ()` return the current uri after it has been set by the previous line?

jeremypw:
@Marekesu I hope I have fixed the `supplied_uri` problem now, but needs properly testing in Gtk4 as I had to fake a test in Gtk3.  I have not (yet) handled the situation where the `current_name` option is set.  Please advise whether the overwrite check should be done in this situation.

filechooser-portal/Main.vala
```diff
@@ -178,11 +170,17 @@ public class Files.FileChooserPortal : Object {
...
-                Environment.get_home_dir ()
-            ));
+            supplied_uri = FileUtils.sanitize_path (
+                options["current_file"].get_bytestring (), Environment.get_home_dir ()
+            );
+
+            if (supplied_uri != null) {
+                dialog.set_uri (supplied_uri);
+            } else {
+                supplied_uri = "";
```
mcclurgm, line 182: Is this supposed to be `dialog.set_uri ("")`? This line does nothing, setting `supplied_uri` to the value it already has.

filechooser-portal/Main.vala
```diff
@@ -232,9 +220,20 @@ public class Files.FileChooserPortal : Object {
...
+                    if (chosen_uri != supplied_uri) {
+                        check_overwrite_uri.begin (chosen_uri, (obj, res) => {
+                            if (check_overwrite_uri.end (res)) {
+                                _response = 0;
+                                save_file.callback ();
+                            }
+                        });
+
+                        return; // Continue showing dialog until check completes
+                    } else { // No need to check full uri supplied by calling app
```
mcclurgm, line 233: Are you sure this is safe? I can think of cases where an app-supplied URI could conflict. Say that an app provides a "save as" function and passes the existing file as a sample URI for the dialog. In that case, the app-supplied URI would create a file conflict that we would want to check for. For example, I believe that MATLAB does this behavior of passing the current file to save as, at least on Mac (I assume it's the same on other platforms).

If this is something expected by the portal that I'm missing, then you can close this question.

filechooser-portal/Main.vala
jeremypw: This was to ensure `supplied_uri` is not `null` but you are right, the `sanitize_path` function already does this.

filechooser-portal/Main.vala
jeremypw: I think that if the app supplies a URI then it can be assumed that the URI is in fact usable without further check. If not, then that is a problem with the app.  @Marukesu  may wish to comment as he suggested omitting the check.

filechooser-portal/Main.vala
```diff
@@ -343,10 +335,61 @@ public class Files.FileChooserPortal : Object {
...
+
+        var file = GLib.File.new_for_uri (uri);
+        if (file.query_exists ()) {
+            var display_name = file.get_basename ();
+            var primary = _("Replace “%s”?").printf (display_name);
+            unowned var secondary = _("Replacing this file will overwrite its current contents");
+            if (file.query_file_type (FileQueryInfoFlags.NOFOLLOW_SYMLINKS) == FileType.SYMBOLIC_LINK) {
+                try {
+                    var info = file.query_info (FileAttribute.STANDARD_SYMLINK_TARGET, FileQueryInfoFlags.NONE);
+                    if (info != null) {
```
mcclurgm, line 355: According to the docs, `file.query_info` doesn't return null.

filechooser-portal/Main.vala
```diff
@@ -343,10 +333,61 @@ public class Files.FileChooserPortal : Object {
...
+                        primary = _("Replace “%s”?".printf (info.get_symlink_target ()));
+                    } else {
+                        primary = _("Replace the target of “%s”?".printf (display_name));
+                    }
+
+                    secondary = _("Replacing the target file for this link will overwrite its current contents.");
```
mcclurgm, line 361: Trying to workshop the wording for the link case. I think this is more clear to me, but open to input. Especially whether the copy at the end about "the link will remain" is necessary.
Note that this code fails to compile when `secondary` is unowned. I'm also not sure how to handle the very long line that it produces according to elementary style.

```suggestion
                    primary = _("Replace link “%s”?").printf (display_name);
                    if (info != null) {
                        secondary = _("This file is a link to “%s”. Replacing it will overwrite the original's contents. The link will remain.").printf (info.get_symlink_target ());
                    } else {
                        secondary = _("This file is a link. Replacing it will overwrite the original's contents. The link will remain.");
                    }
```

Also note that the original code uses the translation macro improperly. The `_()` call should end before `printf`. This should be fixed whether my new copy is used or not.

mcclurgm (commented):
This is a papercut, but the main file chooser captures the `esc` key, not the overwrite dialog. So if you press `esc` the whole file chooser exits. I think that should only close the frontmost dialog.

filechooser-portal/Main.vala
jeremypw: Fixed

filechooser-portal/Main.vala
jeremypw: Fixed

jeremypw:
@mcclurgm Thanks for the review. I have fixed the `Escape` key issue by overriding keypress rather than keyrelease. Hopefully this will not have unintended consequences. Also fixed the other issue you mentioned, but did not change the text pending agreement on best wording.

mcclurgm:
Thanks for the update. It does work, although performing the action on key press instead of release is inconsistent with other UI. I wonder if it would be possible to capture the key event in the replace dialog instead and then stop propagation there, so it would never reach the `FileChooserDialog` handler. That is how it intuitively would work for me anyway—unless the user chooses to accept, then the `FileChooserDialog` shouldn't be doing any handling. I gave that a try (using `replace_dialog.key_release_event.connect`, but that function is never called. The signal seems to be captured by the `FileChooserDialog` first, and I'm not sure why.

Also, your change was overwritten by the latest merge commit. If we want to use the `key_press_event` solution we'll need to add it back in.

filechooser-portal/Main.vala
```diff
@@ -343,10 +333,58 @@ public class Files.FileChooserPortal : Object {
...
+                }
+
+                secondary = _("Replacing the target file for this link will overwrite its current contents.");
+            }
+
+            var replace_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+                primary, secondary, "dialog-warning", Gtk.ButtonsType.CANCEL
+            );
+
+            replace_dialog.modal = true;
```
mcclurgm, line 368:
```suggestion
            ) {
                modal = true,
                transient_for = parent
            };
```

The dialog should be modal, and I think that it should be modal to the `FileChooserDialog`. Currently the replace dialog appears in the window switcher, which it shouldn't. This fixes that.

This change would also need some changes elsewhere to have a `FileChooserDialog` reference to work with. See other comments in this review.

filechooser-portal/Main.vala
```diff
@@ -343,10 +333,58 @@ public class Files.FileChooserPortal : Object {
         yield;
 
         dialogs.remove (parent_window);
+        dialog.destroy ();
+
         response = _response;
         results = _results;
     }
 
+    private async bool check_overwrite_uri (string? uri) {
```
mcclurgm, line 342:
```suggestion
    private async bool check_overwrite_uri (string? uri, Gtk.Window parent) {
```

Needed to pass around the parent for the replace dialog.

filechooser-portal/Main.vala
```diff
@@ -232,9 +218,20 @@ public class Files.FileChooserPortal : Object {
                     if (dialog.filter != null) {
                         _results["current_filter"] = dialog.filter.to_gvariant ();
                     }
-
-                    _response = 0;
-                    break;
+                    var chosen_uri = dialog.get_uri ();
+                    if (chosen_uri != supplied_uri) {
+                        check_overwrite_uri.begin (chosen_uri, (obj, res) => {
```
mcclurgm, line 223:
```suggestion
                        check_overwrite_uri.begin (chosen_uri, dialog, (obj, res) => {
```

Needed for replace dialog transience.

mcclurgm (commented):
I think the replace dialog needs a transient parent. Currently it doesn't have one. That means that it appears in the window switcher, which it shouldn't.

I assume that it should be transient to the `FileChooserDialog.` The changes suggested here would implement that. This is a set of 3 changes that all need to be committed together to work, since we need to pass around a `FileChooserDialog` reference.

mcclurgm:
Sorry for the late review (again). This looks good to me for the most part. 

I've collected a list of the outstanding issues here.

Potential blocking bugs:

- [ ] If the file you're overwriting is locked (not writeable) the process fails silently. I'm not sure if this should be the responsibility of the filechooser or the app to handle this. This is also separate from asking for conflicts, since the chooser should not allow you to select locked files to begin with (so may be something to address in a separate PR).

Requires design input:

- [ ] Wording for the link case. I think that the current copy is a bit confusing by referring to only the target file by name, and not the link that you selected. I may be wrong on this, but I could see that causing confusion when a completely different filename than you selected appears in the dialog.
- [ ] Should we still perform the conflict check on a URI provided by the app? The solution of checking only with a complete URI and not filename may make sense. To me this depends on how we view the save portal conceptually: is it the user selecting a file to be saved to, or is it the user selecting a file for the app to get access to? I hope this question makes sense. If it's a matter of access, then re-asking for something the app already knows of may be repetitive. However, even if the app has access, if it's a matter of a save location then we want to make sure the user is OK with saving to the app-provided URI.

There are also some minor cosmetic issues, none of which I think are blocking:

- [ ] The main filechooser dialog is not dimmed when the conflict dialog appears. This may be something in Gala, or it may be our fault. Either way, this isn't a big deal and definitely not blocking.
- [ ] The esc key interaction is strange. It dismisses both dialogs, the conflict and the filechooser, which is not ideal. It dismissed the dialog on key down instead of key up unlike other similar interactions in the system. It seems you committed something to fix the behavior dismissing both dialogs, but I'm still getting it on my machine. Maybe we can just give this a pass, since it's a very small issue.

I will also note that there may be some strange behavior when a file is deleted/moved/renamed from under the conflict dialog and you then save to the original, conflicted file. In Code, a warning infobar appears wrongly saying that the file has been deleted. It was deleted, but is now reinstated by saving. This issue is probably outside of the filechooser, but it seems possible that the chooser is passing something strange back to Code that is causing it to present that infobar. I have not been able to check this yet.

Overall, I think this is essentially ready to go in my view. Just needs some input to determine whether we consider any of the above issues worth holding it up for longer.

Marukesu:
>  If the file you're overwriting is locked (not writeable) the process fails silently. I'm not sure if this should be the responsibility of the filechooser or the app to handle this. This is also separate from asking for conflicts, since the chooser should not allow you to select locked files to begin with (so may be something to address in a separate PR).

If the application fails silently with an non-writable file, it's an application bug, the user can always mark the file as non-writable.  Something to check is if the GNOME and KDE portals let you select non-writeable files too, but not for this PR.

> Should we still perform the conflict check on a URI provided by the app? The solution of checking only with a complete URI and not filename may make sense. To me this depends on how we view the save portal conceptually: is it the user selecting a file to be saved to, or is it the user selecting a file for the app to get access to? I hope this question makes sense. If it's a matter of access, then re-asking for something the app already knows of may be repetitive. However, even if the app has access, if it's a matter of a save location then we want to make sure the user is OK with saving to the app-provided URI.

It's both, the user is selecting an file to the app get access, and then save somthing in it.

What i had in mind was "don't warn users about something they're are already aware", if an sandboxed application has an valid "current_file", it's mean that he got the path from an previous portal call.
 
>  The main filechooser dialog is not dimmed when the conflict dialog appears. This may be something in Gala, or it may be our fault. Either way, this isn't a big deal and definitely not blocking.

That's an gala issue, if the FileChooser has an parent, the dim is applied to the parent window again. Launching it without an parent window makes the FileChooserDialog be dimmed as expected.

> The esc key interaction is strange. It dismisses both dialogs, the conflict and the filechooser, which is not ideal. It also dismisses the dialog on key down instead of key up unlike other similar interactions in the system. It seems you committed something to fix the behavior dismissing both dialogs, but I'm still getting it on my machine. Maybe we can just give this a pass, since it's a very small issue.

Maybe we should port the FileChooser to be a Granite/Gtk.Dialog (we would need to port away from Hdy for Gtk4 anyway), but not in this PR.

The `key_release_event` should be kept for now, since that's what the HIG recommends.

> I will also note that there may be some strange behavior when a file is deleted/moved/renamed from under the conflict dialog and you then save to the original, conflicted file. In Code, a warning infobar appears wrongly saying that the file has been deleted. It was deleted, but is now reinstated by saving. This issue is probably outside of the filechooser, but it seems possible that the chooser is passing something strange back to Code that is causing it to present that infobar. I have not been able to check this yet.

All we do is return an URI to the frontend, There's nothing we can do if the file gets deleted after that. If code shows strange behaviour if the file get deleted by it back, it's should be fixed there.

filechooser-portal/Main.vala
```diff
@@ -343,10 +333,59 @@ public class Files.FileChooserPortal : Object {
...
+            });
+
+            replace_dialog.present ();
+            yield;
+            replace_dialog.destroy ();
+            return response_ok;
+        } else {
+            return true;
+        }
+    }
```
Marukesu, line 387: Thinking more about this function, i prefer it begin sync and returning the dialog instead of the dialog response, this way we can handle the `response()` signal directly in the FileChooser callback.

```suggestion
    private Gtk.Dialog create_overwrite_dialog (Gtk.Window parent, GLib.File file) {
        unowned var primary = _("Replace “%s”?");
        unowned var secondary = _("Replacing this file will overwrite its current contents");
        var display_name = file.get_basename ();
        
        if (file.query_file_type (FileQueryInfoFlags.NOFOLLOW_SYMLINKS) == FileType.SYMBOLIC_LINK) {
            try {
                var info = file.query_info (FileAttribute.STANDARD_SYMLINK_TARGET, FileQueryInfoFlags.NONE);
                display_name = info.get_symlink_target ();
            } catch (Error e) {
                warning ("Could not get info for %s", file.get_uri ());
                primary = _("Replace the target of “%s”?");
            }

            secondary = _("Replacing the target file for this link will overwrite its current contents.");
        }

        var replace_dialog = new Granite.MessageDialog.with_image_from_icon_name (
                primary.printf (display_name), secondary, "dialog-warning", Gtk.ButtonsType.CANCEL
            ) {
                modal = true,
                transient_for = parent
            };

        var replace_button = replace_dialog.add_button ("Replace", Gtk.ResponseType.YES);
        replace_button.get_style_context ().add_class (Gtk.STYLE_CLASS_DESTRUCTIVE_ACTION);
        
        return replace_dialog;
    }
```

filechooser-portal/Main.vala
```diff
@@ -232,9 +218,20 @@ public class Files.FileChooserPortal : Object {
...
+                                _response = 0;
+                                save_file.callback ();
+                            }
+                        });
+
+                        return; // Continue showing dialog until check completes
+                    } else { // No need to check full uri supplied by calling app
+                        _response = 0;
+                        break;
+                    }
```
Marukesu, line 234: I think breaking early here make the code more readable: (suggestion already have the changes needed by the `check_overwrite_file/create_overwrite_dialog` sugesstion) 
```suggestion
                   _response = 0;

                    var chosen_file = dialog.get_file ();
                    if (!file.query_exists () || chosen_file.get_uri () == supplied_uri) {
                        break; // No need to check full uri supplied by calling app
                    }
                    
                    var overwrite_dialog = create_overwrite_dialog (dialog, chosen_file);
                    overwrite_dialog.response.connect ((response) => {
                        if (response == Gtk.ResponseType.YES) {
                            save_file.callback ();
                        } else {
                            _results.remove_all ();
                            _response = 2;
                        }
                        
                        overwrite_dialog.destroy ();
                    });
                    overwrite_dialog.present ();
                    return; // Continue showing dialog until check completes
```

jeremypw:
@Marukesu This seems to work pretty well now.  It may be better to merge and iterate with smaller PRs?  There is a Files release coming up and it would be good to get this in.

filechooser-portal/Main.vala
jeremypw: Seems this is not fixed after all ...

jeremypw:
@mcclurgm I have now fixed the `Esc` key issue.

Marukesu:
@jeremypw the portal frontend expect the response begin 0 on success, 1 on cancel, and 2 for any other reason, what we do here is what GNOME do in the Gtk/GNOME portal.

one last change: on `key_press_event()`, return `base.key_press_event()` instead of `Gdk.EVENT_PROPAGATE`. it's breaking the pathbar and name entry now.

jeremypw:
@mcclurgm 
> The esc key interaction is strange. It dismisses both dialogs, the conflict and the filechooser, which is not ideal. It also dismisses the dialog on key down instead of key up unlike other similar interactions in the system. It seems you committed something to fix the behavior dismissing both dialogs, but I'm still getting it on my machine. Maybe we can just give this a pass, since it's a very small issue.

Both the filechooser dialog and the conflict dialog dismiss with `Esc` key down. Before, the conflict dialog dismissed with key down (due to upstream code) and the file chooser dismissed with `Esc` key up.  That was why both dialogs were being dismissed with `Esc` key down followed by key up.  

Actions linked to accelerators fire on key down.

Marukesu (approved):

PR #1993, opened 2022-01-23
Add suggested-action class to accept button

### LINKED ISSUE #1990: FileChooser: highlight accept action
### Problem

Currently, there is no hierarchy to the buttons in the file chooser. I believe that the "accept" button (which can take the form of `Open`, `Save`, or others) should be highlighted. It is the primary action of the window, and according to the HIG, should be highlighted. 

Further advantages: This would also make the dialog easier to parse, since the expected action to be performed is visually prominent. Also, since we use the Enter key as a mnemonic for the accept button, it should be given a visual indication.

Current state:
non-highlighted-save

### Proposal

Add the `suggested-action` style class to the accept button.

highlighted-save

### Prior Art (Optional)

_No response_

If this is something you want, I will happily submit a quick PR. I just wanted to gauge whether this is desired by the design team beforehand.

mcclurgm:
Fixes #1990 

Bildschirmfoto von 2022-01-23 11 43 35

### DIFF
--- a/filechooser-portal/FileChooserDialog.vala
+++ b/filechooser-portal/FileChooserDialog.vala
@@ -112,6 +112,7 @@ public class Files.FileChooserDialog : Hdy.Window, Xdp.Request {
             use_underline = true,
             can_default = true
         };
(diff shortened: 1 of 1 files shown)


danrabbit (approved):
Looks good. Nice work!

PR #1998, opened 2022-01-27
Remove references to Granite.Services.Logger (deprecated)

jeremypw:


src/Application.vala
```diff
@@ -186,10 +182,6 @@ public class Files.Application : Gtk.Application {
         }
 
         /* Handle arguments */
-        if (debug) {
-            Granite.Services.Logger.DisplayLevel = Granite.Services.LogLevel.DEBUG;
-        }
```
tintou, line 191: Either remove the debug argument or do:
```vala
if (debug) {
    GLib.Environment.set_variable ("G_MESSAGES_DEBUG", "all", false);
}

jeremypw: Good point - I was a bit trigger-happy there.

tintou (approved):

PR #1999, opened 2022-01-28
Implement auto-renaming of multiple selected files

jeremypw:
Previous developments squashed into one commit
* Show bulk rename dialog when more than one file selected for renaming
* Allow prefixes and suffixes
* Allow basename same as original, fixed text or partial replacement of original
* Simplified default options
* Allow editing of modifiers if required
* Allow sorting and reversing number sequences
* Allow partial renaming of list

Possible future enhancements in other PRs
* Other file attributes as modifiers or sortby
* Metadata as modifiers
* Character sequence as modifier
* Reorder modifiers


jeremypw:
Initial Dialog:

BulkRenamerOpeningWindow

jeremypw:
Fixed Basename:
BulkRenamerFixedBasename
Original with replacement Basename:
BulkRenamerReplace
Modifier choice:
BulkRenamerModifierOptions
Number Sequence options:
BulkRenamerNumberSeqEdit
Date Options:
BulkRenamerDateEdit
Fixed text options:
BulkRenamerTextEdit

danrabbit:
This is definitely an improvement from the previous branch!

I can see where you're going with laying out the prefix and suffix stuff like this, but I think mixing actions with the output is a little confusing and it means actions move around as the UI becomes more complicated and eventually the renaming method becomes lost in all the buttons.

You can get into some situations here that probably don't make sense adding the same prefixes and suffixes that have already been used:
Screenshot from 2022-02-02 15 19 54

I think another problem here is that you can make a window that's larger than the screen this way pretty easily on a small display

I'm having a hard time thinking of a better way to solve it than the design that GNOME is using. Having a single text entry where you add placeholders like "[Original file name]" or "[1,2,3]" seems incredibly flexible and with a much simpler UI without some of these other issues. It would be very easy to make something into a suffix or prefix or add arbitrary text without having to have explicit actions for that in the UI other than just adding new placeholders for discoverability.

jeremypw:
I was trying not to restrict the user too much but that means allowing to them to do something not sensible (as long as it is not destructive).  I can certainly restrict some things like having more than one prefix/suffix of the same kind and/or the total number of modifications.  Could maybe restyle the "Add -fix" tool to distinguish more obviously from the modifiers?  Could have a single separate the "Add -fix" tool like Gnome which defaults to adding a suffix and have an extra option in the modifier menu to make it a prefix?

Personally, I did not find the Gnome interface particularly flexible (at least not obviously).  It only offers to add a number range as a suffix for me (with few options).  I cannot edit placeholders and if I type one in as a prefix it is not interpreted.  Maybe it has improved in a later version?

jeremypw:
It seems the more advanced abilities of Gnome Files depend on having Tracker installed and working.  I tried installing it but it did not seem to make any difference.  Maybe working better in elementary 7?  I am still using elementary 6.1

jeremypw:
The latest commits limit the modifications to one suffix and/or one prefix (max number is parameterized). The modifier buttons' labels now reflect what they do more precisely.

RenamerLimitedAffixes

jeremypw:
The window size still tends to change so I may push an alternate design to a different PR.

jeremypw:
Using a set of radiobuttons for sorting choice increases the minimum window width and makes width changing on adding modifiers less likely.

RenamerRadioButtonSortby

jeremypw:
Replaced basename combo with radiobuttons:

Unchanged basename (default)
BulkRenamerNoCombo

Modified basename 
BulkRenamerNoComboModified

New fixed basename
BulkRenamerNoComboNew

jeremypw:
@danrabbit I have pared down the interface some more and replaced the combobox with radiobuttons.  Is this going in the right direction?

jeremypw:
Sortby radiobuttons shown when number sequence chosen
BulkRenamerNoComboNumSeq

jeremypw:
@danirabbit Should I continue with this? It has undergone several design iterations.  Maybe time to squash and resubmit?

marco4mp:
Hi, sorry for asking here. But, is there any progress on this? It's a pretty useful feature.

jeremypw:
Sorry Marco,  still awaiting approval.  It is hard to get large amounts of
new code approved but in this case there is no way of breaking it up into
incremental commits.

I could try reminding Danielle again but she is focused on other projects
at the moment.

Jeremy

peteruithoven:
Great work! Would love to see something like this integrated into elementary OS. 

Some quick UX feedback:
People often recommend against horizontal radio buttons:
https://uxplanet.org/radio-buttons-ux-design-588e5c0a50dc#9b5b
For some cases it might be interesting to use the ModeButton? 
Screenshot from 2023-04-15 14 29 26@2x
Screenshot from 2023-04-15 14 54 34@2x
Screenshot from 2023-04-15 14 53 54@2x
Screenshot from 2023-04-15 14 53 34@2x

"The number in order of" might be confused by a kind of sorting/filter option, since it's directly above the before & after preview list.
Why not integrate this "The number in order of" in the Number sequence dialog? 

You must have considered this but why not configure the name parts in a vertical list? I understand it's slightly less intuitive, but it makes the layout easier / more scalable. It might even make it possible to move/drag parts around. 
It could function a bit like the lists you see under Firewall and custom keyboard shortcuts? Having some quick options when you add something, and adding a dropdown per item to customize them
Screenshot from 2023-04-15 14 49 08@2x
Screenshot from 2023-04-15 14 48 28@2x
I've not seen something where you can change the order though, which is a big problem for this usecase.
But  then again, if the max 1 prefix and 1 postfix is enough there is no need. Keeping it simple & constrained might be the way to go to get the first version in.

jeremypw:
@peteruithoven Thanks very much for your feedback :smile:  I am quite happy to iterate on the design as I know that is not my strong point. 

> For some cases it might be interesting to use the ModeButton?

Yes, good idea - I'll try it.


> Why not integrate this "The number in order of" in the Number sequence dialog?

Good idea! If there was a reason I didn't do it this way I've forgotten now :stuck_out_tongue: 

> You must have considered this but why not configure the name parts in a vertical list? 

One of my earlier efforts did have a more vertical layout (see https://github.com/jeremypw/bulk-renamer-plugin) but I liked that the horizontal parts matched the actual name so seems more intuitive.  If width were an issue could probably make the layout responsive. Supporting narrow layouts was probably not a thing when I started this. If we were to want to support arbitrary numbers of prefixes and suffixes then the UI would have to be rethought anyway.

> Keeping it simple & constrained might be the way to go to get the first version in.

Absolutely!  Not sure what the minimum feature set that would be useful for a start is here, but suggestions for paring it down further welcome.

plugins/meson.build
```diff
@@ -4,3 +4,4 @@ subdir('pantheon-files-ctags')
 subdir('pantheon-files-cloud')
 subdir('git')
 subdir('send-by-email')
+# subdir('bulk-renamer')
```
danirabbit, line 7: leftover from previous commits?

danirabbit (approved):
I'm not super in love with this design, but it is a hotly requested feature so I'm inclined to merge and iterate on it later :+1:

jeremypw:
Yes, I wrote this a while back and if I went through it again I would definitely find many ways it could be improved in both coding and design but at least it is something that works and can be improved upon in the future depending on feedback.

plugins/meson.build
jeremypw: Fixed

danirabbit (approved):
:shipit:

jeremypw:
@peteruithoven As this has now been merged, could you raise issues against specific design flaws so that we can iterate on it? Thanks.

Issue #1994, opened 2022-01-24
Unable to show folder


aral:
### What Happened?

Clicked on a regular folder on my laptop’s main drive.

I see this from time to time. No idea how to reproduce. Clicking on the folder again fixes it

![Screenshot from 2022-01-24 12-38-52@2x](https://user-images.githubusercontent.com/41798/150784660-920f9175-8979-435a-8f55-f2520f8a7d80.png)
.

### Steps to Reproduce

n/a (see above)

### Expected Behavior

This message should never appear during regular use.

### OS Version

6.x (Odin)

### Software Version

Latest release (I have run all updates)

jeremypw:
The message indicates that something positively cancelled the operation that enumerates the files in the chosen folder rather than the operation failing for some other reason (e.g. non-existent or connection broken).  If the operation times out (30 seconds) then a different message should be displayed.  Files cancels the operation before closing or reloading the view or after a timeout.  There could possibly be a race condition if the folder is loaded twice in quick succession I guess :shrug:

jeremypw:
The message given about the server is a fallback message given when the reason for the folder not loading cannot be determined except by showing the system error message.

jeremypw:
Not sure if this is related but I notices that sometimes on startup I get an error in the terminal
`Directory.vala:686: Error reported by next_files_async: Operation was cancelled`
and one of the tabs does not reload. This seems to happen with three or more tabs being restored but only for certain folders being focused.  I haven't experienced it during use.

jeremypw:
@aral Please could you confirm which view mode you are using when this happens (Grid, List, Column) and how you open the folder (click on bookmark, click on view, click on pathbar, click on breadcrumb context menu etc).

I am going to submit a PR that stops the pathbar cancelling any directory objects it may be using as a tab may be loading files from the same directory.  This fixes my issue but I cannot reproduce yours yet.

aral:
@jeremypw Hey Jeremy, I believe I was in column view as that’s what I usually use and I likely single clicked but it was a while ago so I don’t remember exactly, sorry (it’s also possible I double-clicked by mistake). Oddly, the screenshot doesn’t show which mode is active.

jeremypw:
Thanks @aral, I'll investigate in column view.  When the alert view is showing the mode switcher is disabled so does not show which mode is active.

aral:
@jeremypw Reproduced it last night (still “random”): column view, double click on folder.

jeremypw:
[...]

PR #2001, opened 2022-01-28
Stop breadcrumbs cancelling directory objects

### LINKED ISSUE #2000: Directory sometimes unexpectedly cancelled while being restored
### What Happened?

Sometimes a directory gets cancelled before it has finished restoring so that tab does not appear.

This seems to be related to the pathbar cancelling its completion directory object when it focuses out so it depends on which folders are being restored and other timing issues.

### Steps to Reproduce

1. Open three tabs as follows:  (1) The Home folder (2) Any folder containing a large number (1000's) of files (3) A small subfolder of Home
2. Focus the small subfolder
3. Close and reopen Files
4. The Home folder is not restored (sometimes)

### Expected Behavior

All restored tabs are successfully loaded (not cancelled during loading).

### OS Version

6.x (Odin)

### Software Version

Compiled from git

### Log Output

```shell
When this issue occurs you get the following message in the terminal, from Directory.vala:

`Error reported by next_files_async: Operation was cancelled`
```

### Hardware Info

_No response_

jeremypw:
Fixes #2000 

Might address #1994  (?)

There does not seem to be any particular reason for pathbar to cancel directory objects it uses for completions when focusing out or otherwise resetting.  Such objects might be being used by a slot being loaded.  This PR stops the pathbar from cancelling directory objects before releasing its reference,


src/View/Widgets/BreadcrumbsEntry.vala
```diff
@@ -170,10 +170,7 @@ namespace Files.View.Chrome {
         }
 
         private void cancel_completion_dir () {
```
danrabbit, line 172: Since this is only called in `reset` can we move this one-liner there? Especially since this function is now a misnomer

danrabbit (changes requested):

danrabbit (approved):

PR #2002, opened 2022-01-31
Ensure view destroyed with associated slot

jeremypw:
This fixes a potential crash when changing modes caused by a race between the view redrawing and the slot is being destroyed.

This does not seem to affect master (afaict) but affects some PRs where the performance of the Directory object is changed,.


danrabbit (approved):
I'm not totally sure how these internals work, but `dir_view` seems to be private to this class, so makes sense to me

PR #2004, opened 2022-02-02
Misc UI cleanups

danrabbit:
Just some preliminary UI cleanups that were easy to do myself

### DIFF
--- a/src/Dialogs/BulkRenamer/RenamerDialog.vala
+++ b/src/Dialogs/BulkRenamer/RenamerDialog.vala
@@ -19,7 +19,7 @@
  *
 */
 
-public class Files.RenamerDialog : Gtk.Dialog {
+public class Files.RenamerDialog : Granite.Dialog {
     public enum RenameBase {
         ORIGINAL,
         REPLACE,
@@ -77,12 +77,12 @@ public class Files.RenamerDialog : Gtk.Dialog {
         actions.add_action_entries (ACTION_ENTRIES, this);
         insert_action_group ("renamer", actions);
 
-        deletable = true;
-        set_title (_("Bulk Renamer"));
+        title = _("Bulk Rename");
         renamer = new Renamer ();
 
         /* Dialog actions */
         var cancel_button = add_button (_("Cancel"), Gtk.ResponseType.CANCEL);
+
         var rename_button = add_button (_("Rename"), Gtk.ResponseType.APPLY);
         rename_button.get_style_context ().add_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);
         renamer.bind_property (
@@ -102,8 +102,9 @@ public class Files.RenamerDialog : Gtk.Dialog {
         suffix_menumodel.append (_("Fixed Text"), Action.print_detailed_name ("renamer.add-text", suffix_var));
 
         prefix_button = new Gtk.MenuButton () {
-            image = new Gtk.Image.from_icon_name ("add-symbolic", Gtk.IconSize.BUTTON),
-            tooltip_text = _("Add Prefix"),
+            always_show_image = true,
+            image = new Gtk.Image.from_icon_name ("list-add-symbolic", Gtk.IconSize.BUTTON),
+            label = _("Add Prefix"),
             menu_model = prefix_menumodel,
             halign = Gtk.Align.END
         };
@@ -112,8 +113,9 @@ public class Files.RenamerDialog : Gtk.Dialog {
         prefix_box.pack_end (prefix_button, false, false);
 
         suffix_button = new Gtk.MenuButton () {
-            image = new Gtk.Image.from_icon_name ("add-symbolic", Gtk.IconSize.BUTTON),
-            tooltip_text = _("Add Suffix"),
+            always_show_image = true,
+            image = new Gtk.Image.from_icon_name ("list-add-symbolic", Gtk.IconSize.BUTTON),
+            label = _("Add Suffix"),
             menu_model = suffix_menumodel,
             halign = Gtk.Align.START
         };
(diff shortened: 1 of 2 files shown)


src/Dialogs/BulkRenamer/RenamerDialog.vala
```diff
@@ -19,7 +19,7 @@
  *
 */
 
-public class Files.RenamerDialog : Gtk.Dialog {
+public class Files.RenamerDialog : Granite.Dialog {
```
danrabbit, line 22: Saves us some styling

src/Dialogs/BulkRenamer/RenamerDialog.vala
```diff
@@ -77,12 +77,12 @@ public class Files.RenamerDialog : Gtk.Dialog {
         actions.add_action_entries (ACTION_ENTRIES, this);
         insert_action_group ("renamer", actions);
 
-        deletable = true;
```
danrabbit, line 80: Dialogs shouldn't have close buttons since that's an ambiguous action. We want to make sure users are making an explicit choice to cancel or rename when closing the dialog

src/Dialogs/BulkRenamer/RenamerDialog.vala
```diff
@@ -102,8 +102,9 @@ public class Files.RenamerDialog : Gtk.Dialog {
         suffix_menumodel.append (_("Fixed Text"), Action.print_detailed_name ("renamer.add-text", suffix_var));
 
         prefix_button = new Gtk.MenuButton () {
-            image = new Gtk.Image.from_icon_name ("add-symbolic", Gtk.IconSize.BUTTON),
-            tooltip_text = _("Add Prefix"),
+            always_show_image = true,
+            image = new Gtk.Image.from_icon_name ("list-add-symbolic", Gtk.IconSize.BUTTON),
+            label = _("Add Prefix"),
```
danrabbit, line 107: I feel like this is such an abstract idea that just image buttons probably aren't sufficient here

src/Dialogs/BulkRenamer/RenamerDialog.vala
```diff
@@ -185,7 +192,7 @@ public class Files.RenamerDialog : Gtk.Dialog {
         var content_box = get_content_area ();
         content_box.pack_start (controls_grid);
         content_box.pack_start (list_grid);
-        content_box.margin = 12;
+        content_box.margin = 10;
```
danrabbit, line 195: this keeps things in line since dialog comes with a border of 2px we can't get rid of for some reason

src/Dialogs/BulkRenamer/RenamerDialog.vala
```diff
@@ -197,7 +204,7 @@ public class Files.RenamerDialog : Gtk.Dialog {
                     replace_entry_revealer.reveal_child = false;
                     break;
                 case RenameBase.REPLACE:
-                    basename_entry.placeholder_text = _("Enter text of original to be replaced");
+                    basename_entry.placeholder_text = _("Text to be replaced");
```
danrabbit, line 207: These should really be labels outside of the entry. Placeholder text should be an example of valid input, it shouldn't be a description for the entry. But less verbose is at least visible

src/Dialogs/BulkRenamer/RenamerDialog.vala
```diff
@@ -294,6 +301,9 @@ public class Files.RenamerDialog : Gtk.Dialog {
         }
         renamer.modifier_chain.add (mod);
         var mod_button = new Gtk.Button.with_label (mod.mode.to_string ()) {
+            always_show_image = true,
+            image = new Gtk.Image.from_icon_name ("pan-down-symbolic", Gtk.IconSize.MENU),
```
danrabbit, line 305: It wasn't immediately obvious to me that these are menu buttons with options, so let's give them a dropdown icon to make that clearer

src/Dialogs/BulkRenamer/RenamerListBox.vala
```diff
@@ -37,35 +37,25 @@ public class Files.RenamerListBox : Gtk.ListBox {
         construct {
             can_focus = false;
             var oldname_label = new Gtk.Label (old_name) {
-                xalign = 0.0f,
-                width_chars = 24
```
danrabbit, line 41: probably shouldn't impose magic number widths here

src/Dialogs/BulkRenamer/RenamerListBox.vala
```diff
@@ -37,35 +37,25 @@ public class Files.RenamerListBox : Gtk.ListBox {
...
             };
 
             var arrow_image = new Gtk.Image.from_icon_name ("go-next-symbolic", Gtk.IconSize.MENU);
-            var ignored_image = new Gtk.Image.from_icon_name ("emblem-disabled", Gtk.IconSize.MENU) {
-                tooltip_text = _("Ignored - name is not changed")
-            };
-            var invalid_image = new Gtk.Image.from_icon_name ("emblem-error", Gtk.IconSize.MENU) {
-                tooltip_text = _("Cannot rename - name is invalid or already exists")
+
+            var status_image = new Gtk.Image () {
```
danrabbit, line 50: We can use a single image and just update its properties to save ourselves some complexity here

src/Dialogs/BulkRenamer/RenamerListBox.vala
```diff
@@ -78,13 +68,16 @@ public class Files.RenamerListBox : Gtk.ListBox {
             notify["status"].connect (() => {
                 switch (status) {
                     case RenameStatus.IGNORED:
-                        status_stack.visible_child = ignored_image;
+                        status_image.icon_name = "radio-mixed-symbolic";
+                        status_image.tooltip_markup = _("Ignored") + "\n" + Granite.TOOLTIP_SECONDARY_TEXT_MARKUP.printf (_("Name is not changed"));
                         break;
                     case RenameStatus.INVALID:
-                        status_stack.visible_child = invalid_image;
+                        status_image.icon_name = "process-error-symbolic";
```
danrabbit, line 75: Where possible we should avoid using only color to convey information

jeremypw:
Thanks @danrabbit - all very sensible suggestions.

Issue #2010, opened 2022-02-13
Executables often fail due to not being launched in the CWD.


ChildishGiant:
### What Happened?

When trying to launch a steam game (just an example) from the executable files seems to do nothing but when ran with the terminal you can see this error:
`/home/allie/.local/share/godot/app_userdata/Luck be a Landlord/install folder/Luck be a Landlord.x86_64: error while loading shared libraries: libsteam_api.so: cannot open shared object file: No such file or directory`

This is because it's being launched with its full path instead of with the CWD (I think). Opening in Gnome Files works just fine so I imagine they run from the CWD. 

### Steps to Reproduce

I obviously can't share the example I'm using here so have this screen recording

https://user-images.githubusercontent.com/13716824/153735767-885d4922-df5f-47d5-84d6-7cd6afd6d98b.mp4

### Expected Behavior

Ideally run executables from the current folder, not their full path or at least show a message when an executable returns non-0 exit code. The former would actually solve the issue and I imagine be easier since there'd be no new UI.

### OS Version

6.x (Odin)

### Software Version

Latest release (I have run all updates)

jeremypw:
What is the filetype and permissions of the executable used in the screencast?  

Note that elementary Files rather discourages launching executables compared with other file managers for security reasons. E.g. desktop files are deliberately not treated as executables but as text files.  Also it looks like scripts that output to a terminal do not have a terminal launched automatically.

jeremypw:
#2011 might fix this - if you are able could you try it out?

ChildishGiant:
> What is the filetype and permissions of the executable used in the screencast?

application/x-executable, 775

> What is the simplest test case for this issue (I do not have Steam games)?

I'll see if I can find/make something :smile:

PR #2011, opened 2022-02-13
Launch commands in terminal so can see output

jeremypw:
May address #2010 

* Always execute files other than desktop files in terminal so can see any output.
* Launch with context like Nautilus
* Set PWD of context to executable's folder

For simplicity all executables are launched in terminal.  If required a dialog could be thrown asking whether to launch in terminal (like Nautilus).

To be considered - any possible security implications - warning required?  Note that activating the item will not run it automatically - this requires choice from the context menu.  However, at the moment, dropping a file onto a script *does* run it so that probably needs some user input to confirm.


jeremypw:
Thanks for trying! The PWD should be set to the location of the executable at least until the process ends by the line `context.setenv ("PWD", FileUtils.get_parent_path_from_path (path));` but maybe that is not working. That should not cause the library not being found though, if it is in one of the folders listed in your PATH variable.  Do you know where `libsteam_api.so` is stored on your computer?  If you open a terminal in the executable folder and launch the executable manually does it work?

If you launch the exec

ChildishGiant:
Yeah, launching through the terminal in the executable folder works. libsteam_api.so is located next to the executable but because it's not being launched from the pwd it can't be found.

jeremypw:
I think this problem with this PR is at least partially due to a bug in elementary Terminal.  See  https://github.com/elementary/terminal/pull/477/  It might be necessary for that PR to be merged for this solution to work.

jeremypw:
Ignore the "thread-safe" and other warnings - they seem to be due to upstream bugs, not this PR.

jeremypw:
Hmm, I put both the executable and path in quotes so the spaces should not have been a problem but admittedly my test script/path did not have spaces.  I'll test again with spaces.  Thanks for testing again - we are making progress!

jeremypw:
@ChildishGiant I have made a workround for the spaces issue so that this PR works with the existing Terminal.  It would be better fixed by improving the way Terminal runs programs but it is hard to get Terminal PRs reviewed.

zeebok (dismissed):
This looks good to me. I trust any significant testing was already between the two of you.

jeremypw:
@zeebok Thanks for having a look at this! It occurs to me now I have looked at it again after a break that it assumes that `io.elementary.terminal` is installed.  While this is usually the case I guess it is possible that the user has replaced it with a different terminal.  I should provide a fallback for that situation.

jeremypw:
Marked as blocked as changes to Terminal seem to prevent work around for spaces in path from working at the moment and no alternative work around immediately obvious.  Will probably require a fix to Terminal to be merged first.

BAProductions:
Threw it’s called backslash!!!!

jeremypw:
@BAProductions The problem seems to lie in `Shell.parse_argv (program_string, out program_with_args);` in Terminal.  This interprets backslashes literally.  However, if you have a more detailed solution please advise.

PR #2012, opened 2022-02-14
Revert using tabular numbers

jeremypw:
May address #2006 (to be tested by reporter)

The reported crash during a file operation is associated with constructing a label and this is the most recent change to the way progress labels are displayed.


jeremypw:
A different cause of the issue has been suggested.

PR #2019, opened 2022-03-28
Drop unnecessary Config.TERMINAL_NAME check

bobby285271:
`Config.TERMINAL_NAME` is set as `pantheon-terminal` so the check will never be true. :thinking:

This is previously dropped in https://github.com/elementary/files/commit/38df48790df3f8a833f2430891e074392256394d ("Respect open with default app from context menu even for folders") but re-added in https://github.com/elementary/files/commit/049b5af08a603e1fdcaecb7b7dd7bc266de14feb ("Meson build"), maybe this is by accident?

`Config.TERMINAL_NAME` likely has no more users after the change so I am dropping it completely.

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -7,7 +7,6 @@
 #define TESTDATA_DIR "@testdata_dir@"
 #define LOCALE_DIR "@locale_dir@"
 #define APP_NAME "@app_name@"
-#define TERMINAL_NAME "@terminal_name@"
 
 #define HAVE_DBUS
 #endif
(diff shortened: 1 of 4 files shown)


jeremypw:
Looks like those two commits were merged in the wrong order.

Isn't the correct thing to do to restore the original intention of those commits to keep the Config.TERMINAL_NAME and set it to `io.elementary.terminal`?

bobby285271:
Ugh... If we keep the `Config.TERMINAL_NAME` and set it to `io.elementary.terminal`, we will (at least) no longer able to open a folder from terminal (see below screenshots) and I guess #589 is trying to "not filter elementary terminal app from open with apps". :crying_cat_face: 

截图 2022-03-30 09-35-01

jeremypw:
Oh, OK. I guess we can merge this then - thanks.

jeremypw (approved):

PR #2021, opened 2022-04-10
Put drop target revealer in separate row

jeremypw:
Alternative approach to #1914, as suggested by @danrabbit 

Needs appropriate fixes to stylesheet to fix the linked issues #1915 and #1969

Drawbacks:
* Even when inactive or unrevealed, the the revealer rows have a finite height resulting in the bookmarks being more widely spaced.  The revealers need to have a non-zero height so that they can act as drop targets.
* When sensitive, the revealer rows are highlighted when hovered which is not desired as they cannot be selected or activated.  Maybe fixable in the stylesheet or with custom CSS.
* Much larger diff.


jeremypw:
Closing as #1914 was merged.

PR #2023, opened 2022-04-12
Do not automatically select files created outside of application

### LINKED ISSUE #2022: Do not automatically select files created outside of application
### Problem

When you have a folder open and a file is created, it is automatically selected no matter the origin of the files.

Peek 2021-09-20 16-56

As said by @HansCronau on #1539: 

> It may seem just a minor annoyance, but this is actually very tricky/risky default behaviour.  
Today I was minding my own business, cleaning out a folder. It turned out that this folder was actively being written (synced) to by another application. All the new files had been (invisibly!) added to my selection of files to delete. Since I hadn't counted my selection, the confirmation dialogue could not warn me. Luckily I noticed that something was off and I could retrieve the unintentionally lost files from the trash.
>
> Most of the time when I've come the window already did have focus. They have in common that an application is actively writing files to a folder and I want to process these files. Currently I have to be very careful to only process those files I have intentionally selected myself instead of also files that just happened to pop in. Currently the safest (and least frustration) option is to simply go do something else, wait for the writing to stop, and only then process all files.  
Common example: process files synced by Syncthing or Nextcloud. (This example also shows how it can be hard to know when the writing has "stopped" and it's safe to start doing things. You may not be the only person doing things.)

### Proposal

Do not automatically select files that have been created external to the application

jeremypw:
Fixes #2022 

* Add another flag to Files.FileChanges.Change
* Pass Change to Directory when notifying additions
* Pass Change.is_internal when signalling file_added
* Only select files when `is_internal` true

Consequential changes
* Dont pass delegate to Directory.query_info_async - use callback closure (so does not need fixed signature)

Other types of file changes are handled as before - the `is_internal` flag defaults to `true` and is ignored.


zeebok (approved):
Works as advertised!

PR #2030, opened 2022-04-19
Remove faulty check blocking pre-selection

### LINKED ISSUE #2029: Regression: When renaming, text is no longer pre-selected
### What Happened?

Previously when renaming a file, the name excluding the extension was preselected.  This no longer happens.

### Steps to Reproduce

* Select a file
* Press F2
* The rename editable opens but no text is pre-selected

### Expected Behavior

Like other file managers and previous behaviour, the name excluding the extension should be preselected.

### OS Version

6.x (Odin)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fix #2029 

A recent commit concerning renaming has resulted in this regression due to the `renaming` flag being set earlier so that the function doing the selection always returned early due to a re-entry check.

This PR just removes the check, which, afaicr was put in place due to duplicate signals calling the function twice when renaming.  This no longer seems to happen so the check is not required.

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -3308,10 +3308,6 @@ namespace Files {
 
     /** name renderer signals */
         protected void on_name_editing_started (Gtk.CellEditable? editable, string path_string) {
-            if (renaming) { /* Ignore duplicate editing-started signal*/
-                return;
-            }
-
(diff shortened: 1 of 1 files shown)


danrabbit (approved):
I can confirm this fixes the regression :+1:

PR #2034, opened 2022-04-21
Show overlay bar only on selection 

### LINKED ISSUE #1389: Display overlay bar on selection

jeremypw:
Fixes #1389


jeremypw:
@danrabbit Are you on board with this suggestion originally from Cassidy?

danrabbit (approved):
Makes sense to me. Makes the overlaybar accessible via keyboard, keeps its behavior more predictable, and looks like it simplifies the code a good amount :+1:

PR #2044, opened 2022-05-19
Prevent hanging on start up under some circumstances

### LINKED ISSUE #2017: unable to start after crash
### What Happened?

So my Files froze again due to [this bug](https://github.com/elementary/files/issues/1858). I terminated it using the "app unresponsive" dialog that showed up (I still don't have any sysmonitor app on my new OS 6.1).

Trying to relaunch it failed a couple of times, so I tried it from terminal and got this message:

```
$ io.elementary.files

** (io.elementary.files:110344): CRITICAL **: 10:05:33.921: files_view_view_container_change_view_mode: assertion 'self != NULL' failed
Zabit (SIGKILL)
```

I don't have a core dump as `coredumpctl` was not installed on my system.

### Steps to Reproduce

I started with a tab with connection to an ssh server from previous day (when asked for password, I let it forget the password immediatelly (first option of the three)). It's one of those servers that drops connection after inactivity (as I mentioned in my [previous report](https://github.com/elementary/files/issues/1858)), but in this case it doesn't matter, because the laptop was suspended over night, so the connection had been dropped regardless.

1. I closed the ssh server tab first.
2. Then I remembered I wanted to download some results, so I reopened it from the history menu.
3. At this point Files froze, the same way as [reported before](https://github.com/elementary/files/issues/1858). So far the same.
4. At this point killing Files should make it work again, so I killed it using the "unresponsive app" dialog.
5. Trying to relaunch failed. Terminal prints the above message.
6. Reboot to finally launch the Files. Success.

### Expected Behavior

As I said before, Files should not freeze just because a connection to a server was dropped. But even then, terminating it should resolve the problem.

### OS Version

6.x (Odin)

### Software Version

### Log Output

```shell
$ io.elementary.files

** (io.elementary.files:110344): CRITICAL **: 10:05:33.921: files_view_view_container_change_view_mode: assertion 'self != NULL' failed
Zabit (SIGKILL)
```

### Hardware Info

Snímek z 2022-03-21 10-48-39

jeremypw:
Partially Fixes #2017  (see also #2018 for remaining fix)

Gvfs seems to return some broken or unresponsive connections as mounted/connected causing Files to attempt to count the number of items in folders - which blocks the interface indefinitely for practical purposes.  Pending an upstream fix, this PR simply does not attempt to count the items in any remote or unmounted folders.

### DIFF
--- a/libcore/File.vala
+++ b/libcore/File.vala
@@ -1100,21 +1100,26 @@ public class Files.File : GLib.Object {
     }
 
     private string item_count () {
-        try {
-            var f_enum = location.enumerate_children ("", FileQueryInfoFlags.NONE, null);
-            var count = 0;
-            while (f_enum.next_file () != null) {
-                count++;
-            }
+        // Risk of blocking indefinitely on some (broken) remote connections
+        if (!is_remote_uri_scheme () &&
+            is_mounted) {
 
-            if (count == 0) {
-                return _("Empty");
-            } else {
-                return ngettext ("%i item", "%i items", count).printf (count);
-            }
-        } catch (Error e) {
-            return _("Inaccessible");
+            try {
+                var f_enum = location.enumerate_children ("", FileQueryInfoFlags.NONE, null);
+                var count = 0;
+                while (f_enum.next_file () != null) {
+                    count++;
+                }
+
+                if (count == 0) {
+                    return _("Empty");
+                } else {
+                    return ngettext ("%i item", "%i items", count).printf (count);
+                }
+            } catch (Error e) {}
         }
+
+        return _(" --- ");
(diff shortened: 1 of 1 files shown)


zeebok (approved):
Code looks good though I haven't been able to recreate the scenario that this fixes

jeremypw:
Just retested after merging master and a regression has appeared - converting to draft till fixed.

jeremypw:
I think I'll close this and look for a better solution.

jeremypw:
@zeebok  Thanks for reviewing anyway!

jeremypw:
Simpler fix now incorporated into #2018

PR #2050, opened 2022-06-10
Rewrite overwrite prompt for link case

mcclurgm:
Follow-up for #1989. Split into a separate PR so the functionality of the main one could be merged.

I think this is a more understandable wording for this case. I am open to feedback on this though.

Note that it fails to compile, seemingly because the result of `printf` is owned, but `primary` and `secondary` are unowned. I'm not quite sure how to fix this.

I also added a doc comment for `create_overwrite_dialog` while here.


jeremypw:
@mcclurgm I have tried to fix this consistent with @Marukesu 's suggestions (although I have reworked the string handling a little differently).  It now compiles and seems to work with the builtin portal tester and also the Flatpak version of Code.  I have little experience of documenting functions - I have tried to follow the format in existing projects but there may be some tweaking required.

jeremypw:
@Marukesu Could cast another eye over this? Particularly the method documentation I added.

filechooser-portal/Main.vala
```diff
@@ -38,6 +38,22 @@ public class Files.FileChooserPortal : Object {
...
+                 "filters", "current-filter", "choices"
+     * @response: (out): the user response
+     * @results: (out): URI to open, choices and read-only information
+
+     * Asynchronously requests the user for a URI to open.
+     * If the URI exists, asks the user for permission to overwrite. If the user
+     * agrees, then the URI is considered safe and added to @results
+     *
+     * If it does not exist, the URI is implicitly considered safe.
+     */
```
Marukesu, line 56: AFAIK, Valadoc style don't include the method name on top, and specify the parameters after the description, Also `@param param_name description` is used instead of `@param_name: description`.

```suggestion
    /**
     * Requests the user for a URI.
     *
     * The URI should point to a file, if "multiple" is true, more than one URI
     * can be chosen by the user. If "directory" is true, the URI should point
     * to a folder instead.
     *
     * A filter can be specified with "current_filter", or "filters", if more
     * than one filter could be used. Application specific options can be added
     * to the dialog with "choices" and a label to the select button set with
     * "accept_label".
     *
     * @param handle 
     *    Object path where the request should be exported
     * @param app_id 
     *    Application originating the call (not used)
     * @param parent_window 
     *    Transient parent handle, in format "type:handle". if type is "x11",
     *    handle should be a XID in hexadecimal format, if type is "wayland",
     *    handle should be a xdg_foreign handle. Other types aren't supported.
     * @param title 
     *    title for the file chooser dialog
     * @param options 
     *    Dictionary of extra options. valid keys are: "accept_label",
     *    "choices", "current_filter", "directory", "filters", "multiple"
     *    and "modal".
     * @param response 
     *    User response, 0 on success, 1 if cancelled, 2 on fail.
     * @param results 
     *    Dictionary with the user choices, possible keys are: "uris",
     *    "choices", "writable" and "current_filter".
     */
```

filechooser-portal/Main.vala
```diff
@@ -139,6 +155,22 @@ public class Files.FileChooserPortal : Object {
...
+                 "filters", "current-filter", "choices"
+     * @response: (out): the user response
+     * @results: (out): URI safe to save and user choices
+
+     * Asynchronously requests the user for a URI to save
+     * If the URI exists, asks the user for permission to overwrite. If the user
+     * agrees, then the URI is considered safe and added to @results
+     *
+     * If it does not exist, the URI is implicitly considered safe.
+     */
```
Marukesu, line 173:
```suggestion
    /**
     * Requests the user for a URI, with intent to write to it.
     *
     * The URI should point to a file, if the URI is already existent,
     * permission is asked for it usage. "current_folder", "current_name" and
     * "current_file" can be used to pre-select a file.
     *
     * A filter can be specified with "current_filter" option, or "filters",
     * if more than one filter could be used. Application specific options
     * can be added to the dialog with "choices" and a label to the select
     * button set with "accept_label".
     *
     * @param handle 
     *     Object path where the request should be exported
     * @param app_id 
     *     Application originating the call (not used)
     * @param parent_window 
     *     Transient parent handle, in format "type:handle". if type is "x11",
     *     handle should be a XID in hexadecimal format, if type is "wayland",
     *     handle should be a xdg_foreign handle. Other types aren't supported.
     * @param title 
     *     title for the file chooser dialog
     * @param options 
     *     Dictionary of extra options. valid keys are: "accept_label",
     *     "choices", "current_file", "current_filter", "current_folder",
     *     "current_name", "filters" and "modal".
     * @param response 
     *     User response, 0 on success, 1 if cancelled, 2 on fail.
     * @param results 
     *     Dictionary with the user choices, possible keys are: "uris",
     *     "choices", and "current_filter".
     */
```

filechooser-portal/Main.vala
```diff
@@ -263,6 +295,21 @@ public class Files.FileChooserPortal : Object {
...
+     * @options: "modal", "current-name", "current-folder", "current-file", "filters", "current-filter", "choices"
+     * @response: (out): the user response
+     * @results: (out): URIs safe to save
+
+     * Asynchronously requests the user for multiple URIs to save
+     * If a URI exists, asks the user for permission to overwrite. If the user
+     * agrees, then the URI is considered safe and added to @results
+     *
+     * If it does not exist, the URI is implicitly considered safe.
+     */
```
Marukesu, line 312:
```suggestion
    /**
     * Requests the user for a URI, with the intent to write files inside it.
     *
     * Filenames are specified via the "files" option, a folder can be
     * pre-selected with "current_folder". If a file with the same name as one
     * in the "files" list exists, it's replaced.
     *
     * Application specific options can be added to the dialog with "choices"
     * and a label to the select button set with "accept_label".
     *
     * @param handle 
     *     Object path where the request should be exported
     * @param app_id 
     *     Application originating the call (not used)
     * @param parent_window 
     *     Transient parent handle, in format "type:handle". if type is "x11",
     *     handle should be a XID in hexadecimal format. if type is "wayland",
     *     handle should be a xdg_foreign handle. Other types aren't supported.
     * @param title 
     *     title for the file chooser dialog
     * @param options 
     *     Dictionary of extra options. valid keys are: "accept_label",
     *     "choices", "current_folder", "files" and "modal".
     * @param response 
     *     User response, 0 on success, 1 if cancelled, 2 on fail.
     * @param results 
     *     Dictionary with the user choices, possible keys are: "uris" and
     *     "choices".
     */
```

Marukesu (commented):
I think code is good. added suggestions about commentary style and expanding the descriptions.

PR #2056, opened 2022-06-24
IconRenderer: thumbnail backgrounds are always iconsize

### LINKED ISSUE #728: Center too-small thumbnails on correctly-sized card
Instead of trying to intelligently resize the whole interactive icon, we should always keep the "icon" size the same, but center any too-small thumbnails (like for small icons) on a background the correct size.

danrabbit:
Drafting because I don't know how to fix list view :upside_down_face: 

Fixes #728

Always render at icon size even if the thumbnail is smaller. Fixes some issues with not being able to see previews under emblems and selection helpers or when the thumbnail is very small or has an extreme aspect ratio

Screenshot from 2022-06-23 20 03 27

### DIFF
--- a/src/IconRenderer.vala
+++ b/src/IconRenderer.vala
@@ -212,9 +212,16 @@ namespace Files {
             }
 
             cr.scale (1.0 / icon_scale, 1.0 / icon_scale);
-            style_context.render_background (cr,
-                                             draw_rect.x * icon_scale, draw_rect.y * icon_scale,
-                                             draw_rect.width * icon_scale, draw_rect.height * icon_scale);
+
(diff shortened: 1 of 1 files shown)


jeremypw:
I think I have fixed issues with List/Column view and emblems now.

jeremypw:
Screenshot from 2022-06-26 10 30 19
Screenshot from 2022-06-26 10 32 18

jeremypw:
@danrabbit Would you like to cast an  eye over my changes for any regressions? Then I will approve.

danrabbit:
@jeremypw thank you for fixing this for me! Really helps a lot in list and column view imo :bowing_woman:

PR #2059, opened 2022-06-27
Fix case sensitivity when providing completions for paths

### LINKED ISSUE #1997: Pathbar completion should be case sensitive
### What Happened?

When typing a path into the pathbar, I typed a directory name with an upper case letter - it was completed automatically even though the actual directory name had a lower case letter.  Pressing enter then unexpectedly produced the "Directory not found" view.

### Steps to Reproduce

1. Type a path to a folder starting with a lower case letter
2. Type the first letter of the folder in uppercase (suggested completion should appear)
3. Press TAB to accept the suggested completion
4. Press Enter
5. See the "Directory not found" view

### Expected Behavior

No suggested completion should appear when the incorrect case is typed (all suggested completions should actually exist).

### OS Version

6.x (Odin)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1997 

* Compute completion only after iterating through all files
* Make work with non-Latin characters
* Improve some variable names
* Keep trailing separator when focusing in so can start typing subpath immediately


zeebok (approved):

jeremypw:
Need discover reason for merged master causing CI failure in all PRs before can merge :disappointed:

PR #2064, opened 2022-07-25
Allow horizontal scrolling when necessary when columns resized in Column View

### LINKED ISSUE #1992: List view columns can extend outside of viewport without scrolling
### What Happened?

Horizontal scrolling is disabled in the list view mode. I have checked the inspector, and it appears that the `hscrollbar_policy` is set to `NEVER`. 

It is possible to make the contents of the list wider than the window, however. This is quite simple to do by resizing the columns, so that the name column is as wide as the window. Any more columns are invisible off the right margin of the window.

### Steps to Reproduce

1. Open Files and change the view mode to list view
2. Resize any column so that it extends to the right (or trailing) margin of the window
3. Notice how all other columns are extending past the viewport, but it is impossible to scroll to see them

### Expected Behavior

Either we should disallow extending the columns wider than the window, or we should allow horizontal scrolling.

As an example. the Mac OS Finder allows horizontal scrolling when the contents are wider than the window.

### OS Version

6.x (Odin)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Add monitoring the size of each column and recalculate total width request when changes (throttled).


jeremypw:
@mcclurgm Are you able to review this? It fixes an issue you raised.

zeebok:
This definitely works. Two thing I notice but I am not sure are important for this PR, two-finger scrolling doesn't work and there is always more scroll area than actual content:

Screenshot from 2023-03-06 23 01 48

Maybe this is on purpose, I don't know, but I figured I would at least point them out in case this affects either somehow

zeebok (approved):

jeremypw:
Removing linked issue as that refers to List View not Column View and is not fixed by this PR.  This PR does fix a related (unreported) issue in Column view.

PR #2069, opened 2022-08-14
Reset pathbar on iconification

### LINKED ISSUE #2068: Path bar text seems to overlap the path bar drawing.
### What Happened?

Please, see below.

https://user-images.githubusercontent.com/38897885/184465283-960b5375-c6ba-41f7-a37a-6fa7e686763e.mp4

### Steps to Reproduce

1. Double-click on the path bar.
2. Minimize and restore the window.

### Expected Behavior

I'm not sure, I think just the path bar draw should appear

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #2068 

This issue seems to be due to the pathbar receiving a `focus_out_event` when the window is iconified but, because it is no longer in a focused toplevel window it does not reset (to maintain entry focus while another window is focused).  However, it does not receive a `focus_in_event` when the window is uniconified.  This results in the entry having keyboard focus but the `is_focus` property returns false (maybe a Gtk bug?).

The simplest solution to this is to have the pathbar reset when the window is iconified.  In practice imo a user is unlikely to want to iconify the window in the middle of editing the pathbar without abandoning the edit so a more complex solution does not seem worthwhile.

### DIFF
--- a/libcore/Widgets/BasicBreadcrumbsEntry.vala
+++ b/libcore/Widgets/BasicBreadcrumbsEntry.vala
@@ -327,18 +327,21 @@ namespace Files.View.Chrome {
                 return true;
             } else {
                 /* This the delayed propagated event */
-                focus_out_timeout_id = 0;
                 base.focus_out_event (event);
 
                 if (context_menu_showing) {
                     return true;
                 }
 
+                // Reset if window has top level focus or is iconified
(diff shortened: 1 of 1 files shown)


jeremypw:
CLosed in favor of #2090

PR #2074, opened 2022-08-19
Fix logic for revealing infobar

### LINKED ISSUE #2073: ConnectServer dialog does not show error information
### What Happened?

Attempted to connect to an FTP server with anonymous login.  The connection failed and returned to the details entry dialog but no reason for failure was visible.

### Steps to Reproduce

1. Open Connect Server dialog
2. Enter incorrect details
3. Try to connect
4. Reason for failure not shown

### Expected Behavior

When a connect attempt fails the reason should be shown.

### OS Version

7.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #2073 

The logic for revealing the infobar was incorrect - this fixes it.


danrabbit (approved):
Makes sense to me!

PR #2081, opened 2022-08-30
Fix tab label disambiguation regression

jeremypw:
At some point tab labels stopped being disambiguated.  This PR moves the diambiguation code to `FileUtils` and further unit tests added to valid the function for some corner cases.

The revised disambiguation code allows losing the `id` property from the ViewContainer class.


danirabbit:
@jeremypw can you resolve conflicts here?

jeremypw:
@danirabbit Conflict fixed.

zeebok:
@jeremypw Conflicts with main again

libcore/tests/UtilTests/UtilTests.vala
```diff
@@ -346,6 +346,64 @@ void add_file_utils_tests () {
...
+        );
+        assert (result == "B/folder");
+    });
+
+    Test.add_func ("/FileUtils/disambiguate_path/different_length", () => {
+        var result = Files.FileUtils.disambiguate_uri (
+            "/A/B/folder",
+            "X/A/B/folder"
+        );
+        assert (result == "A/B/folder");
```
zeebok, line 363: Should this actually be "/A/B/folder" for the expected result since it implies an absolute path rather than relative?

zeebok (approved):
Only have one question otherwise this looks good

libcore/tests/UtilTests/UtilTests.vala
jeremypw: Yes - you are right!  This function is supposed to return enough  to distinguish the two paths.  I'll fix.

zeebok (approved):

PR #2082, opened 2022-09-07
Port to Gtk4

jeremypw:
Very much a work in progress but pushing for visibility.

- [x] Comment out or modify anything that does not compile
- [x] Replace DynamicNotebook with Adwaita widgets
- [x] Use Adw.Headerbar
- [x] Replace Handy with Adwaita   
- [ ] Use EventControllers
- [ ] Use PopoverMenus
- [ ] Rework Drag and Drop
- [ ] Replace IconView with GridView /DirectoryList(?)
- [ ] Replace ColumnView with Gtk.ListView / DirectoryList (?)
- [ ] Replace Files.ListView with ?  to avoid CellRenderers altogether if possible
- [ ] Replace multipurpose LocationBar with stack of separate widgets

PR #2083, opened 2022-09-11
Replace Files.IconView with  Gtk.GridView

jeremypw:
Experiment to scope out difficulties in using new Gtk.GridView widget instead of Gtk.IconView.  Major changes are required to support both views based on Gtk.TreeView/Files.ListModel (List and Column views) and Gtk.GridView/Gtk.MultiSelection.

As the new widget and model are final classes it may not be possible to replicate all of the existing behaviour.


jeremypw:
Suoerceded by latest commits to #2082

PR #2089, opened 2022-10-07
Prevent unexpected file activation after navigation with double-click

### LINKED ISSUE #2088: Double-clicking a folder may open file with single click
### What Happened?

When double-clicking folder the folder opens, and file that appeared under the cursor is open with only single click instead of double-click

### Steps to Reproduce

1. Double-click folder with many files
2. Watch file open with single-click

### Expected Behavior

File always opens with double-click

### OS Version

7.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #2088 

Simplest fix since button handling will need to be rewritten for Gtk4 anyway.  The fix assumes non-directory files always need a double-click to activate.  This fixes the stated issue but note that the stray click on the new folder will *select* a non-directory file instead of activating it.  Users who want to double-click on folders should switch to the reinstated double-click mode anyway.

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -686,6 +686,8 @@ namespace Files {
             block_model ();
             model.clear ();
             all_selected = false;
+            /* Prevent unexpected file activation after navigation with double-click in mixed mode */
+            on_directory = false;
             unblock_model ();
         }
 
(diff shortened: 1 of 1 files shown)


lenemter (approved):
Fixes the issue. Thanks!

PR #2091, opened 2022-10-19
FileOperations: Prepare the move to Vala-generated structures

tintou:
Avoid direct access to the common substructure.


jeremypw (approved):
Makes sense and would not appear to affect function both by inspection and testing.

PR #2092, opened 2022-10-19
More work towards the Vala port

tintou:
Let's continue to make incremental steps towards the Vala port


libcore/FileOperations/CopyMoveJob.vala
```diff
@@ -0,0 +1,45 @@
...
+ * License along with this program; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301, USA.
+ */
+
+public class Files.FileOperations.CopyMoveJob : CommonJob {
+    protected bool is_move = false;
+    protected GLib.List<GLib.File> files;
+    protected GLib.File? destination;
+    protected GLib.HashTable<GLib.File,bool> debuting_files = new GLib.HashTable<GLib.File,bool> (GLib.File.hash, GLib.File.equal);
```
jeremypw, line 23: I do not think `debuting_files` is actually used anymore so maybe take the opportunity to drop it?

jeremypw (commented):
One inline comment. Will dog food for a while.

jeremypw (approved):
I tried a wide range of file operations (create new, create from template, trash, delete, copy, move, with and without permission, progress window). No obvious regressions.  Code looks good.

PR #2097, opened 2022-11-11
CommonJob: Continue the port to Vala

tintou:
Port the common parts to Vala


libcore/FileOperations/CommonJob.vala
```diff
@@ -64,4 +64,184 @@ public class Files.FileOperations.CommonJob {
...
+                                        cancellable);
+            var file_type = info.get_file_type ();
+            dest_fs_id = info.get_attribute_string (GLib.FileAttribute.ID_FILESYSTEM);
+
+            if (file_type != GLib.FileType.DIRECTORY) {
+                var dest_name = dest.get_parse_name ();
+                /// TRANSLATORS: '\"%s\"' is a placeholder for the quoted basename of a file.  It may change position but must not be translated or removed
+                /// '\"' is an escaped quoted mark.  This may be replaced with another suitable character (escaped if necessary)
+                string primary = _("Error while copying to \"%s\".").printf (dest_name);
+                unowned string secondary = _("The destination is not a folder.");
```
jeremypw, line 119: Why is the secondary string unowned?  It does not seem to belong to anything else.

jeremypw (approved):
No regressions found and code looks good.  I could not test all error conditions as there is some duplication of error handling by the UI and some errors never reach file operations. However, the permissions errors and filespace errors where handled correctly.  Undo continued to work.

libcore/FileOperations/CommonJob.vala
tintou: because it is a static string in the code, it is equivalent in C to:
```c
const char *secondary = "my static string";
```
while in Vala:
```vala
string secondary = "my string";
```
is translated in C to:
```c
char *secondary = g_strdup ("my string");
[…]
g_free (secondary);
```
So it means one less string allocation

jeremypw: I see, thanks for the explanation.

Issue #2098, opened 2022-11-12
With new v6.2.0, new empty file closes the app


WebShapedBiz:
### What Happened?

Right-clicking inside the folder and selecting New -> Empty file immediately closes the app with this error in the terminal:

`** (io.elementary.files:8526): CRITICAL **: 08:21:39.090: marlin_file_operations_create_job_construct_file: assertion 'target_filename != NULL' failed` 

No issues with creating a new folder using the same technique.

Also, this has never happened before version 6.2.0.

### Steps to Reproduce

1. Open files
2. Open a folder inside User's home
3. Right-click inside that folder and select New -> Empty file

### Expected Behavior

Well, not to break

### OS Version

6.x (Odin)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
** (io.elementary.files:8526): CRITICAL **: 08:21:39.090: marlin_file_operations_create_job_construct_file: assertion 'target_filename != NULL' failed
```

### Hardware Info

Vanilla installation on Lenovo IdeaCentre AMD Ryzen 7, all apps coming from .deb files (manual and repo) or Flatpack. Nothing compiled, no meddling with the kernel.

JoseExposito:
Hi! Thanks for reporting this crash. It is fixed in https://github.com/elementary/files/pull/2099

WebShapedBiz:
Thank you @JoseExposito for your time and effort.

PR #2099, opened 2022-11-12
Fix crash creating new empty file (#2098)

JoseExposito:
Creating a new empty file (Right click -> New -> Empty File) crashes Files.

There were 2 issues causing the crash:

 - `CreateJob.file` didn't allow to set an empty `target_filename`.
 - `create_job ()` accessed uninitialized memory.

This commit fixes both issues.

Fix: https://github.com/elementary/files/issues/2098
Fixes: 53983bde62c4 ("More work towards the Vala port (#2092)")


jeremypw (approved):
Thanks for fixing so quickly!

PR #2106, opened 2022-12-01
Ensure special user directory icon used for bookmark

### LINKED ISSUE #2105: v6.2.1. at some point, while opened on the desktop, left column icons change
### What Happened?

I've noticed several times now that when Files are opened on the desktop, after a while, the icons in the left column change from styled:
ksnip_20221119-092929

to generic:
ksnip_20221119-110419

Also, I've noticed that when opening Files, the icons in the left column are always all generic and then, entirely visible, get changed to the styled version. This change is so slow that you can actually see it happening every time, and that's something I've never experienced before.

After closing and opening Files again, the icons are always restored.

### Steps to Reproduce

1. Open Files and leave them on the Desktop
2. Do whatever you're doing for an unknown period of time
3. Check the Files window periodically for icon changes.

### Expected Behavior

Well, for the icons not to change

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

- 8× AMD Ryzen 7 4800u with Radeon graphics 
- Advanced Micro Devices, Inc. [AMD®/ATI] Renoir
- 16GB RAM
- 500GB NVMe M.2 SSD

jeremypw:
Fixes #2105

### DIFF
--- a/libcore/Bookmark.vala
+++ b/libcore/Bookmark.vala
@@ -82,13 +82,28 @@ namespace Files {
                 // Get minimal info to determine icon
                 var ftype = gof_file.location.query_file_type (FileQueryInfoFlags.NONE);
                 if (ftype == FileType.DIRECTORY) {
-                    return new ThemedIcon.with_default_fallbacks ("folder");
+                    try {
+                        var path = GLib.Filename.from_uri (uri);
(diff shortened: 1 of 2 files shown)


zeebok (approved):
Never noticed this bug during that old PR or after. This looks like it covers the problem though, thanks!

PR #2107, opened 2022-12-02
AbstractDirectoryView: Change double-click setting label

### LINKED ISSUE #2120: Inconsistent behaviour clicking on edge of row in List View
### What Happened?

New version of Files (see https://blog.elementary.io/updates-for-november-2022/)
works in some areas but not in others...
Screenshot from 2023-01-25 15 06 41

Having selected single click option:
Screenshot from 2023-01-25 15 19 03

I have observed that Bookmarks folders open with single click (as promised :+1: )
All the others require two clicks - one to select, another to open .

### Steps to Reproduce

1. open any folder in Bookmarks - single click works fine
2. try opening folders elsewhere - single click results in folder properties being displayed
3. second click is required to open selected folder in areas other than Bookmarks

### Expected Behavior

I would expect the behaviour (single click) should be the same for all the folders?

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

danirabbit:
Fixes #2120 

This inverts the language of the Single click select option to "Double-click to Navigate".
* When folks talk about this feature they usually say they want "double-click" and don't want "single-click" so this aligns more with the common mental model of how they're interacting with the view
* I think it also makes it a bit less ambiguous since you can always single-click to select with the selection helpers anyways.
* As for writing style, this makes the phrase use a more active voice (action → result) rather than passive (result → action) and we hyphenate because it's the verb form https://www.merriam-webster.com/dictionary/double-click


jeremypw:
I worded it this way since double click navigates either way - it is the behaviour of single-click that changes.  However, I do not object to the change.

jeremypw:
I was waiting for further Desktop Developer input but as there has been an issue raised about this I'll go ahead with merging it.

jeremypw (approved):

PR #2110, opened 2022-12-04
Count items if local

### LINKED ISSUE #2108: Files no longer shows size of folders in list view

jeremypw:
Fixes #2108

Seems to be due to local files no longer defaulting to `is_mounted = true`, so introduced a separate flag `target_is_local` instead. Items are counted if this is true or the file is mounted and native.

Before this change local files failed to give an enclosing mount so the default value of `is_mounted` was changed to `false`.  It now defaults to `false` instead.


jeremypw:
Not clear why CI failing - passes locally.

jeremypw:
Fixed in another PR

PR #2111, opened 2022-12-16
FileOperations: Move the link and copy functions to Vala

tintou:


jeremypw (changes requested):
This works as before with single transfers but introduces an is issue if a second transfer is started before the first has completed:

1) Start copying a large file to a slow disk so that the progress window appears and the operation takes some significant time (e.g. 30 seconds)
2) Start moving a similarly sized file to the same location
3) A second progress bar appears but both stop updating so that Files appears to hang
This does not happen with master - both progress bars update.

The operations do complete if no actions are taken but I did manage to get into a strange state when I tried cancelling the first and then the second operation - then Files really did hang and a reboot was required.

libcore/FileOperations/CopyMoveJob.vala
```diff
@@ -34,12 +34,211 @@ public class Files.FileOperations.CopyMoveJob : CommonJob {
...
+        base (parent_window);
+        this.files = files.copy_deep ((GLib.CopyFunc<GLib.File>) GLib.Object.ref);
+
+        undo_redo_data = new UndoActionData (Files.UndoActionType.DUPLICATE, files.length ());
+        var parent_dir = files.data.get_parent ();
+        undo_redo_data.set_src_dir (parent_dir);
+        undo_redo_data.set_dest_dir (parent_dir);
+    }
+
+    public async bool do_link (GLib.Cancellable? _cancellable) throws GLib.Error {
```
jeremypw, line 78: Not sure if this needs to be async as it does not seem to call any async functions?

tintou: It helps to have a direct replacement of the equivalent function that was on C side

jeremypw:
To retarget on Gtk4 version

PR #2121, opened 2023-01-27
Fix click above and below text in listview

### LINKED ISSUE #2120: Inconsistent behaviour clicking on edge of row in List View

jeremypw:
Fixes #2120 

This was annoyingly awkward to fix in Gtk3 (using Gtk.TreeView).  

It should be easier to ensure this bug does not occur in the Gtk4 port.


src/TextRenderer.vala
```diff
@@ -20,8 +20,8 @@ namespace Files {
     public class TextRenderer: Gtk.CellRendererText {
 
         const int MAX_LINES = 5;
-        private int border_radius;
-        private int double_border_radius;
+        public int border_radius { get; private set; }
```
zeebok, line 23: I don't see this property ever used outside of TextRenderer. This could probably stay private.

zeebok (approved):

src/TextRenderer.vala
jeremypw: Yeah, I ended up using only `double_border_radius`. I'll fix.

zeebok (approved):

PR #2122, opened 2023-01-27
pantheon-files-cloud: Avoid crashes when disconnecting the provider

tintou:
We can't iterate and delete an item in the loop, use GenericArray as we do not really need Gee here.

### DIFF
--- a/plugins/pantheon-files-cloud/plugin.vala
+++ b/plugins/pantheon-files-cloud/plugin.vala
@@ -20,10 +20,10 @@
 public class Files.Plugins.Cloud.Plugin : Files.Plugins.Base {
     Files.SidebarInterface? sidebar;
     CloudProviders.Collector collector;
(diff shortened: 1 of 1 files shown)


jeremypw:
@tintou The code looks good but I am unable to connect to a NextCloud server to test - do you have any suggestions of how to do this or what other services are compatible with the the current version of `libcloudproviders` that the plugin could be used with?.  It looks like `libcloudproviders` only supports the NextCloud client v 2.5 - the current version is v3.4.  Gnome manages to integrate NextCloud through online accounts and Nautius but I haven't studied how.

jeremypw:
I can confirm this fixes the crash when you exit the NextCloud client.  However, NextCloud is automatically restarted and Files remains on the NextCloud plugin bookmark folder - that is something for another PR if it not desired behaviour.

jeremypw (approved):

PR #2125, opened 2023-02-02
Focus FileChooser name entry when saving

### LINKED ISSUE #2047: Filechooser default typing action is search instead of rename
### What Happened?

When saving a file, typing opens up the search action instead of changing the file name. I guess this makes sense on the open file dialogue, but when saving, most of the time one want to edit the name of the file, not search existing files/folders.

### Steps to Reproduce

1. Open the native "Save As" dialogue (for example, right click -> save as on this page)
2. Start typing to change the name of the file
3. The Filechooser starts searching

### Expected Behavior

The filename to save should change.

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #2047


filechooser-portal/FileChooserDialog.vala
```diff
@@ -355,6 +355,14 @@ public class Files.FileChooserDialog : Hdy.Window, Xdp.Request {
...
                 revealer.reveal_child = stack.visible_child_name == "search";
             });
+
+            if (action == Gtk.FileChooserAction.SAVE) {
+                Idle.add (() => {
+                    entry.set_placeholder_text (_("Enter new filename"));
+                    entry.grab_focus ();
+                    return Source.REMOVE;
+                });
+            }
```
Marukesu, line 365: there's no need for the Idle callback, we can set the placeholder right bellow when we are configuring the entry for the first time, and grab the focus in the show() method.

jeremypw: I thought I had tried that and it didn't work but I'll try again.

jeremypw: > there's no need for the Idle callback, we can set the placeholder right bellow when we are configuring the entry for the first time, and grab the focus in the show() method.

I find that an Idle callback is still needed when `entry.grab_focus` is put in the `show ()` method (after `window.focus (Gdk.CURRENT_TIME);`).  Something is grabbing focus away again later otherwise.  I agree it would be good to get rid of the Idle if possible.

I am using Code "Save As" to test and restarting the `io.elementary.files.xdg-desktop-portal` process between changes.

Marukesu: i tested it and it was the TreeView re-grabbing the focus.

in the `current_folder_changed()` signal, we grab the focus to the TreeView to it don't end up in the LocationBar. we can grab the focus to the entry there too, but we will be focusing the entry always when the user change the current folder.

jeremypw: That could be inconvenient for keyboard-only operation I guess but shouldn't be a problem for someone using pointer operation.  I do not know how many keyboard-only users there are or how to prioritise them?

@danirabbit Do you have any input on this, please?

jeremypw:
@danirabbit  Your input would be valued regarding on how to balance the needs of touch/mouse users and keyboard only users in relation this PR.

jeremypw:
@mcclurgm Requesting a review as you contributed to the linked issue.

filechooser-portal/FileChooserDialog.vala
danirabbit: I'm not sure why we had the treeview grabbing focus every time the view changes. AFAICT, keyboard navigation in the view works as expected without having to re-focus.

I think it makes sense to have the view grab focus on `show ()` and then not try to be clever about moving focus around afterwards and let Gtk do the right thing from there

jeremypw:
After the latest commit, either the entry or the treeview is focused on `realize` depending on whether we are saving or opening.  The only other place a widget is focused is in `set_current_name` where the entry is focused.  Should this remain?.

danirabbit:
@jeremypw I would guess that the intentional of that is to do what you're already doing here on realize, so it's probably a duplicate since I think these methods are only called on activate. I don't think apps are allowed to interfere with the portal while it's running anymore

filechooser-portal/FileChooserDialog.vala
```diff
@@ -355,6 +354,13 @@ public class Files.FileChooserDialog : Hdy.Window, Xdp.Request {
             stack.notify["visible-child"].connect (() => {
                 revealer.reveal_child = stack.visible_child_name == "search";
             });
+
+            if (action == Gtk.FileChooserAction.SAVE) {
+                entry.set_placeholder_text (_("Enter new filename"));
+                entry.grab_focus ();
+            } else {
+                tree_view.grab_focus ();
+            }
```
Marukesu, line 363: this don't need to be done during realize, Gtk don't seem to do any modification in the placeholder and the focus widget after construction (like it does with the search entry).

`entry.placeholder_text` can be set when we move it to the action bar. The focus grab is better placed in the construct method itself, since it aren't dealing with modifications in the internal functionality of the chooser widget.

jeremypw: Yes, now the other grab focus calls are removed you are right.  Will fix.

jeremypw: Done

danirabbit (approved):
Let's go!

PR #2132, opened 2023-02-09
Add AppMenu button to headerbar

### LINKED ISSUE #2123: Files should have a settings/cog wheel menu in the headerbar
### Problem

1. Since there are now three settings options in the context menu, it is starting to feel too cramped.
2. Since the settings are global for the app, they have no place in the context menu for a given folder, it would be different if those settings where pertaining to the folder currently opened.
3. With a settings menu we could also expand on the available settings, like the missing icon size/viewport zoom setting, which is accessible through a keyboard/ keyboard + mouse wheel combination but is not teached, hinted or conveyed to the user in any way. Also, a theme setting to follow the system defined dark/light mode or have one independent.

### Proposal

A way to address all those issues would be to place a pop up settings/cog wheel menu in the headerbar.

### Prior Art (Optional)

There are already several default apps that follow this design pattern:
- Code
- Terminal
- Mail

jeremypw:
Fixes #2123 

 - [x] Zoom controls
 - - [x]  Update label of "normal" button (?)
 - - [x] Make zoom insensitive when at max/min zoom
 - [x] Undo/Redo controls
 - [x] Global boolean options
 - - [x] Selection mode for folders
 - - [x] Show/hide hidden files  
 - - [x] Show/hide remote thumbnails
 - - [x] Show/hide local thumbnails
 - - [x] Sort directories first
 - [x] Remove corresponding  context menu items (TBD)

Left for possible future PRs

Screenshot from 2023-02-11 21 01 40
Screenshot from 2023-02-11 21 00 37

### DIFF
--- a/data/io.elementary.files.appdata.xml.in.in
+++ b/data/io.elementary.files.appdata.xml.in.in
@@ -27,7 +27,21 @@
     </screenshot>
   </screenshots>
   <content_rating type="oars-1.1" />
+
   <releases>
+    <release version="6.3.0" date="2023-02-12" urgency="medium">
+      <description>
+        <p>Improvements:</p>
+        <ul>
+          <li>Move global settings from the context menu to an app menu in the headerbar</li>
+        </ul>
+        <p>Minor updates:</p>
+        <ul>
+          <li>Updated translations</li>
+        </ul>
+      </description>
+    </release>
+
     <release version="6.2.2" date="2023-02-03" urgency="medium">
       <description>
         <p>Minor updates:</p>
@@ -42,6 +56,7 @@
         <issue url="https://github.com/elementary/files/issues/2105">Sidebar icons sometimes change</issue>
       </issues>
     </release>
+
     <release version="6.2.1" date="2022-11-15" urgency="medium">
       <description>
         <p>Minor updates:</p>
--- a/data/schemas/io.elementary.files.gschema.xml
+++ b/data/schemas/io.elementary.files.gschema.xml
@@ -34,10 +34,10 @@
       <summary>Thumbnail remote files</summary>
       <description>Show thumbnails for files on remote filesystems</description>
     </key>
-    <key type="b" name="hide-local-thumbnails">
-      <default>false</default>
-      <summary>Do not thumbnail local files</summary>
-      <description>Do not thumbnail files on the local filesystem</description>
+    <key type="b" name="show-local-thumbnails">
+      <default>true</default>
+      <summary>Thumbnail local files</summary>
+      <description>Show thumbnail files on the local filesystem</description>
     </key>
     <key type="b" name="show-sidebar">
       <default>true</default>
--- a/data/schemas/io.elementary.files.gschema.xml
+++ b/data/schemas/io.elementary.files.gschema.xml
@@ -157,6 +157,11 @@
       <summary>icon zoom level</summary>
       <description>Zoom level used by the List View.</description>
     </key>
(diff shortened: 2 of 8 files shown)


jeremypw:
@danirabbit A few questions to resolve regarding this PR: 

 * How to handle labelling of "normal zoom" button. Should it change like in Code and Terminal, in which case how to denote the icon size relative to "normal"?  Or is it OK to leave it with a fixed label or icon?
 * What is optimum order of the boolean option switches.
 * Is there a need/demand to override the system style in Files?  Terminal and Code widgets have their own built-in style handling but for Files this is not the case.

Please apply your design expertise to the appearance.

danirabbit:
@jeremypw this looks awesome!
> How to handle labelling of "normal zoom" button. Should it change like in Code and Terminal, in which case how to denote the icon size relative to "normal"? Or is it OK to leave it with a fixed label or icon?

I think using percentages still makes sense here since we have some views with lots of options that saying like "extra small" or "extra large" starts to get meaningless. I think we should still adjust the label to match what is in the view to be consistent here.

> What is optimum order of the boolean option switches.

I proposed https://github.com/elementary/files/pull/2133 with a few suggested changes to organization/structure/appearance

> Is there a need/demand to override the system style in Files? Terminal and Code widgets have their own built-in style handling but for Files this is not the case.

I don't think so? The reason that Code/Terminal have these manual overrides is because they're using syntax highlighting schemes where people have pretty strong preferences about using dark/light syntax highlighting separate from the rest of the system. I'm not sure that similar demand exists for other apps

> Is it OK to remove the corresponding menuitems from the directory context menu?

Yes! I think we should definitely clean those up so that menu can become simpler

danirabbit:
Ah something else I noticed is that the zoom actions don't become disabled when they git their min/max values. So for example, the zoom in button is still sensitive even when we're at the max zoom and pressing it has no effect

jeremypw:
@danirabbit Thanks for the comments! I'll get on with implementing those tomorrow.

danirabbit:
Oh, actually I'm curious about the "Sort Folders Before Files" option. Is there a reason you wanted to move that here instead of having it with the other sort options in the context menu?

jeremypw:
> Oh, actually I'm curious about the "Sort Folders Before Files" option. Is there a reason you wanted to move that here instead of having it with the other sort options in the context menu?

This is because that is a global option affecting all folders and the file chooser.  The other sort options only affect the individual folder you set them on (because in that case they are stored in the file metadata).

danirabbit:
I think that one still makes sense to have in the context menu with the other sort options, even if we duplicate it in both locations. But just to make sure sorting options are all available at once and someone won't be confused about where to find this when they're changing other sorting options

jeremypw:
> I think that one still makes sense to have in the context menu with the other sort options, even if we duplicate it in both locations. But just to make sure sorting options are all available at once and someone won't be confused about where to find this when they're changing other sorting options

I have no problem with duplicating it in the context menu.

danirabbit (changes requested):
It looks like the "local-thumbnails" check is always disabled when files starts. Somewhere along the way action state isn't being set on startup

jeremypw:
> It looks like the "local-thumbnails" check is always disabled when files starts. Somewhere along the way action state isn't being set on startup

Seems that was a pre-existing bug that this PR exposes.  Now fixed.

danirabbit (approved):
Let's go! :rocket: 

Great work :)

PR #2135, opened 2023-02-10
Invert show local logic

danirabbit:
Inverts the logic of `show-local-thumbnails` and removes duplicates from the context menu

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -1342,7 +1342,7 @@ namespace Files {
                 model.file_changed (file, dir);
                 /* 2nd parameter is for returned request id if required - we do not use it? */
                 /* This is required if we need to dequeue the request */
-                if ((!slot.directory.is_network && !show_local_thumbnails) ||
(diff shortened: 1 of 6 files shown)


src/View/AbstractDirectoryView.vala
```diff
@@ -1342,7 +1342,7 @@ namespace Files {
                 model.file_changed (file, dir);
                 /* 2nd parameter is for returned request id if required - we do not use it? */
                 /* This is required if we need to dequeue the request */
-                if ((!slot.directory.is_network && !hide_local_thumbnails) ||
+                if ((!slot.directory.is_network && !show_local_thumbnails) ||
```
jeremypw, line 1345: Looks like the logic is not inverted here? Shoudn't `!hide_local_thumbnails` -> `show_local_thumbnails`?

jeremypw (changes requested):
One question about the logic otherwise looks good.

src/View/AbstractDirectoryView.vala
danirabbit: Yep, thank you!

jeremypw (approved):

PR #2137, opened 2023-02-12
Update Undo/Redo button tooltips with operation type

### LINKED ISSUE #2136: "Undo" and "Redo" tooltips should specify the task to be undone/redone
### Problem

Currently the "Undo" and "Redo" menu buttons have fairly lengthy and cryptic tooltip labels. It's not clear exactly what will happen when you press those buttons, especially if you've performed a lot of actions

### Proposal

Tooltip labels should be more specific to the action performed such as "Undo paste", "Redo renaming" etc

### Prior Art (Optional)

_No response_

jeremypw:
Fixes #2136 

Simple solution using only the operation type.  Further details regarding the source and target could be added but the tooltip may get long.

### DIFF
--- a/libcore/UndoManager.vala
+++ b/libcore/UndoManager.vala
@@ -16,7 +16,7 @@ namespace Files {
         CHANGEOWNER,
         CHANGEGROUP;
 
-        public string to_string () {
+        public unowned string to_action_string () {
             switch (this) {
                 case COPY:
                     return _("Copy");
--- a/libcore/UndoManager.vala
+++ b/libcore/UndoManager.vala
@@ -14,7 +14,42 @@ namespace Files {
         SETPERMISSIONS,
         RECURSIVESETPERMISSIONS,
         CHANGEOWNER,
-        CHANGEGROUP
+        CHANGEGROUP;
+
+        public string to_string () {
+            switch (this) {
+                case COPY:
+                    return _("Copy");
+                case DUPLICATE:
+                    return _("Duplicate");
+                case MOVE:
+                    return _("Move");
+                case RENAME:
+                    return _("Rename");
+                case CREATEEMPTYFILE:
+                    return _("Create Empty File");
+                case CREATEFILEFROMTEMPLATE:
+                    return _("Create File from Template");
+                case CREATEFOLDER:
+                    return _("Create Folder");
+                case MOVETOTRASH:
+                    return _("Move to Trash");
+                case CREATELINK:
+                    return _("Create Link");
+                case RESTOREFROMTRASH:
+                    return _("Restore from Trash");
+                case SETPERMISSIONS:
+                    return _("Set Permissions");
+                case RECURSIVESETPERMISSIONS:
+                    return _("Set Permissions Recursively");
+                case CHANGEOWNER:
+                    return _("Change Owner");
+                case CHANGEGROUP:
+                    return _("Change Group");
+                default:
+                    assert_not_reached ();
+            }
+        }
     }
 
     public class UndoActionData {
@@ -595,6 +630,23 @@ namespace Files {
                 return null;
             } else {
                 return action;
+            }
+        }
+
+        public string get_next_undo_description () {
+            var action = get_next_undo_action ();
+            if (action != null) {
+                return action.action_type.to_string ();
+            } else {
+                return "";
+            }
+        }
+        public string get_next_redo_description () {
+            var action = get_next_redo_action ();
+            if (action != null) {
+                return action.action_type.to_string ();
+            } else {
+                return "";
             }
         }
 
--- a/libcore/UndoManager.vala
+++ b/libcore/UndoManager.vala
@@ -633,12 +633,12 @@ namespace Files {
             }
         }
 
-        public string get_next_undo_description () {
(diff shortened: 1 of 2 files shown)


jeremypw:
@danirabbit Is this enough information in the tooltip?  If not what format do you suggest?

src/View/Widgets/HeaderBar.vala
```diff
@@ -266,6 +266,32 @@ public class Files.View.Chrome.HeaderBar : Hdy.HeaderBar {
...
+    private void set_undo_redo_tooltips () {
+        var undo_action_s = undo_manager.get_next_undo_description ();
+        var redo_action_s = undo_manager.get_next_redo_description ();
+        if (undo_action_s != "") {
+            undo_button.tooltip_markup = Granite.markup_accel_tooltip (
+                undo_accels,
+                _("Undo %s").printf (undo_action_s)
+            );
+        } else {
+            undo_button.tooltip_text = _("No operation to undo");
```
danirabbit, line 284: I think we should still show the accel here since it will still be useful to teach about its existence

danirabbit:
This is fantastic! I think this is a perfect level of detail

It would be nice to have a comment from someone on locale team like @ryonakano if it's better to translate the whole string `"Undo move"` or if it's okay to do something like `"Undo %s".printf ("Move")`

jeremypw:
> It would be nice to have a comment from someone on locale team like @ryonakano if it's better to translate the whole string "Undo move" or if it's okay to do something like "Undo %s".printf ("Move")

If its not OK then I can use two functions in UndoActionData `get_undo_string ()` and `get_redo_string ()`.

libcore/UndoManager.vala
```diff
@@ -14,7 +14,42 @@ namespace Files {
         SETPERMISSIONS,
         RECURSIVESETPERMISSIONS,
         CHANGEOWNER,
-        CHANGEGROUP
+        CHANGEGROUP;
+
+        public string to_string () {
```
tintou, line 19:
```suggestion
        public unowned string to_action_string () {
```

libcore/UndoManager.vala
```diff
@@ -598,6 +633,23 @@ namespace Files {
...
         }
 
+        public string get_next_undo_description () {
+            var action = get_next_undo_action ();
+            if (action != null) {
+                return action.action_type.to_string ();
+            } else {
+                return "";
+            }
+        }
```
tintou, line 643:
```suggestion
        public unowned string? get_next_undo_description () {
            unowned var action = get_next_undo_action ();
            if (action != null) {
                return action.action_type.to_string ();
            } else {
                return null;
            }
        }
```

src/View/Widgets/HeaderBar.vala
```diff
@@ -266,6 +266,31 @@ public class Files.View.Chrome.HeaderBar : Hdy.HeaderBar {
...
         location_bar.escape.connect (() => {escape ();});
+
+        undo_manager = UndoManager.instance ();
+        undo_manager.request_menu_update.connect (set_undo_redo_tooltips);
+        set_undo_redo_tooltips ();
+    }
+
+    private void set_undo_redo_tooltips () {
+        var undo_action_s = undo_manager.get_next_undo_description ();
+        var redo_action_s = undo_manager.get_next_redo_description ();
```
tintou, line 277:
```suggestion
        unowned var undo_action_s = undo_manager.get_next_undo_description ();
        unowned var redo_action_s = undo_manager.get_next_redo_description ();
```

src/View/Widgets/HeaderBar.vala
```diff
@@ -266,6 +266,31 @@ public class Files.View.Chrome.HeaderBar : Hdy.HeaderBar {
...
+
+        undo_button.tooltip_markup = Granite.markup_accel_tooltip (
+            undo_accels,
+            undo_action_s != "" ?
```
tintou, line 281:
```suggestion
            undo_action_s != null ?
```

src/View/Widgets/HeaderBar.vala
```diff
@@ -266,6 +266,31 @@ public class Files.View.Chrome.HeaderBar : Hdy.HeaderBar {
...
+
+    private void set_undo_redo_tooltips () {
+        unowned var undo_action_s = undo_manager.get_next_undo_description ();
+        unowned var redo_action_s = undo_manager.get_next_redo_description ();
+
+        undo_button.tooltip_markup = Granite.markup_accel_tooltip (
+            undo_accels,
+            undo_action_s != null ?
+            ///TRANSLATORS %s is a placeholder for a file operation type such as "Move"
+            _("Undo %s").printf (undo_action_s) :
```
tintou, line 283: I think that we would have better quality by having a proper undo and redo sentence

jeremypw:
I have now provided separate basic translatable strings for undo and redo actions.  There is still a question as to whether the for "Undo X" where X is a bare verb (which in English is sometimes also a noun) is the best for translatability?  Alternatives e.g. "Undo Copying"  "Revert Copying" or something more verbose.

danirabbit (approved):
I'm super happy with this as is! I think we can revisit later if after some time in use it doesn't feel quite right :)

PR #2146, opened 2023-02-21
Investigate CI failure on GitHub but not locally

jeremypw:
Just trying reverting latest changes to workflow as there is no obvious code changes to cause this (and there is no error locally).


tintou:
Figured-out the issue, we need to add elementary-icon-theme to the docker image

tintou:
Should be fixed with https://github.com/elementary/docker/pull/30

jeremypw:
Solved by docker image update

PR #2147, opened 2023-02-23
Update appdata 6.3.0

jeremypw:
As well as release updates, translation and launchable tags were added similar to those for `io.elementary.terminal`.


danirabbit:
Adding the translation tag would make it show in AppCenter if Files is 100% translated in a certain language. It's not strictly necessary since Files is pre-installed but it is nice to set a good example

danirabbit (approved):
🎉 🚀

PR #2155, opened 2023-03-07
Fix filechooser portal compile warnings

jeremypw:
Silence some compiler criticals coming from the filechooser portal.

 - [x] Lose an unused var
 - [x] Do not set Gdk.allowed_backends to null contrary to its (Vala) signature.  Gdk source code indicates that when allowed_backends is null, then "*" will be used.
 - [ ] Do not use null as parameter to Gdk.Display.open_display contrary to its (Gtk3) signature.  NOTE: In Gtk4, the signature allows a null.  It is not clear what should be used instead of null in Gtk3 - I could not determine what will be returned with a null parameter.


jeremypw:
To retarget on Gtk4 version

PR #2156, opened 2023-03-08
Ensure overlay appears after rubberband selection in ListView

### LINKED ISSUE #2151: Overlay does not appear after rubberband selection in List View if started outside list
### What Happened?

A properties overlay is supposed to appear whenever files are manually selected and in most cases this happens.  However when a group of files was selected in List View by rubberbanding no overlay bar appeared. It was noticed that this happened only if the starting point of the rubberband was below the last item in the list.

### Steps to Reproduce

1. Open a folder with a few children in List View
2. Size the window so that blank space appears below the last item
3. Button down on the blank space and drag upwards to rubberband-select some items
4. After button up, observe no property overlay appears.
5. Repeat but start the rubberband on one of the items - the overlay appears

### Expected Behavior

The overlay appears after manual selection however performed.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

### LINKED ISSUE #2152: Multiselection in List View does not persist when changing to Icon View
### What Happened?

When more than one file was selected in List View and the ViewMode changed to Icon View, the selection was not reproduced in the Icon View.   However, if the operation was reversed, i.e. the selection was made in Icon View and the view mode changed to List View, the selection was reproduced in the new view.  It was also noticed that a single selected file persisted in both directions.

### Steps to Reproduce

1. Open a folder in List View
2. Select more than one item
3. Change to Icon View
4. No files are pre-selected in the new view

### Expected Behavior

The selection should persist when changing view modes regardless of in which direction.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #2151 
Fixes #2152 

This issue was caused by a subtle difference in Gtk.TreeView and Gtk.IconView.  The former does *not* signal selection changes caused by pointer movement during rubberbanding whereas the latter does.  So if the rubberband starts outside the list the selection change never gets signalled.

So we have to ensure AbstractDirectoryView updates and signals the selection change in the button release handler.

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -3644,7 +3644,6 @@ namespace Files {
             uint button;
             event.get_coords (out x, out y);
             event.get_button (out button);
-            update_selected_files_and_menu ();
             /* Only take action if pointer has not moved */
             if (!Gtk.drag_check_threshold (get_child (), (int)drag_x, (int)drag_y, (int)x, (int)y)) {
                 if (should_activate) {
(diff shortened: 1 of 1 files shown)


lenemter (approved):
Fixes the issues. Nice work!

PR #2159, opened 2023-03-16
Switch to GHCR containers

davidmhewitt:


danirabbit:
@davidmhewitt what's the "stable" target here? Do we just need to update that in the docker repo to point to OS 7 instead of OS 6?

davidmhewitt:
`stable` is OS 7 (with stable PPA), and `unstable` is OS 7 (with daily PPA).

This is either a genuine failure, or a false negative due to a ropey test.

danirabbit (approved):

PR #2157, opened 2023-03-08
Ensure overlay appears after drag-dropped files are selected

jeremypw:
Fixes #2153

PR #2150, opened 2023-03-05
Show git status correctly after reload

jeremypw:
Fixes #1161  - refresh status list when folder reloaded
Fixes #2168 - do not ignore folders incorrectly  (do not add ignored files to the status map)

The latter issue was caused by ignored files' status being bubbled up to their non-ignored parent folder. Excluding them from the status map fixes this.

Install the PR in order to test.

PR #2167, opened 2023-04-05
Release 6.3.1

jeremypw:
- [x] Update version in meson
 - [x] Update appdata
 - [x] Merge bitesize PRs 
 - - [x] #2150
 - - [x] #2157

### DIFF
--- a/data/io.elementary.files.appdata.xml.in.in
+++ b/data/io.elementary.files.appdata.xml.in.in
@@ -31,6 +31,22 @@
   <content_rating type="oars-1.1" />
 
   <releases>
+    <release version="6.3.1" date="2023-04-05" urgency="medium">
+      <description>
+        <p>Minor updates:</p>
+        <ul>
+          <li>Updated translations</li>
+        </ul>
+      </description>
+      <issues>
+        <issue url="https://github.com/elementary/files/issues/2141">Reinstate mount option in storage menu</issue>
+        <issue url="https://github.com/elementary/files/issues/2148">Fix crash when entering path on other Linux</issue>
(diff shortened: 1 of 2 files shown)


jeremypw:
@danirabbit Just to clarify - does one need to update the screenshot locations for every release even if the screenshots are not new?  Or were they changed?

danirabbit:
@jeremypw probably not? But I figure to be safe :sweat_smile:

danirabbit (approved):
Looks good! Nice release :rocket:

jeremypw:
> @jeremypw probably not? But I figure to be safe sweat_smile

OK, I never used to but may AppCenter requires it now :shrug: .  I'll make a note to do it from now on.

jeremypw:
@danirabbit Just another clarification; appdata should now include *all* issues fixed in the <issues> section, even technical or internal changes? We did not used to mention these in the <description> section.

danirabbit:
@jeremypw Yeah the `<issues>` tag should include any github issues that were closed and the label there is supposed to match the title of the issue in GitHub, according to the spec. And the `description` tag should be a non-technical description of changes and can be basically whatever we want to say. So if you wanted to write release notes that were more personal or like a mini blog or whatever you could :) I've seen folks say they appreciate more descriptive release notes for iOS updates for example

jeremypw:
@danirabbit Ah right, that's helpful.  Presumably you *could* edit the issue description in GitHub if were unacceptable in some way e.g. bad spelling, bad grammar or unintelligible although I guess that should have been done earlier ...  But basically you describe the problem not the fix.

danirabbit:
Yeah exactly!

PR #2171, opened 2023-04-05
AbstractPropertiesDialog: replace StorageBar with Gtk.LevelBar

danirabbit:
Since we only ever use capacity here and don't break things down by file type we don't need all the added complexity. Like the Gtk4 port but maintains the label.

## BEFORE
Screenshot from 2023-04-05 16 10 10

## AFTER
Screenshot from 2023-04-05 16 13 38


jeremypw (changes requested):
Code looks good - just a query as to the value shown on the storage bar

danirabbit:
@jeremypw was there supposed to be an inline comment maybe? :sweat_smile:

src/Dialogs/AbstractPropertiesDialog.vala
```diff
@@ -154,10 +154,25 @@ protected abstract class Files.View.AbstractPropertiesDialog : Granite.Dialog {
             uint64 fs_available = file_info.get_attribute_uint64 (FileAttribute.FILESYSTEM_FREE);
             uint64 fs_reserved = fs_capacity - fs_used - fs_available;
 
-            storagebar = new Files.StorageBar.with_total_usage (fs_capacity, fs_used + fs_reserved);
-            update_storage_block_size (fs_reserved, Files.StorageBar.ItemDescription.OTHER);
+            storage_levelbar = new Gtk.LevelBar.for_interval (0, fs_capacity) {
+                value = fs_used,
```
jeremypw, line 158: Shouldn't this be fs_used + fs_reserved, which is the value used in the storage label?  I am not sure how big fs_reserved is compared to the capacity but you could have the storage bar showing "not full" but the label showing "no available space".

jeremypw:
@danirabbit Oops, I wonder where it went? I have re-entered it.

src/Dialogs/AbstractPropertiesDialog.vala
danirabbit: Oops thanks for catching that!

src/Dialogs/PropertiesWindow.vala
```diff
@@ -250,7 +250,7 @@ public class PropertiesWindow : AbstractPropertiesDialog {
         size_value.label = format_size (total_size);
         contains_value.label = get_contains_value (folder_count, file_count);
         update_widgets_state ();
-        update_storage_block_size (total_size, Files.StorageBar.ItemDescription.FILES);
+        storage_levelbar.value = total_size;
```
jeremypw, line 253: You should not update the storage bar total size here now - it results with the storage bar only showing the selection size, overriding the desired filesystem used + reserved.  This was intended to show a separate block for the selection size (but didnt work)

src/Dialogs/VolumePropertiesWindow.vala
```diff
@@ -97,7 +97,7 @@ public class VolumePropertiesWindow : AbstractPropertiesDialog {
         }
 
         create_storage_bar (info, 3);
-        update_storage_block_size (used_space, Files.StorageBar.ItemDescription.FILES);
+        storage_levelbar.value = used_space;
```
jeremypw, line 100: You should not update the storage bar total size here now - it results with the storage bar only showing the selection size, overriding the desired filesystem used + reserved.

src/Dialogs/AbstractPropertiesDialog.vala
```diff
@@ -154,10 +154,25 @@ protected abstract class Files.View.AbstractPropertiesDialog : Granite.Dialog {
...
+                create_storage_label (fs_used + fs_reserved, fs_capacity)
+            );
 
-            info_grid.attach (storagebar, 0, line + 1, 4, 1);
+            info_grid.attach (storage_levelbar, 0, line + 1, 4);
+            info_grid.attach (storage_label, 0, line + 2, 4);
+
+            storage_levelbar.notify["value"].connect (() => {
+                storage_label.label = create_storage_label ((uint64) storage_levelbar.value + fs_reserved, fs_capacity);
+            });
```
jeremypw, line 175: This is no longer needed (and the storage bar value already includes fs_reserved now)

jeremypw (changes requested):

jeremypw:
Noticed some more problems when showing the properties of a small folder.  It appears to work on large folders as long as the spinner is active because the storage value has not been overwritten yet.

danirabbit:
That makes things MUCH simpler haha

jeremypw (approved):
This now reproduces master behaviour with much less code (and extra color!).  Noted that the storage bar only appears if your selection includes a folder.  Not sure if this is desired behaviour but it matches master.  A single (e.g. video) file could easily consume more space than a small folder so I am not sure what the logic of this was.

PR #2185, opened 2023-05-02
Get current date properly

### LINKED ISSUE #2177: [Renamer] "Current Date" indicates wrong date & time
### What Happened?

Selecting "Current Date" in the Renamer shows wrong date & time:

スクリーンショット 2023-04-28 20 59 21

### Steps to Reproduce

I tested with the following revision: 94cb0f7c4b192f001798f61e68360d786c6ac8c5

1. Open Files with and create multiple files/folders
2. Select them, perform secondary click, and "Rename…"
3. Click "Add Prefix" and select "Creation Date"
4. Click "Date created" and select "Current Date" & "Default Format - Date and Time", then click "Apply":
スクリーンショット 2023-04-28 20 59 04
5. See the error

### Expected Behavior

Selecting "Current Date" should indicate the actual current date & time.

### OS Version

Early Access

### Software Version

Compiled from git

### Log Output

```shell
N/A
```

### Hardware Info

Not relevant

jeremypw:
Fixes #2177


src/Dialogs/BulkRenamer/RenamerModifier.vala
```diff
@@ -247,7 +247,7 @@ public class Files.RenamerModifier : Object {
                         dt = file.modified;
                         break;
                     case RenameDateSource.NOW:
-                        dt = (uint64)get_monotonic_time ();
+                        dt = 0;
                         break;
                     default: // Created
                         dt = file.info.get_attribute_uint64 (GLib.FileAttribute.TIME_CREATED);
```
ryonakano, line 253: Is it fine to consider as the case of `RenameDateSource.NOW` if dt is 0?

I think the case of `default` at this line also might return 0, according to the valadoc of [GLib.FileInfo.get_attribute_uint64()](https://valadoc.org/gio-2.0/GLib.FileInfo.get_attribute_uint64.html):

`If the attribute does not contain an unsigned 64-bit integer, or is invalid, 0 will be returned.`

If this happens, dt is set to 0 and `get_formated_datetime()` would return the current datetime.

jeremypw: Thanks for pointing that out. We can add a test for the presence of the TIME_CREATED attribute, but what would be the appropriate date if the attribute is absent?  Maybe it would be better to not show the option at all if the attribute is absent,

jeremypw:
@ryonakano Thanks for the review. I have tried to address your concern with missing date attributes in a simple way.  Now if a date attribute is missing an empty string prepended.  I have not attempted to change the option combobox according to what attributes are present in this PR to keep it small. It is only likely be an issue if bulk renaming over e.g. FTP so a corner case which can be fixed later.

ryonakano (approved):
Code looks good and confirmed the issue gets fixed.

Thank you very much for your work on this fix :tada:

PR #2186, opened 2023-05-04
BulkRename: Differentiate identical strings for translating

lenemter:
Some of the strings used by Bulk Rename are identical to the used ones, which can result in inaccurate translations, this branch marks such strings with "bulk-rename" context.


tintou:
I would suggest to add `/// TRANSLATORS: the context explained` to the line before the string definition to allow translators to actually know why it might be different from other strings :)

tintou (approved):

Issue #2194, opened 2023-05-07
Improve bulk rename UI


peteruithoven:
### Problem

The new multiple / bulk rename feature is great but could use some improvements regarding it's UI. 

For reference:
![Screenshot from 2023-05-07 18 58 20@2x](https://user-images.githubusercontent.com/523210/236691586-41d67dcd-5a99-41a9-bf77-391e2cd8afe0.png)
![Screenshot from 2023-05-07 18 56 09@2x](https://user-images.githubusercontent.com/523210/236691483-e8d05926-abf2-43fd-96db-9d420bd90e7c.png)
![Screenshot from 2023-05-07 18 55 54@2x](https://user-images.githubusercontent.com/523210/236691485-374d266a-bc59-428e-913c-ded82c03ca01.png)
![Screenshot from 2023-05-07 18 55 45@2x](https://user-images.githubusercontent.com/523210/236691486-24426ce9-e546-4d3d-90ec-c26191400dd0.png)

Some problems:
1. After selecting a prefix/suffix types the button is replaced for a settings dropdown button
  1. People need to open the popup to find the settings while people might think the dropdown only contains the prefix/suffix types. It's not obvious that the dropdown button contains all the settings.
  2. It's not possible to easily switch between different prefix/suffix types.
  3. It's not possible to preview the effects of prefix/suffix options without pressing apply.
2. One of the number sequence settings is outside of the popup. 
3. Better to use active language for basename options? 
4. Subjective but selecting the basename through a ModeButton might be more appropriate? 
5. Some input fields rely on only the placeholder for it's purpose, meaning that when the field has focus or has content it's purpose is hidden. 
6. The second dropdown in the Date dropdown does 2 things, describing format and whether to include time. 

### Proposal

Some quick ideas:
- Use Modebutton for basename.
- Use active language for basename options.
- Using labels to describe purpose input fields. 
- If we only support 1 prefix and 1 postfix we might as well inline the settings. Enabling quicker intuitive access to the settings. 
- Separate Include time checkbox, which could be hidden when not appropriate.
 
<img width="688" alt="Bulk renamer dialog" src="https://user-images.githubusercontent.com/523210/236692296-e3f1beea-8e46-4591-9184-01697aedf5ec.png">
<img width="894" alt="Bulk renamer dialog2" src="https://user-images.githubusercontent.com/523210/236692300-d09df3b1-f2ea-48bd-a521-552e13c3f0fd.png">
[...]

PR #2201, opened 2023-05-12
Bulk Rename: Use ModeButton for basename options

jeremypw:
Partially addresses #2194


danirabbit:
I haven’t built this yet, but ModeButton is removed in Gtk4 so I’m not sure it’s a good idea to introduce a new one

jeremypw:
@danirabbit Ah - I forgot that. I'll try and do it a Gtk4 compatible way - or do you want to leave it as it is?

jeremypw:
Screenshot from 2023-05-14 12 18 17

danirabbit (approved):
LGTM!

PR #2219, opened 2023-06-27
Update freespace when device list box entered

### LINKED ISSUE #2035: Free disk space not updated
### What Happened?

When i hover over the root filesystem, it reports 6,3 GB free disk space, but more is already used and only 200 MB are actually free.

### Steps to Reproduce

1. Check free disk space in files

Screenshot from 2022-04-28 19-34-41

2. Use some of that space
3. check disk space again. it has not changed
4. check with a different tool like `df -h` in terminal

```
[davidak@gaming:~]$ df -h
Filesystem      Size  Used Avail Use% Mounted on
devtmpfs        1.6G     0  1.6G   0% /dev
tmpfs            16G  477M   16G   3% /dev/shm
tmpfs           7.9G  7.7M  7.9G   1% /run
tmpfs            16G  480K   16G   1% /run/wrappers
/dev/sda2       450G  427G  2.1G 100% /
tmpfs            29G  153M   29G   1% /tmp
tmpfs           3.2G   84K  3.2G   1% /run/user/1000
```

the available space is different!

### Expected Behavior

the free disk space should be displayed in real time when i hover it

### OS Version

Other Linux

### Software Version

### Log Output

_No response_

### Hardware Info

Internal SATA SSD

jeremypw:
Closes #2035 

Rather than polling the filesystem and in the absence of a signal emitted when a tooltip is shown, the freespace is updated when the pointer enters the device list box so that the tooltip will be accurate if an external file operation takes place that is not reflected in any of the open views.

### DIFF
--- a/src/View/Sidebar/AbstractMountableRow.vala
+++ b/src/View/Sidebar/AbstractMountableRow.vala
@@ -326,6 +326,7 @@ public abstract class Sidebar.AbstractMountableRow : Sidebar.BookmarkRow, Sideba
     }
 
     public virtual void update_free_space () {
+        debug ("update free space");
         add_mountable_tooltip.begin ();
     }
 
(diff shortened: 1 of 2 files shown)


jeremypw:
To retarget on Gtk4 version

PR #2220, opened 2023-06-28
Avoid short-lived temporary files appearing in view

### LINKED ISSUE #2216: Renaming a file creates a ghost file in the same directory
### What Happened?

Renaming a file creates a ghost file with the file's previous name and a grey text icon. After refreshing the folder, it's gone.

image

### Steps to Reproduce

1. Select a file
2. Hit F2 to rename
3. Enter a new name
4. Hit enter to confirm

### Expected Behavior

The ghost file should not be present.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

image

jeremypw:
Fixes #2216 

The issue is caused by adding a file to the model occurring asynchronously after query the fileinfo whereas files are removed immediately.  This causes very short-lived files being deleted before they are created.

This PR fixes the problem by adding the file to the model before querying the file info.  In order to do this it is necessary to use the fallback method of determining whether the file is hidden but this should make little practical difference.

### DIFF
--- a/libcore/Directory.vala
+++ b/libcore/Directory.vala
@@ -912,39 +912,33 @@ public class Files.Directory : Object {
         }
     }
 
-    private void add_and_refresh (Files.File gof, bool is_internal) {
-        // Check whether we have FileInfo
-        if (!gof.ensure_query_info ()) {
-            critical ("FILE INFO unavailable");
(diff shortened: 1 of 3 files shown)


jeremypw:
Closing - based on wrong branch

PR #2224, opened 2023-06-29
Actions: master → main

danirabbit:


tintou:
I believe that we should pause the Gettext action until we have a solid port, I don't want to mess with translator's time

danirabbit:
@tintou I think weblate is still pointer at `master` right now. So we can pause weblate but either way this branch wouldn’t change what they see I don’t think

tintou:
I would just pause it here too to avoid useless commits coming in the way

davidmhewitt:
The `6.4.1` stable release was released from `main` and I've re-pointed weblate at `main` today because of that, so we can get translations for release notes etc. It seems `master` is no longer active?

And then I've noticed this PR. It feels like this should be merged if `main` is now our active development branch that we're releasing from?

jeremypw:
@davidmhewitt Yes, `main` was originally intended for Gtk4 development only and the next release was going to be Gtk4 but it turns out that some intermediate Gtk3 work was needed and a further Gtk3 release to fix an issue in the last `master` release.

davidmhewitt:
I think now that there are strings that need to be translated on the `main` branch because a release was done here, we should leave weblate pointing at `main`.

Since this PR hasn't been merged, the Gettext action isn't going to run on `main` and cause unnecessary string changes. So I think we're fine in this state, so long as we don't merge this PR. Will leave it in draft status.

If we need to do another Gtk3 release from this branch, the POT update commands will have to be run and changes from them committed manually. Then once the strings are stabilised, we can merge this branch and have the translation files being automatically updated again.

jeremypw:
> If we need to do another Gtk3 release from this branch, the POT update commands will have to be run and changes from them committed manually. Then once the strings are stabilised, we can merge this branch and have the translation files being automatically updated again.

Looks like @danirabbit is gearing up to do another Gtk release (version 6.5.0) see #2275

danirabbit:
Just trying to track release notes as we go because it's more of a chore to try to write them in retrospect than to include them in pull requests

jeremypw:
> Just trying to track release notes as we go because it's more of a chore to try to write them in retrospect than to include them in pull requests.

Except that can create conflicts with master if all PRs are trying to modify the same part of the metainfo file.

danirabbit:
It's fairly easy to resolve merge conflicts if they happen 🤷🏻‍♀️

danirabbit:
I'm going to reopen this since it seems @jeremypw is still wanting to make regular releases out of main. There really haven't been any major string changes afaict. There hasn't been any big redesigns involved in Gtk4 porting, it's been pretty noninvasive in terms of UI changes. So I'm of the opinion that we should allow translators to keep doing their thing here

jeremypw:
While Gtk4 preparation is still going on in the Gtk3 version, and switching to Gtk4 seems further off than originally envisaged there seems no reason not to release fixes for significant bugs especially where the relevant code will not be affected in the Gtk4 port. Every effort will be made to avoid changing translatable strings.

danirabbit:
Not sure when it happened but weblate is already pointed at main and unfrozen, so gonna go ahead and merge to make sure we have updated translation templates

PR #2225, opened 2023-06-30
Make compile

jeremypw:
This based on `gtk4` branch up to commit "First compilable lint-free version"(https://github.com/elementary/files/commit/be0ce20d8679315424f58b88dff93e225ddce861)`  rebased on `main`  plus a few extra commits to make it compile again.

The changes should be near the minimum to make it compile without removing widgets but still affects a large amount of code unfortunately. 

The PR builds and runs (as `io.elementary.files`) but very little works

PR #2226, opened 2023-06-30
ConnectServerDialog: Gtk4 prep

danirabbit:
Includes changes from #2225 

* Bump copyright header
* Replace Grids with Boxes
* Code style
* Remove unnecessary use of DetailLabel (fields that are never hidden or are hidden by another means)
* Replace remaining `new DetailLabel` with `new_detail_label`
* Use `child` instead of `add` when relevant
* Replace ButtonBox with Box


jeremypw (approved):
I guess there are parts of the `make-compile` branch that will compile under both Gtk3 and Gtk4 so it makes sense to merge these first - thanks.

Code looks good, compiles and runs

PR #2227, opened 2023-06-30
ConnectServerDialog: fix missing window contents and dialog style

danirabbit:


jeremypw (approved):
Code looks good and the dialog (invoked with `<Alt>C`) now has contents!

PR #2228, opened 2023-06-30
FileConflictDialog: Gtk4 prep

danirabbit:
Includes changes from #2225 

* Remove `skip_taskbar_hint`
* Boxes instead of Grids
* Use `child` property where appropriate

### DIFF
--- a/libcore/FileConflictDialog.vala
+++ b/libcore/FileConflictDialog.vala
@@ -60,8 +60,7 @@ public class Files.FileConflictDialog : Granite.MessageDialog {
         Object (
             title: _("File conflict"),
             transient_for: parent,
-            resizable: false,
-            skip_taskbar_hint: true
+            resizable: false
         );
 
         source = Files.File.@get (_source);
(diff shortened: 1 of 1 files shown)


jeremypw:
Would it be better to target this at `make-compile` otherwise it creates conflicts that need resolving and make it compile under Gtk4?

danirabbit:
@jeremypw ideally I’d like to reduce the size of #2225 which is why I’m targeting main. There’s a lot we could do so that we’re reviewing much smaller branches at a time and hopefully avoiding a lot of regressions. And then it’ll be a lot more feasible to actually review #2225 when the changes are smaller

jeremypw:
But this creates extra work keeping #2225 non-conflicting.

I cannot see a way of switching from Gtk3 to Gtk4 in one buildable commit while building all the UI without it being large.  

If you want to work on `main` then we had better close #2225 and start again once work on `main` is done.

danirabbit:
@jeremypw I think having a PR like #2225 is good to be able to see where we're needing to make changes, but there is a really a lot of things we could do to avoid regression and make the port much much smaller. Even if that's like a 3,000 line diff versus a 6,000 line diff I think in the long run this would save us a lot of time trying to hunt down bugs and regressions that could be avoided

jeremypw (approved):

PR #2229, opened 2023-06-30
ViewSwitcher: fix behavior, don't use observe

danirabbit:
* Use `group` so that we get radio behavior
* Granite style class constant
* Docs advise against using `observe_children` when possible because it's expensive

### DIFF
--- a/libcore/Widgets/ViewSwitcher.vala
+++ b/libcore/Widgets/ViewSwitcher.vala
@@ -24,14 +24,12 @@
 namespace Files.View.Chrome {
     public class ViewSwitcher : Gtk.Box {
         public GLib.SimpleAction action { get; construct; }
-        private GLib.ListModel children;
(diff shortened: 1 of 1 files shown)


libcore/Widgets/ViewSwitcher.vala
```diff
@@ -83,12 +83,13 @@ namespace Files.View.Chrome {
...
         public void set_mode (uint32 mode) {
-            for (uint i = 0; i < children.get_n_items (); i++) {
-                var child = children.get_item (i);
+            unowned var child = get_first_child ();
+            while (child != null) {
                 if (child.get_data<uint32> ("id") == mode) {
                     ((Gtk.ToggleButton)child).active = true;
                     action.activate (child.get_data<uint32> ("id"));
                 }
+                child = child.get_next_sibling ();
```
danirabbit, line 92: Should we return early here if we activate a child?

PR #2233, opened 2023-07-01
Move HeaderBar into Window

danirabbit:
Since we can't subclass HeaderBar in Gtk4, this puts the HeaderBar directly into the main window. From what I can tell this actually removes some pointless lasagna so :tada:


jeremypw:
@danirabbit Please fix conflicts.

jeremypw (changes requested):
This seems to introduce some regressions:
1. `<Ctrl>F` shortcut no longer triggers search mode
2. Pressing `<Ctrl>F` after starting rubberband select makes rubberbanding persist until `Esc` is pressed.
3. `<Ctrl><Shift>N` creates new folder but also sends a notification and interferes with rubberbanding.

In fact many/most keyboard shortcuts interfere with rubberbanding now.

danirabbit:
@jeremypw I can't confirm an issue with triggering search mode. Can you try that again?


unless I'm misunderstanding, I can reproduce the rubberbanding issues in main as well. This seems to be related to these lines in main:

```
            top_menu.focus_in_event.connect (() => {
                current_tab.is_frozen = true;
                return true;
            });
            top_menu.focus_out_event.connect (() => {
                current_tab.is_frozen = false;
                return true;
            });
```

jeremypw:
I can now confirm the rubberband issue in master as well - sorry about that. I'll raise an issue, though whether its serious enough to fix in Gtk3 is questionable. 

I have still several issues with this branch however - I'll investigate further before reporting this time.

jeremypw:
Issues noticed so far:
* Browser buttons do not react to click but *do* react to long press
* Clicking on a folder does not navigate
* Pressing `<Ctrl>F` has no apparent effect but ..
* After pressing `<Ctrl>F` and *then* starting a rubberband selection, the rubberbanding does not end on releasing the primary button.   This is different from in main where the issue only occurs if the shortcut is pressed *while* rubberbanding.

It is not obvious from the diff what the problem is but it is something to do with focus and/or event capturing.

jeremypw:
Could be something to do with the timing/order of events changed by collapsing chained signals/signal handlers into one. May need to use `connect_after`??

src/View/Window.vala
```diff
@@ -209,28 +239,30 @@ public class Files.View.Window : Hdy.ApplicationWindow {
...
+        location_bar.focus_file_request.connect ((loc) => {
             current_tab.focus_location_if_in_current_directory (loc, true);
         });
-        top_menu.focus_in_event.connect (() => {
-            current_tab.is_frozen = true;
-            return true;
+
+        location_bar.focus_in_event.connect ((event) => {
+            locked_focus = true;
+            return focus_in_event (event);
```
jeremypw, line 260: Should be `header_bar.focus_in_event (event)`

src/View/Window.vala
```diff
@@ -209,28 +239,30 @@ public class Files.View.Window : Hdy.ApplicationWindow {
...
         });
-        top_menu.focus_out_event.connect (() => {
-            current_tab.is_frozen = false;
-            return true;
+
+        location_bar.focus_out_event.connect ((event) => {
+            locked_focus = false;
+            return focus_out_event (event);
```
jeremypw, line 265: Should be `header_bar.focus_out_event (event)`

jeremypw:
The above changes to the focus event handlers seems to fix the problems!

jeremypw:
Hmm, it seems the browser button problem is still not fixed - the others are.  I'll have another look.

jeremypw:
@danirabbit  I don't think you have connected the browser buttons' `activated` signal?

danirabbit:
@jeremypw oops you're right. Somehow I removed the connections in one of these conflict resolves. I think it's fixed now

jeremypw (approved):
Everything seems to work now.

PR #2237, opened 2023-07-03
RenamerDialog: cleanup and Gtk4 prep

danirabbit:
Screenshot from 2023-07-03 13 57 28
Screenshot from 2023-07-03 13 57 42

* Bump copyright header
* Code style
* Use `child` property where relevant
* Create buttons with labels and icons in a Gtk4 friendly way
* Fix menu appearance for suffix and prefix menubuttons
* Set revealer transitions so they preserve space correctly
* Remove keypress handler for `REJECT`. It doesn't appear that anything listens to this response ID
* Replace grids with boxes
* Pack boxes with `add` where relevant
* Fix margins, aligns, expands, etc
* Use MenuButton for `mod_button`
* Reduce minimum usable width

### DIFF
--- a/src/Dialogs/BulkRenamer/Renamer.vala
+++ b/src/Dialogs/BulkRenamer/Renamer.vala
@@ -34,6 +34,7 @@ public class Files.Renamer : Object {
         listbox = new Gtk.ListBox () {
             can_focus = false,
             selection_mode = Gtk.SelectionMode.NONE,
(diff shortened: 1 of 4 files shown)


jeremypw (approved):
Code looks good.  There are a few minor issues involving some buttons needing to be pressed more than once to work but these are also present in `main` and outside scope of this PR.

PR #2242, opened 2023-07-10
Create AppMenu class

danirabbit:
Just the appmenu part of #2233 since the second commit seems to introduce some issues that need more investigating


jeremypw (approved):
Looks good, no regressions found.

PR #2244, opened 2023-07-10
LocationBar: use refresh action

danirabbit:
Avoid chaining signals


jeremypw (approved):
Looks good, no regressions

PR #2247, opened 2023-07-11
Window: inline namespace

danirabbit:
This was done in #2225, so this should make the diff easier to read as well as make it easier to merge changes etc. Review with hide whitespace. No other changes besides the inline namespace


jeremypw (approved):
I didn't know "hide whitespace" hid so much in these circumstances!

PR #2248, opened 2023-07-11
EmptyTrashJob: rewrite without dialog.run

danirabbit:
Use `dialog.present` and `dialog.response.connect` instead of `dialog.run`


jeremypw (commented):
This PR no longer shows the confirmation dialog when pressing "Empty the Trash" in the Trash view - presumably because the `confirm_trash ()` function has been removed.  Was this the intention?

danirabbit:
Ah nope. thanks for noticing that. Will fix

jeremypw (approved):
On second view I think it is OK to remove that extra confirmation dialog.  A confirmation dialog is already potentially shown when needed (Preferences.get_default ().confirm_trash is `true`)

PR #2249, opened 2023-07-11
Revealer.add → Revealer.child

danirabbit:
Gtk4 prep. Replace instances of `revaler.add ()` with `revealer.child =`


jeremypw (approved):
Loogs good, no regressions

PR #2250, opened 2023-07-11
AbstractPropertiesDialog: Gtk4 prep

danirabbit:
* Granite dialog takes care of `deletable` already
* Doesn't seem to be a good reason not to allow resizing
* `window_position` doesn't exist in Gtk4
* `stack_switcher` is only used in PropertiesWindow so it saves us some abstraction and makes things easier to just create it there. We don't have to worry about hiding and showing etc
* Explicit margins
* Remove unnecessary cast of `get_content_area`
* DRY properties of `header_title` when it's a label
* Simplify emblem overlay and use a box
* inline namespaces while I'm here


jeremypw (approved):
Not sure making the window resizable adds anything at the moment (the layout does not adapt very well) but does no harm I guess.

PR #2251, opened 2023-07-12
Window: don't get position

danirabbit:
We're not able to get window position in Gtk4. It also doesn't appear we do anything with tiling information anyways, so remove all the unused stuff.

Get maximized state from Gtk.Window and make a note about get fullscreen state here in the future as well

### DIFF
--- a/data/schemas/io.elementary.files.gschema.xml
+++ b/data/schemas/io.elementary.files.gschema.xml
@@ -81,11 +81,6 @@
       <summary>Categorie Network expander</summary>
       <description>Expand/Collapse categorie Network</description>
     </key>
-    <key name="window-position" type="(ii)">
-      <default>(-1, -1)</default>
-      <summary>Window position</summary>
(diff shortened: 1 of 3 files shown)


jeremypw (commented):
This PR no longer saves and restores the tiled state of the window - is that the intention?  It seems odd that when closing a tiled Files window, reopening it restores it to the centre of the screen but with the same size as a tiled window.

danirabbit:
@jeremypw afaict we never restored tile state. The tiling state behavior should be the same as in main, but I pushed a commit to only save window size if we're neither maximized nor tiled

jeremypw:
I tried it out it out in main and does appear to work, but by positioning the window so it is pseudo-tiled.  This will not work in Gtk4 :frowning_face: so will have to be accepted.

jeremypw (approved):
I think this is a reasonable solution.

PR #2252, opened 2023-07-12
Window: remove broken key events

danirabbit:
Not sure if this still behaves as intended because I'm not exactly sure what this is intended to do :sweat_smile:


jeremypw:
This appears to no longer work in `main` anyway :disappointed:  It was intended to allow switching the keyboard between the view and the sidebar by using the keyboard only.  I think it can be removed for now. I don't know how many people find this useful but if there is an outcry a better solution in Gtk4 can be sought.

danirabbit:
@jeremypw alright did a removal instead

jeremypw (approved):

PR #2254, opened 2023-07-12
PropertiesWindow: Gtk4 prep

danirabbit:
* Avoid `no_show_all`
* add `pixel_size` where we can't rely on `IconSize` in the future
* explicit `hexpand` and `vexpand`
* Don't use `add` with grid
* show dialog with `present` not `show_all`
* Remove a single-use method
* Code style changes cherried from #2225


jeremypw (approved):
Looks good, no regressions

PR #2256, opened 2023-07-12
Window: DynamicNotebook → Hdy.TabBar

### LINKED ISSUE #1339: When Single Tab, DnD will create a new window (Design)
## Description
When there is a single tab open, if we Drag and Drop it, it will create a new instance. Which actually is a very convenient way to create new windows, but at the same time breaks the logic of "moving tabs", plus it breaks consistent behavior with everything else, including Terminal 

## Reproduce
1. Open Files
2. Have a single Tab open
3. Drag and Drop the tab

**Result:** A new Files window will be created
**Expected:** The window should be moved

## Artifacts
That's a demo that shows Terminal Vs Files tabs behavior: https://youtu.be/FUEuBQbAM6o

However, this bug report is mostly a question whats the proper design here; Because again, I do like we can create new windows from a single tab!!

danirabbit:
Fixes #1339

* Use Hdy.TabBar instead of Granite.DynamicNotebook
* Reorder tab menu items to match Epiphany while we're here
* Double click to add tab and tab bar context menu are gone, doesn't seem to be a safe way to do this. These can't be accessed when the tab bar is full anyways and they duplicate functionality that already exists.

 Todo:
- [ ] Dragging tab into a new window also creates home folder tab


jeremypw:
Should be able to cherry-pick a lot from #2082.  Although that PR uses Adw widgets the interface is practically the same.

danirabbit:
@jeremypw I'm not sure how to solve creating a new window without the default tab. I'd appreciate help there since you know a lot more than I do about how Files creates new windows! Otherwise, I think this is ready for your feedback/review :)

src/View/Window.vala
```diff
@@ -323,78 +351,150 @@ public class Files.View.Window : Hdy.ApplicationWindow {
...
-        sidebar.request_focus.connect (() => {
-            return !current_tab.locked_focus && !locked_focus;
+        var open_tab_section = new Menu ();
+        open_tab_section.append (_("Open in New Window"), "win.tabmenu-new-window");
+        open_tab_section.append (_("Duplicate Tab"), "win.tabmenu-duplicate");
+
+        var close_tab_section = new Menu ();
+        close_tab_section.append (_("Close Tabs to the Right"), "win.tabmenu-close-end");
+        close_tab_section.append (_("Close Other Tabs"), "win.tabmenu-close-others");
+        /// TRANSLATORS: For RTL this should be "to the left"
```
jeremypw, line 450: Misplaced translators comment - refers to to lines above.

danirabbit: Oops thanks for catching that

jeremypw (approved):
Everything seems to work now.  There are new critical warnings when tabs and windows are closed but they can be addressed separately if wished.

PR #2259, opened 2023-07-18
Fix issues creating a new window

jeremypw:
* Do not create default tab unnecessarily
* Remove tab from old window when appropriate
* View background action has same effect as tab action (same shortcut)
* Avoid same name for action and function

### DIFF
--- a/src/View/Window.vala
+++ b/src/View/Window.vala
@@ -360,7 +360,7 @@ public class Files.View.Window : Hdy.ApplicationWindow {
         });
 
         tab_view.create_window.connect (() => {
-            return marlin_app.create_window ().tab_view;
+            return marlin_app.create_empty_window ().tab_view;
         });
 
         tab_view.page_attached.connect ((tab, pos) => {
--- a/src/View/Window.vala
+++ b/src/View/Window.vala
@@ -425,17 +425,17 @@ public class Files.View.Window : Hdy.ApplicationWindow {
         var action_close_end = new SimpleAction ("tabmenu-close-end", null);
         var action_close_others = new SimpleAction ("tabmenu-close-others", null);
         var action_duplicate = new SimpleAction ("tabmenu-duplicate", null);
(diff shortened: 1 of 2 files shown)


src/View/Window.vala
```diff
@@ -425,17 +425,17 @@ public class Files.View.Window : Hdy.ApplicationWindow {
         var action_close_end = new SimpleAction ("tabmenu-close-end", null);
         var action_close_others = new SimpleAction ("tabmenu-close-others", null);
         var action_duplicate = new SimpleAction ("tabmenu-duplicate", null);
-        var action_new_window = new SimpleAction ("tabmenu-new-window", null);
+        var action_move_to_new_window = new SimpleAction ("tabmenu-new-window", null);
```
danirabbit, line 428:
```suggestion
        var action_move_to_new_window = new SimpleAction ("tabmenu-move-to-window", null);
```

src/View/Window.vala
```diff
@@ -425,17 +425,17 @@ public class Files.View.Window : Hdy.ApplicationWindow {
...
         add_action (action_close_end);
         add_action (action_close_others);
         add_action (action_duplicate);
-        add_action (action_new_window);
+        add_action (action_move_to_new_window);
 
         marlin_app.set_accels_for_action ("win.tabmenu-close", {"<Ctrl>W"});
         marlin_app.set_accels_for_action ("win.tabmenu-duplicate", {"<Ctrl><Alt>T"});
-        marlin_app.set_accels_for_action ("win.tabmenu-new-window", {"<Ctrl><Alt>N"});
+        marlin_app.set_accels_for_action ("win.tabmenu-move-to-window", {"<Ctrl><Alt>N"});
```
danirabbit, line 438: This is only for the accel, so you have to change the action name below on line 444 as well

danirabbit (approved):
Perfect, thank you for helping me finish this branch!!

PR #2265, opened 2023-07-20
Fix handling of status changes by git plugin when file content changes

### LINKED ISSUE #2261: Changes to git status not reflected in view until view refreshed
### What Happened?

While a git folder was displayed in Files, one of the tracked files was modified and then committed.  In each case the emblem of the corresponding file icon did not update unless the view was refreshed.

### Steps to Reproduce

1. Open a folder containing git tracked files in Files
2. Open one of the tracked files in Code and edit it
3. Observe emblem does not change unless view refreshed
4. Commit the changes
5. Observe emblem does not change unless view refreshed

### Expected Behavior

The emblems should reflect the git status without refreshing the view.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Partially fixes #2261 

* Fix how git plugin handles `update_file_info` - update git status for single file
* Emblem on new files in working directory as well as index
* Ensure `update_file_info` is called for file changes
* Ensure `update_file_info` called for file-added events even if file already in cache
* Ignore transient outputstream files so they do not trigger calls to `update_file_info`


jeremypw:
Linked issue not fixed after merging `main` - investigating.

jeremypw:
So, it looks like this never solved the entire issue.  It only solves the case where the file content changes, so the emblem will to change to "modified" when there are uncommitted changes and is removed if the uncommitted changes are reverted.  New untracked files will get the appropriate emblem when they appear.

However, the emblem does not change automatically when there is a commit made.  This is because the file itself does not change - it is the git database that changes.   The Code application *does* detect this because it monitors the `.git` folder of the active project.  It would add significant complication to Files to do this as it is not project-centric.  It *may* be necessary to promote the git plugin to main code like in Code.

Not sure if this partial fix is still worth merging or whether to try for a full fix.

jeremypw:
Closing as not complete fix and no reviews

PR #2266, opened 2023-07-21
Fix restoring focused tab position

jeremypw:
Fixes recent regression in restoring the focused tab position.

### DIFF
--- a/src/View/Window.vala
+++ b/src/View/Window.vala
@@ -355,9 +355,7 @@ public class Files.View.Window : Hdy.ApplicationWindow {
 
         tab_view.close_page.connect (tab_view_close_page);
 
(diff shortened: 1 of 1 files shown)


danirabbit:
@jeremypw I'm not sure I understand what this is for. Can you add some reproduction steps?

jeremypw:
Steps to reproduce: 

1. Open Files with at least two tabs
2. Select a tab other than the last (rightmost) one
3. Close Files
4. Re-open Files
5. Observe that the last tab is always focused (selected) regardless of the state when closed.

danirabbit (approved):
Can reproduce and can confirm this fixes it! Nice work and thanks for explaining :)

PR #2267, opened 2023-07-22
Do not lose previous tabs when opened from another app

### LINKED ISSUE #2235: Opening from another apps remove the tabs history
### What Happened?

Opening folders from another application like Firefox, e.g. when clicking on Show in Folder for downloaded files launches the elementary Files in a blank state, forgetting all the previous tabs opened in the last session.

### Steps to Reproduce

1. Open Firefox or other browsers, whatever
2. Download something safe from the internet 
3. Click Show in Folder

### Expected Behavior

elementary Files should restore all the tabs from the last session instead of opening just the Download folder.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
FIxes #2235 

* Only save tabs on closing if tabs were originally restored (or restore attempted) i.e. `restored_tabs` is `true`
* If `restore-tabs` setting or `remember-history` changes while open then honor it on closing (first window only)
* Use `-n` option with FileManager1 interface so other apps always open a new Files window which does not save or restore its tabs

Before testing this PR must be installed and any pre-existing pantheon-files-daemon process ended.

To test:

1 Open Files (with History and retore-tabs both ON and open several tabs 
2 Close Files
3 Use another app with a "Show in Folder" function (e.g. a brower's downloads dialog) to open a folder.
4 Only the requested folder should appear in a Files window
5. Close FIles and then re-open normally
6. The tabs saved in step 2 should be restored (in `main` there are overwritten)
7.  Leaving the Files window open, use another app to show a folder as before
8. A second Files window appears showing only the requested folder (in `main` the requested folder is added to the first window)

### DIFF
--- a/pantheon-files-daemon/FileManager1.vala
+++ b/pantheon-files-daemon/FileManager1.vala
@@ -37,9 +37,10 @@ public class FileManager1 : Object {
 
     private void open_items_and_folders (string[] uris, string startup_id) throws DBusError, IOError {
         /* The io.elementary.files app will open folder uris as view, other items will cause the parent folder
-         * to open and the item be selected.  Each view will open in a separate tab in one window */
+         * to open and the item be selected.  Each view will open in a separate tab in a new window   They 
+         * will not be remembered or restored */
 
-        StringBuilder sb = new StringBuilder ("io.elementary.files -t");
+        StringBuilder sb = new StringBuilder ("io.elementary.files -n");
         foreach (string s in uris) {
                 sb.append (" ");
                 sb.append (FileManager1.prepare_uri_for_appinfo_create (s));
--- a/src/View/Window.vala
+++ b/src/View/Window.vala
@@ -160,7 +160,7 @@ public class Files.View.Window : Hdy.ApplicationWindow {
         default_height = height;
 
         if (is_first_window) {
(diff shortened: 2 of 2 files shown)


jlnr:
As far as I can tell, this is different behavior from every browser that I have used, where opening new links always restored previous tabs (if any). If I encountered this behavior in Files, I would probably assume that my previous list of tabs was gone and open them manually, instead of closing and re-opening the window to get them back. So tl;dr I think I would prefer to drop the -t flag instead, as you briefly suggested in #2235.

Just my two cents, but I am not on the UX team and I can't even get Horus to work on my Nvidia machine to test anything ☹️

jeremypw:
@jlnr Thanks for your input.  So, I'll open an alternative PR with the behaviour described - that is, opening a folder in Files from another app always opens it in a window that has the saved tabs restored - and saves the new tab when the window is closed (unless that tab is closed first).

That leaves the question whether there is any use for an option to open a folder in Files as an independent event i.e. the only purpose is to operate on that folder so opening other tabs or saving the opened folder is not needed.  I guess this is what the `-n` option already does.

jeremypw:
Closing as there have been two negative comments on this behaviour.

PR #2270, opened 2023-07-26
Do not lose tabs when opened from another app (alternative).

### LINKED ISSUE #2235: Opening from another apps remove the tabs history

jeremypw:
Fixes #2235

This is an alternative to #2267.  In this PR, the FileManager1 interface restores (and saves) tabs instead opening the folder in a new window with no restore/save.

Using the `-n` option now causes a window to open that does not restore or save tabs. Only the first opened window restores and saves its tabs.  This prevents a second window overwriting saved tabs from the main window if closed last.

As in the alternative this PR also handles settings changes while the first window is open (this could be split out into a separate PR if desired). 

Before testing this PR must be installed and any pre-existing pantheon-files-daemon process ended.

To test:

1 Open Files (with History and retore-tabs both ON and open several tabs 
2 Close Files
3 Use another app with a "Show in Folder" function (e.g. a brower's downloads dialog) to open a folder.
4 The saved tabs *and* the requested folder should appear in a Files window
5. Close FIles and then re-open normally
6. The tabs saved in step 2 *and* the requested folder should be restored (in `main` there are overwritten)
7.  Leaving the Files window open, use another app to show a folder as before
8.  The requested folder is added to the first window as in `main`
9. In a terminal open Files with a command like `io.elementary.files -n <path-to-folder>`
10. A window opens only containing the requested folder.  When this window is closed, saved tabs are not overwritten.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -155,7 +155,7 @@ public class Files.Application : Gtk.Application {
         options [1] = { "tab", 't', 0, OptionArg.NONE, ref open_in_tab,
                         N_("Open one or more URIs, each in their own tab"), null };
         options [2] = { "new-window", 'n', 0, OptionArg.NONE, out create_new_window,
-                        N_("New Window"), null };
+                        N_("Open a New Window that does not restore or save tabs"), null };
         options [3] = { "quit", 'q', 0, OptionArg.NONE, ref kill_shell,
                         N_("Quit Files"), null };
         options [4] = { "debug", 'd', 0, OptionArg.NONE, ref debug,
@@ -220,14 +220,16 @@ public class Files.Application : Gtk.Application {
         /* Open application */
         if (files != null) {
             if (create_new_window || window_count == 0) {
-                /* Open window with tabs at each requested location. */
-                create_window_with_tabs (files);
-            } else {
+                /* Open window with tabs at each requested location 
+                 * If the -n option is given, no restoring or saving of tabs occurs. */
+                create_window_with_tabs (files, ViewMode.PREFERRED, create_new_window);
+            } else { // window_count must be > 0
                 var win = (View.Window)(get_active_window ());
                 win.open_tabs (files, ViewMode.PREFERRED, true); /* Ignore if duplicate tab in existing window */
             }
         } else if (create_new_window || window_count == 0) {
-            create_window_with_tabs ();
+            /* A default tab is created */
+            create_window_with_tabs (null, ViewMode.PREFERRED, create_new_window);
         }
 
         if (window_count > 0) {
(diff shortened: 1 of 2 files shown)


Marukesu:
> Using the `-n` option now causes a window to open that does not restore or save tabs. Only the first opened window restores and saves its tabs. This prevents a second window overwriting saved tabs from the main window if closed last.

I believe the rationale for state saving is that the application will keep their last state. saving the state only of the first opened window seems wrong, since if another window is opened and last longer we will lose all the state after the first window is closed.

Reading the linked issue, the problem comes from `-t` not restoring the state when it's creating the window, what IMHO seems wrong, what the reason of having `-t` not restore the state in this case? terminal does restore the state when using it, code seems to act as the same as if `-t` wasn't passed, though i have a lot of non-saved files in it that could be interfering.

jeremypw:
> I believe the rationale for state saving is that the application will keep their last state. saving the state only of the first opened window seems wrong, since if another window is opened and last longer we will lose all the state after the first window is closed.

My rationale for doing it this way is that the first window you open is for your "normal" workflow and additional windows are for short-lived one-off operations that do not need remembering.  It is possible that a second window could be opened on a separate workspace, "forgotten" about and left open while the first window is closed.  It could then overwrite your "normal" restored tabs.  

> Reading the linked issue, the problem comes from -t not restoring the state when it's creating the window, what IMHO seems wrong, what the reason of having -t not restore the state in this case? terminal does restore the state when using it, code seems to act as the same as if -t wasn't passed, though i have a lot of non-saved files in it that could be interfering.

That is what this PR implements.  The `-t` option (and no option) restores tabs and appends the new tabs.

I guess a full solution would be to save and restore *all* windows and their tabs separately, but how many people want that?

Marukesu:
> My rationale for doing it this way is that the first window you open is for your "normal" workflow and additional windows are for short-lived one-off operations that do not need remembering. It is possible that a second window could be opened on a separate workspace, "forgotten" about and left open while the first window is closed. It could then overwrite your "normal" restored tabs.

both code and terminal keep the last closed window state, instead of the first opened one, that ensures that whatever state is saved matches the last state saw by the user.

a user can always choose to close the assumed "normal" workflow window instead of the assumed short-lived one because it is interested in the folder open in that window. after that files will ignore the state for the rest of their session using that window because it choose the wrong window to keep open.

if windows opened with the `-n` flag is meant to behave differently than windows without it, then they got to look different too. otherwise for the user they are all the same type of window and got to behave equally.

jeremypw:
I can confirm that both Code and Terminal behave in the way you suggest and I am not aware of any complaints.  I think consistency is more important so I'll change Files behaviour to match.  Looks like neither Nautilus nor Dolphin restore tabs anyway so cannot compare with those apps.

jeremypw:
Closing in favour of new PR.

Issue #1537, opened 2020-12-14
Colour label backgrounds on file/dir names look fugly in


hanaral:
### Problem

Currently using a colour label on a file or folder draws a solid, dark colour around the black text (in light mode). This is not very accessible, since the contrast isn’t too great. It also looks super out of place with the new stylesheet that seems to prefer dark text having light colour outlines (such as suggested action buttons). 

### Solution:

Make it draw a lighter, transparent background 
OR
Make the name text white under all circumstances of it has a colour label applied to it

jeremypw:
Needs UX team input as to how best to show color tags.  At the moment, the text color is set by using the contrasting foreground color returned by the Granite library for the color beneath the text.  That is why the text on color tags is (at the moment) always black regardless whether light or dark variant is used because the color tags do not change.  To me, the text is more readable using the dark variant especially for the darker tag colors but I think that is an optical illusion caused by the bright white folder background in the light variant.

Other ways of showing the tag color could be considered e.g. a colored frame or a separate tag emblem.

jeremypw:
I have confirmed that, with an experimental black color tag the text does indeed change to white.  However, the legibility is still not good on my screen.  Maybe the weight of the font should be increased??

jeremypw:
Examples for elementary 6.0 master (+ black tag)

![Screenshot from 2020-12-15 05 31 34](https://user-images.githubusercontent.com/10513844/102107791-064a1900-3e2a-11eb-834b-4192d7ce76dc.png)
![Screenshot from 2020-12-15 05 35 54](https://user-images.githubusercontent.com/10513844/102108396-c6376600-3e2a-11eb-93e5-ce207a7ef019.png)

jeremypw:
Maybe this can be considered during or after porting to Gtk4 dynamic views?

PR #2273, opened 2023-07-30
Store color-tag in metadata instead of SQL database

### LINKED ISSUE #491: Color tag does not restored when the file or folder is untrashed
Involved #488.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58491386-color-tag-does-not-restored-when-the-file-or-folder-is-untrashed?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1208: Wishlist idea: use extended file attributes for tags
## Prerequisites

## Extended file attributes for colour tags

This is a more of a wacky proposal than anything else (and yes, it's a wishlist sort of idea), so I'll keep this somewhat brief (and I sadly lost a previous version of this issue draft...) and I'll list the advantages and disadvantages of this approach, but I have a feeling the UX team may like this idea. Maybe also the people working on Files' codebase.

(Also, if you wonder why I refer to tags in general sometimes in this proposal it's because I have a "named/text" tags idea in mind for Files, I might create a separate issue for this at some point)

### Current approach

So basically, the current approach that Files uses to know what files have what colour tags is with a SQL database. Currently, moving a file somewhere else, even within Files, does not preserve its colour tags. I don't think Files even attempts to track the file currently. This could possibly be fixed/implemented for moving or renaming files within Files (updating the SQL file when you move a file within Files). However, this would not work if you did this outside of Files:
```
$ mv ~/myfile ~/somewhereelse/myfile
```
When you then open Files again it wouldn't know that `~/somewhereelse/myfile` is just `~/myfile` moved somewhere else.

Guessing would be highly error-prone and impractical, as we could've easily moved `myfile` to `~/somewhereelse/myfile` and then made a new different file called `myfile` where the other one was previously. Even worse, what if we renamed `myfile` externally? Files couldn't possibly know with absolute certainty that it's still `myfile` with a different name, because we could've made changes to it since then.

A heuristic or comparing algorithm would make Files slower due to all the potential file comparisons and directories we could check in a hypothetically large directory tree, and it'd be cumbersome to think up some sort of efficient algorithm for it and implement it.

What about keeping track of file changes with an external daemon? If you used `inotify` you'd probably run out of kernel watches quickly as you have to _watch every single directory individually_ with `inotify`. Even if you just limited colour tag tracking to the home directory of the current user, used a blacklist for folders like `.local`, `.config`, etc. and extended the watch limit to a really large number, it could potentially take ages to set up all these watches, every single login, for every single non-blacklisted folder in the current user's home directory... etc, etc...

It seems pretty hopeless and utterly cumbersome to be able to practically track metadata with files in this manner.

### Enter extended file attributes

But what if could instead tell the filesystem to keep track of our tag metadata for us instead of trying to manually track it? This is exactly what extended file attributes, or `xattr`s, are for! Let me demonstrate (this is a real example):
<details>
<summary>xattr example</summary>

```
$ mkdir somefolder
$ touch mypinkfiletaggedhello
$ attr -l mypinkfiletaggedhello
(no output)
$ attr
A filename to operate on is required
Usage: attr [-LRSq] -s attrname [-V attrvalue] pathname  # set value
       attr [-LRSq] -g attrname pathname                 # get value
       attr [-LRSq] -r attrname pathname                 # remove attr
       attr [-LRq]  -l pathname                          # list attrs 
      -s reads a value from stdin and -g writes a value to stdout

$ attr -s tag.color.name -V "pink" mypinkfiletaggedhello
Attribute "tag.color.name" set to a 4 byte value for mypinkfiletaggedhello:
pink

$ attr -s tag.color.raw -V "#fe9ab8" mypinkfiletaggedhello
Attribute "tag.color.raw" set to a 7 byte value for mypinkfiletaggedhello:
#fe9ab8

$ attr -s tag.names -V '["hello", "some other tag", "files I like"]' mypinkfiletaggedhello
Attribute "tag.names" set to a 43 byte value for mypinkfiletaggedhello:
["hello", "some other tag", "files I like"]

$ attr -l mypinkfiletaggedhello
Attribute "tag.color.name" has a 4 byte value for mypinkfiletaggedhello
Attribute "tag.color.raw" has a 7 byte value for mypinkfiletaggedhello
Attribute "tag.names" has a 43 byte value for mypinkfiletaggedhello

$ exa -lh --group-directories-first
Permissions Size User Date Modified Name
drwxrwxr-x     - me    1 Feb  2:30  somefolder
.rw-rw-r--@    0 me    1 Feb  2:30  mypinkfiletaggedhello

$ mv mypinkfiletaggedhello somefolder/
$ exa -lh --group-directories-first somefolder/
Permissions Size User Date Modified Name
.rw-rw-r--@    0 me    1 Feb  2:30  mypinkfiletaggedhello

$ attr -l somefolder/mypinkfiletaggedhello
Attribute "tag.color.name" has a 4 byte value for somefolder/mypinkfiletaggedhello
Attribute "tag.color.raw" has a 7 byte value for somefolder/mypinkfiletaggedhello
Attribute "tag.names" has a 43 byte value for somefolder/mypinkfiletaggedhello
```

(By the way, `tags.color.name` is an entirely arbitrary attribute name, it doesn't need any dots or anything, it can even look like this:)
```
$ attr -s "my pancakes" -V "abc" somefolder/mypinkfiletaggedhello
Attribute "my pancakes" set to a 3 byte value for somefolder/mypinkfiletaggedhello:
abc
```
---

</details>

So instead of trying to track a file's tags manually, we can simply use xattrs to store the file's tags instead.

With this, files would now automagically have their colour tags preserved if you moved them, renamed them, or copied them, even if you did it outside of Files or with some other file manager.

#### Advantages:
  - Tags for files can now be stored within the filesystem: we don't have to manually track a file's tags anymore, the filesystem can do this better than we can
  - We don't have to do anything special for moving and renaming files and their tags will be kept (see disadvantages regarding copying files)
  - Simple to get a file's tags or any other stored metadata for it: just read the file's xattrs
  - xattr names and values are UTF-8, so e.g. name/text tags would be compatible with non-English languages:
    ```
    $ attr -s hello.あいうえお -V "こんにちは！元気ですか？" somefolder/mypinkfiletaggedhello
    Attribute "hello.あいうえお" set to a 36 byte value for somefolder/mypinkfiletaggedhello:
    こんにちは！元気ですか？
    ```
  - No more need for a SQL database
  - Could potentially be standardised, meaning much more interoperability with other programs, file managers, desktop environments and OSes (in comparison, Files' SQL database is highly specific to Files)

#### Not really issues but should probably take note:
  - Underlying filesystems would have to explicitly support extended attributes in order for this to work. However as it would seem [many, many filesystems support it](https://en.wikipedia.org/wiki/Extended_file_attributes#Implementations), not just common gnu+linux ones like ext4, but even Windows' NTFS and macOS's HFS+, even FAT(!?), this is probably not an issue
  - Apparently you used to have to explicitly enable xattrs on gnu+linux systems with the option `user_xattr` on mounts in `fstab`, but that doesn't seem to be the case anymore. My `/etc/fstab` file doesn't have `user_xattr` anywhere in the file and yet xattrs work fine here (yes, on elementary 5.1). So this doesn't seem to be an issue either?

#### Disadvantages:
  - Certain archiving formats and syncing programs (e.g. syncthing) currently decide to not preserve xattrs or otherwise ignore them, so tags would be lost if you zipped or synced with those particular programs
    * However they wouldn't be preserved with Files' current SQL database method anyway, so Files' current method doesn't have an advantage over xattrs here
  - ~~I couldn't find any Vala bindings to GNU's xattr interface, so you'd probably have to write/generate bindings for it. See `/usr/include/x86_64-linux-gnu/sys/xattr.h` for the header file. It's not very long though and you could probably leave out the file descriptor ones, but if you're not particularly familiar with C (...I'm not) it will likely give you a big headache trying to do this~~
    Vala has a way to manipulate xattrs, so this would not be required
  - You may have to explicitly set the option to preserve extended attributes when copying files, e.g. for `cp` you have to use `cp --preserve=xattr file1 file2` in order to preserve xattrs

I think, baring any other issues I'm unaware of, this could have a lot of potential.

### LINKED ISSUE #2262: Thumbnail not shown for added file until view refreshed
### What Happened?

Pasted a cut image file into a folder.  It was given a mime-type icon which was not thumbnailed until the view was refreshed (a copied image was thumbailed)

### Steps to Reproduce

1. Cut a (thumbnailed) image file from a folder
2. Paste it into another folder
3. Observe no thumbnailing til refreshed

### Expected Behavior

Newly inserted image files should be thumbnailed if they are visible.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

### LINKED ISSUE #2263: Thumbnails not displayed on startup until refreshed
### What Happened?

Started files with an image folder being the restored, focused location.  The image files were not thumbnailed til the view was refreshed.   However, navigating into the same folder did immediately display thumbnails.

### Steps to Reproduce

1. Display twp image containing folders in two tabs
2. Close Files
3. Reopen Files (with restore tabs ON)
4. The focused folder is displayed without thumbnails
5. Navigate away from the folder then back
6. The thumbnails are displayed. 
7. Switch tabs
8. The other folder is still not thumbnailed

### Expected Behavior

The thumbnails are displayed in folders that are restored on startup or when the view is realized.

### OS Version

9.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1208 
Fixes #491
Fixes #2262
Fixes #2263

Instead of using a complicated SQL database to store certain data about a file, the color tag data (the only data actually used) is stored as a `metadata::color-tag` file attribute.   Migration is addressed by modifying the plugin to still query the database and set the metadata accordingly, on a one-time basis after which the entry is deleted, and not to store any more entries in the database.   Eventually the database and the associated Dbus service will be removed.

Most restrictions regarding which items can be color tagged are removed.  Items in Trash cannot be tagged but items in `recent://` and `network://` and with remote schemes can be.

In order to test this, it must be installed and any existing pantheon-files-daemon process ended.


jeremypw:
Interestingly, using cut/paste to move a color-tagged file to the Trash and to restore it *does* preserve the color-tag so there is hope of fixing the issue after all.  Copying a file does *not* currently copy the color-tag.  Drag and drop between local folders (i.e. moving it) *does* preserver the color-tag.

jeremypw:
Looking again at #1208, there are a number of possible drawbacks with this approach that may need to be addressed in this or future PRs:

 * The color-tag is associated with the file not the user so will be the same for everyone unless we can include user info in the tag
 * You may not want to send the color-tag when sending the file to someone else.
 * The tag might better be a color name rather than an index into a hard-coded array

jeremypw:
> The color-tag is associated with the file not the user so will be the same for everyone unless we can include user info in the tag

My view on this is that a file that is being shared/worked on by several people should not be tagged anyway or should be tagged with an agreed color or color-syntax.  Wanting different color is like wanting different content for the same file!

jeremypw:
> You may not want to send the color-tag when sending the file to someone else.

This is easily worked around by sending a *copy* of the file (if metadata would be sent anyway)

jeremypw:
> The tag might better be a color name rather than an index into a hard-coded array

This can be implemented in the future if desired but is only of benefit if there is standardisation of color-tagging between different apps.

jeremypw:
Further testing indicates that this does fix #491 so above comments amended accordingly.

jeremypw:
@danirabbit Do you think it is a major problem for color tags to no longer be personal but be an attribute of the file itself?  It would be possible for each user to apply and retreive their own tags e.g. `metadata::color-tag-<userid>` I guess but I am not sure this is worth the extra complexity.

jeremypw:
Just confirmed that sending a file to a remote filesystem (SMB) using copy/paste, cut/paste, or DnD does *not* transfer the color-tag.

jeremypw:
Restoring color-tag from Trash seems to have stopped working today for no apparent reason - investigating.

jeremypw:
> Restoring color-tag from Trash seems to have stopped working today for no apparent reason - investigating.

This appears to be due to a timing issue where the view was not always refreshing the color-tags on restore.  They were in fact surviving trash/restore but were not always visible.  Recent commits should fix that.

danirabbit (commented):
I'm having trouble with the migration. This is losing color tags for me:

1. Create color tags for a file and a folder
2. kill files-daemon and install this branch
3. Open files and see color tags are present
4. Refresh, color tags are now gone

jeremypw:
@danirabbit Thanks for noticing that. It did work initially but subsequent refactoring introduced a regression that I should have noticed.  Should be fixed now.

jeremypw:
@danirabbit While we are looking at color-tags, is there any mileage in the idea to reflect color-tags on bookmarked folders in the sidebar as well? Otherwise the related issue #1002 can be closed as "Design conflict".
When we rewrite the view in Gtk4 we should take the opportunity to reconsider how color-tags are displayed too (see #1537)

danirabbit:
@jeremypw Yeah that's probably something to revisit after we finish the Gtk4 port for sure to try to make them more relevant and useful

danirabbit (approved):
Nice, everything works as expected and we love a massive code reduction!

PR #2279, opened 2023-08-06
Do not add transient files to cache or model

### LINKED ISSUE #2344: Transient outputstream files appear but do not disappear until refreshed
### What Happened?

While viewing a folder containing a file being edited in Code, multiple files in the form `goutputstream-XXXX` appeared.  Only if the view was refreshed  did they disappear again.

### Steps to Reproduce

 - Navigate to a folder containing a text file
 - Set "Show Hidden" ON.
 - Open the text file in Code
 - Edit the file with autosave on
 - Observe persistent transient files

### Expected Behavior

Transient files should be ignored

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #2344

Instead of trying to get fileinfo this PR simply ignores files named ".goutputstream..." and they are not inserted in the Directory file cache nor added to view models. 

 Not sure if the name filter is optimal or whether other filename forms should also be ignored.  Is there a better (and synchronous) way of immediately determining files that should be ignored?


libcore/Directory.vala
```diff
@@ -860,7 +860,16 @@ public class Files.Directory : Object {
         file_hash.insert (gof.location, gof);
     }
 
-    private Files.File file_cache_find_or_insert (GLib.File file, out bool was_in_hash, bool update_hash = false) {
+    private Files.File? file_cache_find_or_insert (
+        GLib.File file,
+        out bool was_in_hash,
+        bool update_hash
```
zeebok, line 866: How come you want to remove the default false value here?

zeebok (approved):

libcore/Directory.vala
jeremypw: Not sure now, but I am not in general a fan of default parameters. I probably wanted to ensure a conscious decision from the dev in this case.

jeremypw:
Closing as issue seems to have resolved in latest source.

PR #2280, opened 2023-08-09
Filemanager1 scroll to item

### LINKED ISSUE #2214: When opened with specific file to select, the view is not showing that file
### What Happened?

When i select "Open in File Browser..." in [Rapid Photo Downloader](https://damonlynch.net/rapid/), Files is opened and the file is selected, but it is not shown in the view. I have to scroll down to see it.

Demonstration Video: https://www.youtube.com/watch?v=QoJYM3_LbUE

Rapid Photo Downloader uses the Python Library [show-in-file-manager](https://github.com/damonlynch/showinfilemanager) to do it, which basically does:

`io.elementary.files FILE`

With that, the issue is easily reproducible.

### Steps to Reproduce

0. Open Terminal
1. Open Files
2. Copy this command in Terminal `io.elementary.files`, add a Whitespace
3. Scroll down in a folder with many files in Files and drag-and-drop a file from far down into the terminal, to get it's full path
4. Close Files
5. Run the command in Terminal, e.g. `io.elementary.files '/home/davidak/2023-06-10_14-28-09.mkv'`
6. Files is opened and you see the folder the selected file is in, you have to manually scroll down to see it

### Expected Behavior

The view is scrolled automatically to show the selected file.

### OS Version

Other Linux
NixOS 22.11.4479.d4a9ff82fc1

### Software Version

Older release (I have not run all updates)
elementary-files-6.2.1

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #2214 

In order to make the behaviour that expected and the same regardless of whether Files is already running or not when the request is made it was necessary to fix several problems, in particular some races occurring when Files was starting up.

 - [x] Fix identifying whether a requested location was a child of a tab
 - [x] Fix AbstractDirectory function selecting files race with ffocus file on startup
 - [x] ViewContainer function focusing a location waits for loading to complete if called at startup

To test:

Call Files from the command line with a non-directory file parameter that is a child of one of the tabs that is restored on startup, on that contains a lot of files and requires scrolling to view the requested child.

Do similar where requested tab is not restored.

Do similar when Files is already running and is/is not showing the requested tab.

Use "Open in Folder" function of browser or other app with similar function. 

In each case Files should show the requested item selected and visible.

Check for regressions where the requested item is a folder.  The requested folder should open in a new tab if it is not already open, even if it is a child of an existing tab.


jeremypw:
Needs further testing following change to async adding tabs.

jeremypw:
There is some uncertainty about when Files should create duplicate a tab when the requested folder is already showing (has been restored).

zeebok:
The code itself looks fine to me. I can try to fit in time to test this over the weekend, but if you feel you tested enough already I don't mind approving

zeebok:
I am testing on eOS7:

> Call Files from the command line with a non-directory file parameter that is a child of one of the tabs that is restored on startup, on that contains a lot of files and requires scrolling to view the requested child.
Do similar where requested tab is not restored.

Both seem to always open a new tab, but it *does* scroll to the correct file with it already selected. Is it creating the tab before restoring or something?

> Do similar when Files is already running and is/is not showing the requested tab.

This behaves as expected. Creates a tab if needed or switches to an existing one of the correct directory with desired file selected

> Use "Open in Folder" function of browser or other app with similar function.

Works as expected I think. Only situation of note is that if Files is closed, Open in will open a new tab even if the correct folder is a restored tab. (ie Downloads/ tab will be restored *and* a new Downloads/ tab will open from the browser) similar to the issue I saw from terminal.

> Check for regressions where the requested item is a folder. The requested folder should open in a new tab if it is not already open, even if it is a child of an existing tab.

This looks to work alright

jeremypw:
@zeebok Are you able to approve this for merging? You had a look some time ago.

zeebok (approved):
The only thing that you never responded to was the behavior of a new tab always being created. If this isn't a concern then go ahead and merge

jeremypw:
@zeebok Ah, I overlooked/forgot that - I havent looked at this for a while.  I think it could be annoying if a series of tabs at the same location got opened so I'll fix it so that doesn't happen (if possible).

jeremypw:
@zeebok On second thoughts, the issue you mentioned only seems to occur if Files is not already running and the requested folder/file was saved for restoration.  If Files is already running when the request is made the duplicate tabs are not created.  I think there has been some discussion in the past (maybe in the context of Terminal) about the correct behaviour when a requested folder is amongst those that are going to be restored and I *think* the current wisdom is that the restored folders should be independent of the requested folder(s).  It is certainly easier that way as folders are restored asynchronously.
So I'll merge for now and see if any issues are raised.

PR #2281, opened 2023-08-11
Gtk4Prep: Rewrite ViewSwitcher

jeremypw:
Towards #1941

* Do not iterate container children - keep list (compatible with Gtk4)
* Lose unnecessary signal - use action
* Lose unnecessary data - use action target
* Lose unnecessary action name parameter - never changes
* Lose Chrome namespace
* DRY

Only one or two lines need changing to compile under Gtk4. Some lines can be saved under Gtk4 by assigning ToggleButtons to a group instead of implementing radiobutton behaviour explicitly.

### DIFF
--- a/libcore/Widgets/ViewSwitcher.vala
+++ b/libcore/Widgets/ViewSwitcher.vala
@@ -1,10 +1,5 @@
 /***
-    ViewSwicher.cs
-
-    Authors:
-       mathijshenquet <mathijs.henquet@gmail.com>
-       ammonkey <am.monkeyd@gmail.com>
-
+*   Copyright (c) 2016-2023 elementary LLC. <https://elementary.io>
     Copyright (c) 2010 mathijshenquet
 
     This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


libcore/Widgets/ViewSwitcher.vala
```diff
@@ -19,78 +14,95 @@
...
-    public class ViewSwitcher : Gtk.Box {
-        public GLib.SimpleAction action { get; construct; }
+public class Files.ViewSwitcher : Gtk.Box {
+    private const string ACTION_VIEW_MODE = "win.view-mode";
+    // Keep a list of children to iterate - compatible with both Gtk3 and Gtk4
+    private GLib.List<Gtk.ToggleButton> children = null;
+
+    construct {
+        valign = Gtk.Align.CENTER;
+        get_style_context ().add_class ("linked");
```
danirabbit, line 30: Should use style constants
```suggestion
        get_style_context ().add_class (Gtk.STYLE_CLASS_LINKED);
```

danirabbit (changes requested):
This consistently crashes when using keyboard shortcuts. It's also already a very small diff between `main` and `make-compile` for this class already

PR #2282, opened 2023-08-12
Rewrite ButtonWithMenu nearer to Gtk4 style

jeremypw:
Towards #1941

Backported as far as possible from the version in the `gtk4` branch

 * Base on Widget with Bin layout (Used EventBox for Gtk3)
 * Use GestureLongPress
 * Move some code from Window
 * Lose View.Chrome namespace

One drawback is that in Gtk3, the image does not dim when the widget is set insensitive whereas in Gtk4 it does.  Suggestions for a simple fix welcome.

### DIFF
--- a/libcore/Widgets/ButtonWithMenu.vala
+++ b/libcore/Widgets/ButtonWithMenu.vala
@@ -14,7 +14,9 @@ public class Files.ButtonWithMenu : Gtk.EventBox {
     }
 
     construct {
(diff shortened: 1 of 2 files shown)


danirabbit:
This doesn't seem to work correctly at all. Whats the motivation here? The diff between `main` and `make-compile` for this class is already only like 20 lines and it already uses `Gtk.Gesture`. Using Popover menus in Gtk3 has lots of style issues, and it's already nearly a drop-in replacement to use `Gtk.PopoverMenu`

PR #2283, opened 2023-08-13
Gtk4Prep:  Replace Granite.Widgets.Welcome

jeremypw:
Towards #1941 

 * Implement Gtk3 equivalent of Granite.Placeholder
 * Adapt other code to use Placeholder instead of Welcome
 * Remove Welcome

It should be possible to drop Granite.Placeholder in place of Files.Placeholder when porting to Gtk4.  Some change of appearance is occasioned by this replacement, no attempt was made to reproduce exact appearance in `main`.  This can be done in a separate PR if required.

Most of the increase in code lines is due to reformatting with shorter lines.


jeremypw:
###  BEFORE
Screenshot from 2023-08-13 13 47 31

###  AFTER
Screenshot from 2023-08-13 13 46 53

danirabbit (commented):
What's the motivation here? This doesn't seem worth doing ahead of time since it's a very small API change and you're having to add a whole new class here. This is also missing all the CSS from Granite 7 to make this styled correctly etc.

PR #2284, opened 2023-08-13
SidebarWindow: gtk4 prep

danirabbit:
* Update copyright header
* Replace grids with boxes
* Explicit variable scope
* Use `child` instead of `add` where appropriate
* Construct connect to server button in a Gtk4 friendly way
* Use action name and target instead of connecting button through signals etc
* Hide the action bar without using `no_show_all` which isn't available in Gtk4


jeremypw (approved):
Code looks good, works as normal

PR #2285, opened 2023-08-14
Replace ScrolledWindow.add with child property

danirabbit:
Gtk4 prep. Remove an unnecessary frame in `src/View/Widgets/SearchResults.vala` while here, borders are drawn with CSS


jeremypw (approved):
There is no guarantee that AbstractDirectoryView will survive the port to Gtk4 especially if dynamic views are used - the `gtk4` branch replaces it entirely with a *much* simpler interface based on dynamic views.

However these changes are harmless.

PR #2286, opened 2023-08-14
Sidebar Gtk4 Prep

danirabbit:
We can't subclass ListBox in Gtk4 so rework things for that


jeremypw:
Did you check out the solution in the `gtk4` branch, which avoids repeating code in each listbox?

danirabbit:
I'll need to look at those more closely! I tried a few different things yesterday and wasn't totally satisfied with any direction I was going 😅

jeremypw:
Yes, the `gtk4` branch was intended to provide a possible destination but I am unclear as to what steps along the way would be acceptable.

danirabbit:
Closing in favor of #2287

PR #2287, opened 2023-08-15
Sidebar Gtk4 Prep

danirabbit:
Starting over based on changes from #2082

I think this is The Least™ to make sure these don't subclass ListBox without any errors or warnings in terminal etc

`SidebarInterface`, `SidebarWindow`
* Remove unused `remove_item_by_id`

`SidebarItemInterface`
* subclasses Object rather than Gtk.Widget

`SidebarListInterface`
* subclasses Object rather than Gtk.Widget
* remove unused `remove_bookmark_by_uri`
* Use `Gtk.ListBoxRow` instead of `SidebarItemInterface` where possible to avoid issues with pointers/casting
* Misc code style

`BookmarkListBox`, `DeviceListBox`, `NetworkListBox`
* Subclass Box, ListBox is a child
* Use `Gtk.ListBoxRow` instead of `SidebarItemInterface` where possible to avoid issues with pointers/casting
* Misc code style

### DIFF
--- a/libcore/Interfaces/SidebarListInterface.vala
+++ b/libcore/Interfaces/SidebarListInterface.vala
@@ -9,7 +9,7 @@ public interface Sidebar.SidebarListInterface : Object {
     public abstract Files.SidebarInterface sidebar { get; construct; }
     public abstract Gtk.ListBox list_box { get; internal set; }
 
-    public abstract void select_item (SidebarItemInterface? item);
+    public abstract void select_item (Gtk.ListBoxRow? item);
     public abstract void unselect_all_items ();
 
     public virtual void open_item (SidebarItemInterface item, Files.OpenFlag flag = DEFAULT) {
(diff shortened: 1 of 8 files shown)


libcore/Interfaces/SidebarListInterface.vala
```diff
@@ -1,69 +1,46 @@
...
     public virtual void clear () {
-        foreach (Gtk.Widget child in get_children ()) {
-            remove (child);
+        foreach (unowned var child in list_box.get_children ()) {
             if (child is SidebarItemInterface) {
-                ((SidebarItemInterface)child).destroy_bookmark ();
+                ((SidebarItemInterface) child).destroy_bookmark ();
             }
+
+            list_box.remove (child);
```
danirabbit, line 32: I don't fully understand this, but if we call `remove` before `destroy` we get a whole bunch of errors about how this isn't a widget or isn't a container and there's unclassed pointers etc etc. But if we destroy and then remove there's nothing :woman_shrugging:

jeremypw: It is because you changed `Gtk.Widget` to  `unowned var `.  If you remove `child` from the listbox there is nothing referencing it so you cannot safely refer to it again.  Using `Gtk.Widget` adds another reference so the order of subsequent actions does not matter.

jeremypw (changes requested):
You can reduce the diff by reverting changes to the `clear ()` function.  Otherwise everything seems to work as before.

libcore/Interfaces/SidebarListInterface.vala
danirabbit: Ah I had no idea that the `var` keyword could have an effect on references. Thanks!

PR #2290, opened 2023-08-17
Release version 6.5.0

jeremypw:


danirabbit:
blocked until we figure out translations I guess :shrug:

jeremypw:
So is there any overwhelming reason not to merge #2224? I know @tintou was worried about useless translations but is it likely that many translatable strings will change temporarily during the Gtk4 port?  Presumably just changing the target branch or changing the location of a string will not involve re-translation?

danirabbit:
@jeremypw I don't think there's really a reason not to merge it at this point, like you said there's not really string changes happening right now

zeebok (approved):

Issue #1508, opened 2020-10-20
Provide a Flatpak manifest


danrabbit:
See https://github.com/elementary/calculator/blob/master/io.elementary.calculator.yml as an example

jeremypw:
I guess this will need a pretty porous sandbox :stuck_out_tongue: 

Nautilus uses
```
    "finish-args": [
        "--share=ipc", "--socket=x11",
        "--socket=wayland",
        "--talk-name=org.gnome.OnlineAccounts",
        "--talk-name=org.freedesktop.Tracker1",
        "--filesystem=host",
        "--talk-name=org.gtk.vfs", "--talk-name=org.gtk.vfs.*",
        "--talk-name=ca.desrt.dconf", "--env=DCONF_USER_CONFIG_DIR=.config/dconf"
    ],
```

But Dolphin has more restrictive access to the filesystem for some reason
```
    "finish-args": [ "--device=dri",
                     "--share=ipc",
                     "--share=network",
                     "--socket=x11",
                     "--socket=wayland",
                     "--filesystem=home",
                     "--filesystem=/run/media",
                     "--talk-name=org.kde.kiod5",
                     "--talk-name=org.kde.JobViewServer",
                     "--talk-name=org.kde.kpasswdserver",
                     "--system-talk-name=org.freedesktop.UDisks2" ],
```

Marukesu:
I tend to prefer the dolphin implementation over nautilus here

davidmhewitt:
> I tend to prefer the dolphin implementation over nautilus here

Yeah, well we don't really get a choice 😛

Dolphin's seems more restrictive because it uses the KIO bus. KIO is a daemon which essentially provides a file manager backend and Dolphin is just the user interface to that.

In theory, you can do that to some extent with gvfs, but I don't find it particularly weird that a file manager would ask for permissions to access the entire filesystem.

Marukesu:
So the actual status of a files flatpak:

- Can't delete folders, but deleting files works.
- Can't open external apps

To build the manifest I need to make some changes on meson and services files:

- added a `portal` and `with-plank` options so we can skip building the filechooser and plank.
- renamed `io.elementary.files.services` to `io.elementary.files.db.services` because flatpak-builder fails because of the mismatching ids.
- renamed `io.elementary.files.Filemanager1.service` to `org.freedesktop.Filemanager1.service` because `flatpak install` fails because of the wrong service name.

danrabbit:
[...]

PR #2291, opened 2023-08-19
Create flatpak.yml

danirabbit:
Just kind of an experiment to see what's missing/broken and if it could ever make sense to ship Files in a flatpak

Things to do:
- [ ] Not depend on LibPlank. Plank is going away anyways with the new dock. If the poofwindow is a pattern we want to keep we should move that into Granite probably or move it here for now
- [ ] `error: Name in service file /files/build/export/share/dbus-1/services/io.elementary.files.service does not match app id: io.elementary.files.db`
- [ ] No icon cache in the container? Why are we messing with the icon cache anyways :thinking: 
- [ ] `Icon not matching app id in /home/dani/Projects/files/build/export/share/applications/io.elementary.files.desktop: system-file-manager`
- [ ] `io.elementary.files.Filemanager1.service` is not an allowed file name?

Portal issues:
- [ ] `Not exporting share/dbus-1/services/org.freedesktop.impl.portal.desktop.elementary.files.service, non-allowed export filename`
- [ ] `../filechooser-portal/meson.build:29:4: ERROR: Assert failed: systemd required but not found, please provide a valid systemd user unit dir or disable it`

### DIFF
--- a/data/io.elementary.files.service.in
+++ b/data/io.elementary.files.service.in
@@ -1,3 +1,3 @@
 [D-BUS Service]
-Name=@exec_name@.db
+Name=@exec_name@
 Exec=@install_prefix@/@bin_dir@/@exec_name@-daemon
--- a/data/meson.build
+++ b/data/meson.build
@@ -13,7 +13,7 @@ configure_file(
 
 configure_file(
     input: meson.project_name() + '-filemanager1.service.in',
-    output: meson.project_name() + '.Filemanager1.service',
+    output: 'org.freedesktop.FileManager1.service',
     configuration: config_data,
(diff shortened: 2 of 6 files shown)


Marukesu:
I believe much of the status on this remains the same from [my last try on this](https://github.com/elementary/files/issues/1508#issuecomment-793265156). i don't think it gives us much making files a flatpak because the portal will need to be a host process anyway.

danirabbit:
Ah I was looking for this! I thought we tried this before and I wasn't sure what happened. Thanks for linking. Closing this then since there's already a good discussion :)

PR #2293, opened 2023-08-23
Control: add desktop-file-utils

danirabbit:
See: https://launchpadlibrarian.net/683369241/buildlog_ubuntu-jammy-amd64.pantheon-files_6.5.0+r6004+pkg121~ubuntu7.1_BUILDING.txt.gz

Looks like we need desktop-file-utils to use the meson GNOME module's `update-desktop-database` action

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -3,6 +3,7 @@ Section: x11
 Priority: optional
 Maintainer: elementary, Inc. <builds@elementary.io>
 Build-Depends: debhelper (>= 10.5.1),
+               desktop-file-utils,
                gettext,
                gnome-common,
                libcanberra-dev (>= 0.30),


jeremypw (approved):

jeremypw:
Did the other projects for which a similar change was made already have desktop-file-utils?

danirabbit:
None of the other projects are using the "update_desktop_database" or "gtk_update_icon_cache` actions, only "glib_compile_schemas"

jeremypw:
Are those actions still needed? I am not sure what makes Files different in that respect.

danirabbit:
@jeremypw I don't really think they should be needed tbh

jeremypw:
@bobby285271 Suggests we keep the desktop update (although its not needed when installed from package). I merged his PR removing the icon-cache update

PR #2299, opened 2023-08-28
Update build dependencies

danirabbit:
Removals:
* libgala
* libnotify
* libgail

LibHandy needs to be 1.2.0 for tabbar


jeremypw (approved):
Still builds with no different warnings so lets do it.

Issue #2305, opened 2023-08-31
Help option does not work if there is a busy Files process


jeremypw:
### What Happened?

Typed `io.elementary.files -h` and got output:
`GDBus.Error:org.freedesktop.DBus.Error.NoReply: Message recipient disconnected from message bus without replying
`
Prior to this I was using Files to display a location on a server that was inaccessible.

### Steps to Reproduce

1.  Connect a server and display a location on it
2. Close Files (with History/Restore Tabs ON)
3. Disconnect the server by e.g. turning off the network
4. Re-open Files - the tab enters a persistent working state
5. In a Terminal type `io.elementary.files -h`
6. Observe error message - Files may also crash

Also happens if the command is issued while a large folder is being loaded.

### Expected Behavior

Help output appears

### OS Version

8.x (Horus)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
See above
```

jeremypw:
Closing this as this option has started working after a reboot.

Marukesu:
i would reopen this and re-target to the fact that we currently do a DBus call for this when we have all the information needed to do in the new process itself.

jeremypw:
Maybe better to open a wishlist report proposing the desired change in coding, unless steps to consistently reproduce the issue can be found.

Marukesu:
i believe you can trigger this easily by blocking the global main context for long enough time. think about, loading a folder with a lot of files in it. and while it's loading, call `io.elementary.files -h`.

jeremypw:
OK, thanks but, forgive my ignorance, I am not sure how to fix this - are you able to push a PR some time?

Marukesu:
Sure, it's pretty straight-forward, i can send a PR later today.

janxkoci:
Will it also fix a timeout I observed [here](https://github.com/elementary/files/issues/1858#issuecomment-1701191315)? My error didn't mention DBus, but otherwise the symptoms looked the same (Jeremy was trying to reproduce the same issue at the time).

jeremypw:
@janxkoci The linked PR only addresses reacting to the CLI when already running so I do not think it will fix the underlying cause of the UI freezing in your situation.

janxkoci:
I understand that, I just meant if the timeout in my error is the same as your DBus error, or if there may be another similar bug to be fixed. I know the freezing will be more complex.

jeremypw:
Yes it should fix the timeout after the `--version` command.

PR #2307, opened 2023-09-01
Application: Rework Command Line Handling

Marukesu:
Let most of command line to be handled locally, only handling the window/tab creation in the primary instance.

Fixes: https://github.com/elementary/files/issues/2305

### DIFF
--- a/src/View/Window.vala
+++ b/src/View/Window.vala
@@ -24,7 +24,6 @@
 public class Files.View.Window : Hdy.ApplicationWindow {
     const GLib.ActionEntry [] WIN_ENTRIES = {
         {"new-window", action_new_window},
-        {"quit", action_quit},
         {"refresh", action_reload},
         {"undo", action_undo},
         {"redo", action_redo},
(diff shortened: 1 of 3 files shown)


src/View/Window.vala
```diff
@@ -757,8 +755,10 @@ public class Files.View.Window : Hdy.ApplicationWindow {
 
     private void add_window (GLib.File location = GLib.File.new_for_path (PF.UserUtils.get_real_user_home ()),
                              ViewMode mode = ViewMode.PREFERRED) {
-
-        marlin_app.create_window (location, real_mode (mode));
+        with (new Window (marlin_app)) {
```
jeremypw, line 758: I think we need to bump the minimum Vala version to use this syntax?

jeremypw:
Fixes the --help and --version issues completely, and creating a new or new tab works immediately where a tab is blocked e.g. due to disconnected server but if the tab is busy due to loading a large folder then nothing happens until the folder is loaded (and a "Files not responding" dialog may appear). However fixing that is not so easy and can be left for another PR.

src/View/Window.vala
Marukesu: yeah, it was introduced in vala 0.50, will it be fine to bump it? i can use the normal syntax if it preferred.

jeremypw: Yes, should be fine. Version 0.56 is supplied with Horus and we will soon be switching to Gtk4 anyway. We need not worry about compatibility with previous versions of elementary.

jeremypw: I shall have to remember that syntax - looks neat.

jeremypw (approved):
Thanks for doing this!

PR #2321, opened 2023-09-26
Incrementally go Vala-only

tintou:
Three source move operations


jeremypw (approved):
Code looks sane; copy, move and delete operations seem to work as usual.

PR #2333, opened 2023-10-08
Fix scroll to bottom after rename.

### LINKED ISSUE #2325: Rename triggers scroll to bottom?
### What Happened?

When Files is sorted on Modified and then rename a file Files seems to scroll to the bottom. 
File name does not seem to matter. 
Viewing mode does not seem to matter. 
The sorting matters, this only seems to happen when sorted on Modified. 

Rename triggers scroll to bottom
Rename triggers scroll to bottom 2

### Steps to Reproduce

1. Open a folder that has enough files so it's scrolls.
2. Sort on Modified.
3. Rename a file that's high in the list. 

### Expected Behavior

The scroll position doesn't change or follows the renamed file.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #2325 

 - Do not scroll to renamed file until model has resorted.

Because resorting of the model is now throttled and delayed it is necessary to delay selecting and scrolling to the renamed file until the model has resorted else the incorrect path may be scrolled to.  This PR simply idles while a sort is pending rather than introducing a signal and handler.

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -3423,7 +3423,14 @@ namespace Files {
             /* new_file will be null if rename failed */
             if (new_file != null) {
                 selected_files_invalid = true;
-                select_and_scroll_to_gof_file (new_file);
+                Idle.add (() => {
+                    if (model.sort_pending) {
+                        return Source.CONTINUE;
+                    } else {
+                        select_and_scroll_to_gof_file (new_file);
+                        return Source.REMOVE;
+                    }
+                });
             }
         }
 


peteruithoven (approved):
I'm not familiar with the Files code base to properly judge the impact of this. But I've tested the change and it does solve the problem.

jeremypw:
@peteruithoven Thanks for the reviews!  It can be difficult to predict the full effect of changes to Files code just by looking at the code unfortunately. The best most can be do is to test manually for regressions in related and common operations.  Regressions still get through - we just have to hope they are noticed before release :crossed_fingers:

PR #2339, opened 2023-10-10
Save selection for all slots that will reload

### LINKED ISSUE #2328: Selection changed on refresh?
### What Happened?

When you have a file selected and then refresh Files (ctrl+r) the selection seems to change to a random other file.

Refreshing with selection changes selection

Here I select a file, press the down arrow a couple of times, refresh and press the down arrow again a couple of times. 
Refreshing with selection changes selection 3

### Steps to Reproduce

1. Select a file
2. Refresh files (ctrl+r)
3. Press up or down keys

### Expected Behavior

When pressing up you expect the files above your original file to be selected, when pressing down the one below. 

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #2328 

 - Adds a `prepare_reload ()` function to ViewContainer that saves the current selection
 - Every slot that will reload calls `prepare_reload ()` before clearing model.
 - All refresh operations initiated with `win.refresh` action

Now slots opened at the same location in different tabs/windows with different selections will all refresh and keep their selection.

### DIFF
--- a/src/View/DirectoryNotFound.vala
+++ b/src/View/DirectoryNotFound.vala
@@ -54,7 +54,7 @@ namespace Files.View {
                 }
 
                 if (success) {
-                    ctab.reload ();
+                    get_action_group ("win").activate_action ("refresh", null);
                 }
             });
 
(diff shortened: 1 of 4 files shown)


peteruithoven (approved):
I am not really familiar enough with this code base to judge the full implications, but I've tested it and the selection seems to now remain during refreshes.

PR #2340, opened 2023-10-10
Refresh thumbnail after file change

### LINKED ISSUE #2327: The thumbnail does not update?
### What Happened?

The thumbnail of image files does not seem to update when they are changed. For example when cropping the image. Not even after refreshing Files (ctrl+r).
When I copy and paste the updated image the thumbnail of the new file is correct.
Especially when cropping large files like a full screen screenshot, the thumbnails are currently not very helpful. 

Thumbnail not updated after crop

### Steps to Reproduce

1. Open a folder with an image
2. Open the image in for example Photo viewer
3. Crop the image
4. Save changes to the file
5. Optional: Refresh Files (ctrl+R)

### Expected Behavior

Changes to file updates thumbnail.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

### LINKED ISSUE #2455: No large thumbnails
### Discussed in https://github.com/orgs/elementary/discussions/556

<div type='discussions-op-text'>

<sup>Originally posted by **doobes** August 17, 2024</sup>
Okay, this is decidedly sub-optimal:

image

Why is this displayed this way?

These images are 1500 x 1000 pixels.</div>

jeremypw:
Fixes #2327 
Fixes #2455  (not intentional)

* Handle CHANGES_DONE_HINT events by updating icon
* Ignore duplicate events and events from transient files
* Avoid issues causes by caching of thumbnails internally by only caching remote thumbnails
* Improve performance by avoiding unnecessary icon lookups when scrolling
* Move some cache and thumbnail related code from AbstractDirectoryView to File
* Fix an issue (on OS8) where some standard icons were not displayed on startup.

The issue was fundamentally caused by CHANGES_DONE_HINT events from the directory monitor being ignored.  However it was found that just fixing that did not reliably fix the problem - icons did not always update at all sizes. This was found to be due to the way Files caches icons internally.  Trying to work round this proved intractable. It was found that dropping caching for local files and themeable icons combined with reducing the number of lookups being performed unnecessarily resulted in no discernible performance hit on modern hardware (i5 processor and SSD storage).   

Caching was retained for remote files to maintain performance on slow connections. Remote locations cannot be monitored for changes by the directory monitor so the linked issue will always occur for remote files anyway.


jeremypw:
Investigating regression.

libcore/File.vala
```diff
@@ -624,18 +626,126 @@ public class Files.File : GLib.Object {
         }
     }
 
-    public void update_icon (int size, int scale) {
-        if (size <= 1) {
-            return;
+    /* Icon/thumnail related functions */
+
+    public GLib.Icon? get_icon_user_special_dirs (string path) {
```
danirabbit, line 631: Isn't this covered by `FileAttribute.STANDARD_ICON` in FileInfo?

jeremypw: Thanks for taking a look at this. You are probably right - I'll look into it.  The scope of this PR seems to have grown substantially, into an extensive review of icon handling as fixing the issue proved more complicated than envisaged.  A lot of the code is old and hard to understand. Simplifying/clarifying will aid future porting to Gtk4.

jeremypw:
@danirabbit Looks like FileAttribute.STANDARD_ICON only returns a standard folder icon for user folders so cannot substitute for that function I am afraid.  However, while checking this, I noticed there are some new? attributes related to thumbnail status that might be useful. I guess they might not work for non-native/remote files though unless gvfs generates them locally.

libcore/File.vala
jeremypw: If its OK, I would rather leave changing this (if required) to a later PR.  The scope of this one is already larger than intended.

danirabbit:
@jeremypw is there any way to break this up into smaller chunks? I'm having trouble following all of the changes here. I'm not really familiar with this code but I know you've been waiting for review here for quite some time.

jeremypw:
@danirabbit I'll try!

libcore/IconInfo.vala
```diff
@@ -121,8 +121,8 @@ public class Files.IconInfo : GLib.Object {
         }
     }
 
-    public static Files.IconInfo? get_generic_icon (int size, int scale) {
-        var generic_icon = new GLib.ThemedIcon ("text-x-generic");
+    public static Files.IconInfo get_generic_icon (int size, int scale) {
+        var generic_icon = new GLib.ThemedIcon ("text-x-generic"); // Assume theme always contains this?
```
danirabbit, line 125: it is in the fd.o spec so we should be able to assume it exists https://specifications.freedesktop.org/icon-naming-spec/latest/#names

jeremypw: Thanks, I'll remove that comment then.

jeremypw:
I think the linked issues have now been resolved by the partial PRs that have been merged so its not worth trying to sort out the conflicts.  I am going to close this.

Issue #2342, opened 2023-10-13
Files no longer opens


gvansanden:
### What Happened?

Nothing when I click the icon, yet it will run as root
Debug output below

### Steps to Reproduce

1. click files, nothing happens
2. open files in the CLI, throws error

### Expected Behavior

Files should open

### OS Version

7.x (Horus)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
❯ /usr/bin/io.elementary.files -d  
** (io.elementary.files:13838): DEBUG: 14:58:23.143: Client loading provider: com.nextcloudgmbh.Nextcloud /com/nextcloudgmbh/Nextcloud

** (io.elementary.files:13838): DEBUG: 14:58:23.144: Window.vala:578: No info for requested location - abandon loading

(io.elementary.files:13838): Handy-CRITICAL **: 14:58:23.144: hdy_tab_page_get_child: assertion 'HDY_IS_TAB_PAGE (self)' failed

Reset dconf settings for files and that does not help

(gdb) run
Starting program: /usr/bin/io.elementary.files 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff2dff640 (LWP 15500)]
[New Thread 0x7ffff25fe640 (LWP 15501)]
[New Thread 0x7ffff1dfd640 (LWP 15502)]
[New Thread 0x7ffff15bb640 (LWP 15503)]
[New Thread 0x7fffe3fff640 (LWP 15506)]
[Thread 0x7fffe3fff640 (LWP 15506) exited]
[New Thread 0x7fffe3fff640 (LWP 15507)]
[New Thread 0x7fffe33ff640 (LWP 15508)]
[Thread 0x7fffe3fff640 (LWP 15507) exited]
[New Thread 0x7fffe3fff640 (LWP 15509)]
[New Thread 0x7fffe2bfe640 (LWP 15510)]
[Thread 0x7fffe33ff640 (LWP 15508) exited]
[Thread 0x7fffe3fff640 (LWP 15509) exited]
[Thread 0x7fffe2bfe640 (LWP 15510) exited]
[New Thread 0x7fffe2bfe640 (LWP 15511)]
[New Thread 0x7fffe3fff640 (LWP 15512)]
[Thread 0x7fffe2bfe640 (LWP 15511) exited]
[New Thread 0x7fffe2bfe640 (LWP 15513)]
[...]

PR #2343, opened 2023-10-13
Expose restore tabs on startup option

jeremypw:
At the moment the only way to turn off restoring tabs in Files without turning of History for *all* apps in the System Settings is to edit the gsettings using either `dconf-editor` or the CLI, which is non-obvious.

This PR adds another switch to the AppMenu to toggle this setting.  This is also useful for allowing non-technical users to workaround/diagnose some issues (see https://github.com/elementary/files/issues/2342#issuecomment-1761801959)


src/View/Widgets/AppMenu.vala
```diff
@@ -83,10 +83,14 @@ public class Files.AppMenu : Gtk.Popover {
             action_name = "win.singleclick-select"
         };
 
-        var foldes_before_files = new Granite.SwitchModelButton (_("Sort Folders before Files")) {
+        var folders_before_files = new Granite.SwitchModelButton (_("Sort Folders before Files")) {
             action_name = "win.folders-before-files"
         };
 
+        var restore_tabs = new Granite.SwitchModelButton (_("Restore Tabs on Starting")) {
```
zeebok, line 90: Not sure if this wording is best? When I see "start-up" I think more the machine/OS starting up. I am almost tempted to keep it layman and suggest "Restore Tabs from Last Time" but I am not totally convinced that is a good alternative lol

jeremypw: How about "Restore Tabs when Launching"

danirabbit: I feel like "when Launching" might be too ambiguous. 

Web calls this "Restore Tabs on Startup" which would we could go with for consistency
Chrome calls it "Continue where you left off" which feels super ambiguous :grimacing:
Firefox calls this "Show Tabs from Last Time" which I think validates @zeebok here and might be the most clear

jeremypw: OK, I'll go with @zeebok 's suggestion then.

zeebok (approved):

zeebok:
Seems like checks failing due to this compile error:

`../src/Application.vala:188.17-188.36: error: `GLib.ApplicationCommandLine.printerr_literal' is not available in gio-2.0 2.72.4. Use gio-2.0 >= 2.80`

jeremypw:
@zeebok  Yes, this change originated a while back in #2307  but at the time checks passed - I am not sure what has changed since then to cause this failure.  @Marukesu Any ideas?

Marukesu:
It seems to be a mismatch between the vapi and library version.

glib [added a wrapper function to the printerr_literal vfunc in 2.80](https://gitlab.gnome.org/GNOME/glib/-/commit/b9c90c3a3e17f067870703c51d818020436acb1b). but the vfunc is older.

jeremypw:
@Marukesu Thanks for the info.  Can it be changed back to `printerr` for now so it passes CI or can it be fixed by changing the workflow (in which case how)?  The tests pass when run on OS8.

Marukesu:
Yes, there is no harm in changing it to cmd.printerr().

PR #2347, opened 2023-10-22
[Feature] Allow hiding any file

### LINKED ISSUE #2345: Hiding Files and Folders Without Adding dot in the File Name
### What Happened?

In Nautilus and probably other file browsers too (I tried on KDE Index) it is possible to hide files and folders without adding a dot in front of their names. We can list the names of the folders and files inside a text file called `.hidden`, which will hide them. 

I need to do this because one of my device drivers puts its configuration files in the home directory and renaming them could potentially remove my settings. I want to declutter by hiding them but without renaming them. One way to do that is by listing them on a `.hidden` file as I mentioned before.

### Steps to Reproduce

1. Want to hide file without renaming
2. Can't do anything, I have to rename

### Expected Behavior

Hide every file and folder listed in `.hidden`

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #2345 

 - Add a selection menuitem to set/unset a new "metadata::hidden" file attribute
 - Prevent altering hidden status of certain filename formats
 - Make consequential changes to handling of hidden files in ListModel, File and Directory to allow hidden state to change after File construction.

Could not use `standard::is-hidden` because this attribute is not writable so the tag could not persist.

### DIFF
--- a/libcore/Directory.vala
+++ b/libcore/Directory.vala
@@ -809,8 +809,18 @@ public class Files.Directory : Object {
         }
     }
 
+    // Called by Model when loading a subdirectory
     public List<unowned Files.File> get_files () {
-        return file_hash.get_values ();
+        List<unowned Files.File> file_list = null;
+        foreach (unowned Files.File gof in file_hash.get_values ()) {
+            if (gof != null && gof.info != null &&
+                (!gof.is_hidden || Preferences.get_default ().show_hidden_files)) {
+
+                file_list.prepend (gof);
+            }
+        }
+
+        return file_list;
     }
 
     public void load_hiddens () {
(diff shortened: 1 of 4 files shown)


danirabbit:
I'm not sure about creating new UI for what sounds like is a workaround for an edge case/bug/inappropriate behavior in a 3rd party component.

The solution of a .hidden file seems more appropriate here imo, especially because this isn't something that it sounds like one needs quick access too or does frequently. If that's something other file browsers already support then it's probably documented, maybe in tutorials regarding installing OP's drivers, and is probably what some folks are already expecting to be able to do or have created scripts for etc

jeremypw:
@danirabbit Comparing each filename against the contents of a manually created `.hidden` text file seems even less optimal - it is very inefficient and would be affected by renaming the file. I agree that another menu option is not desirable either for a rarely used function but there are other rarely used items as well.  Maybe consider having an "Other Actions" submenu for these like in Photos?

I am not that invested in implementing this feature tbh - this was just a proof of principle.

Reverting to draft for now.

jeremypw:
Closing as it is  :-1: for UI.

PR #2348, opened 2023-10-23
Disable drop onto parent folder

### LINKED ISSUE #1759: Dragging and dropping file in same folder should not create link
## What Happened

When dragging and dropping a file within the same folder, a symlink should not be created. Most of the time, I dragged the file by mistake and found no convenient way of canceling the action short of dragging the file out of the window, which is not an elegant solution. Less advanced users may also mistake the file for a duplicate and accidentally delete the original file.

## Expected Behavior

If a file is dragged and dropped into the same folder, the action should be ignored.

## Steps to Reproduce

1. Drag a file
2. Drop it into the same folder

## Platform Information

Screenshot from 2021-07-14 10-55-34

jeremypw:
Fixes #1759 

Simplest possible fix.  As DND will have to be completely rewritten for Gtk4 there is no point in doing more.


libcore/FileUtils.vala
```diff
@@ -828,7 +828,7 @@ namespace Files.FileUtils {
             var parent = drop_file.get_parent ();
 
             if (parent != null && parent.equal (target_location)) {
-                valid_actions &= Gdk.DragAction.LINK; // Only LINK is valid
+                valid_actions = Gdk.DragAction.DEFAULT; // Ignore (GitHub bug #1759)
```
danirabbit, line 831: According to docs:
> [DEFAULT](https://valadoc.org/gdk-3.0/Gdk.DragAction.DEFAULT.html) - Means nothing, and should not be used.

Also I'm getting the replace dialog here now for files which seems even less desirable than the symlink behavior

danirabbit (changes requested):

libcore/FileUtils.vala
jeremypw: Not sure what to use for "Do Nothing" then, I guess use `0`?  I'll  just drop this and revisit in Gtk4.

jeremypw:
Closing - will revisit in Gtk4.

PR #2353, opened 2023-10-31
Add focus sidebar action to window

### LINKED ISSUE #51: Cannot rename bookmark entry with F2 because focus is given to fileview after click
When I click a bookmark item in sidebar (a custom bookmark that I have created myself, not the default ones), I can't seem to be able to trigger rename with F2 right after I have clicked on it. 

I have to hover my mouse over the file view and then back on the sidebar and press f2 to do this. @jeremypw  was saying this is because the focus is given to the file view after clicking on a bookmark entry to make Files friendly to keyboard users. 

Should the focus be hijacked and given to the file view even though the user is oblivious to it? Why not better keep the focus where it is and give the user a key to press to shift focus to the file view? 

I am unable to check now how mac OS or Windows handle this. 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/46557003-cannot-rename-bookmark-entry-with-f2-because-focus-is-given-to-fileview-after-click?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F65602118&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Possible solution to allow users access to the sidebar without resorting to the mouse by pressing `<Ctrl>Left`.  Note that the sidebar can already be unfocused with the shortcut `<Ctrl>Right`.

Partially addresses #51.

A possible alternative, or additional, solution would be to have a "Rename currently focused bookmark" action with an accelerator such as `<Ctrl>F2`.


zeebok:
While the code seems good, when I test on an OS8 vm it does not seem to move focus to the sidebar.

jeremypw:
@zeebok Thanks for look at this!  I'll look at it again now I am working on OS8 Wayland.

jeremypw:
@zeebok Just tested under Wayland and X on OS8 and it works for me.  It did stop working after a system update but reinstalling the branch fixed that.  Maybe something to do with the vm or maybe that key combination is used for something else on your system?

jeremypw:
It is an open question whether we should also facilitate focusing the Storage listbox.  After this PR is merged (which only focuses the Bookmark list) the Storage list can be focused by pressing `Tab` (`<Shift>Tab` to go the other way) but this is not so obvious.

danirabbit:
I don't think this solves #51. I feel like OP is right that the current focus behavior is just weird. Gonna comment on the original issue

src/View/Sidebar/SidebarWindow.vala
```diff
@@ -230,6 +230,10 @@ public class Sidebar.SidebarWindow : Gtk.Box, Files.SidebarInterface {
         device_listbox.refresh_info ();
     }
 
+    public void focus () {
```
danirabbit, line 233: `hides inherited method `Gtk.Widget.focus'. Use the `new' keyword if hiding was intentional`

danirabbit (changes requested):
I don't totally love this, but it works. I tried something else with `move_focus` but it seems like it always puts focus inside the device list. Just one requested change because of terminal warning

src/View/Sidebar/SidebarWindow.vala
jeremypw: Fixed.  Renamed a couple of functions to match more closely their actual effect.

jeremypw:
@danirabbit I agree it is hard to find a good solution for the case where a user is using both mouse/touch *and* keyboard.   Maybe the focus should not switch to the view when a bookmark is activated with the keyboard?  Either way, I imagine a keyboard (only) user would appreciate a convenient way of focusing the bookmarks that works whatever else is focused.   Not sure what to do about focusing the storage or network entries though.

jeremypw:
@zeebok This is a possibility, provided that as soon as a directory view item is selected then the rename action acts on that.  It would still be no possible to rename the bookmark using the keyboard once the view has been used.  The only solution for that (other than this one) would be to have a different shortcut for renaming bookmarks.

zeebok:
> It would still be no possible to rename the bookmark using the keyboard once the view has been used.
I think this would probably be fine, given the context of actions that have happened. Might be worth asking the issuer what they think since it is a scenario they are in more than we are?

> It is a bit weird to have the same shortcut for two different actions anyway.
Totally agree, this is a very weird problem to solve hahaha

jeremypw:
Is there anything about this that blocks merging (even if it is not a full or ideal solution for the renaming issue)?  As we already have `<Ctrl>Right` to move focus from sidebar to view a matching shortcut to move focus the other way seems logical.

zeebok:
I don't think so though I don't see any kind of <Ctrl>Right shortcut and I still can't actually use the shortcut on my local install of OS8 on my ThinkPad. So I am at a bit of a loss :sweat_smile:

zeebok (approved):
The `<Ctrl>Right` is a little odd, I have to hit `Down` after to navigate the pane and highlight and it always skips down to the second row.

danirabbit (approved):

PR #2358, opened 2023-12-09
Fix unexpected selection after file added

### LINKED ISSUE #2357: Unexpected selection after creating a new file
### What Happened?

Probably is missing a visual file selecting in the code when is created Empty file. I did not check that.

### Steps to Reproduce

A newly created file (no selected file in folder -> the right click outside of files -> select Empty file -> enter) is not selected visually but is opened when is pressed enter.

### Expected Behavior

Add selecting a newly created file visually or cancel opening the new file with the enter key when is not selected visually.

### OS Version

7.x (Horus)

### Software Version

### Log Output

```shell
$ io.elementary.files --version
$ io.elementary.files 6.5.2
```
```


### Hardware Info

_No response_

jeremypw:
Fixes #2357

For performance reasons, resorting the model after adding files was throttled in a previous commit.  So files should not be added to the selection until the model has resorted else the wrong file may be selected.

It was noted that adding and, in particular, removing files does not scale well but that will need to be addressed later, maybe after porting to Gtk4.

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -945,9 +945,13 @@ namespace Files {
         // file operations
         private void add_file (Files.File file, Directory dir, bool is_internal = true) {
             model.insert_sorted (file, dir);
-
             if (is_internal) { /* This true once view finished loading */
-                add_gof_file_to_selection (file);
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
Makes sense, nice job

PR #2360, opened 2023-12-11
Release 6.5.3

jeremypw:


danirabbit (approved):
Thanks for preparing this release and addressing all of these issues!

In the future, please update release descriptions ahead of time, separately, so we can give translators at least a couple days with them. These release notes unfortunately won't be translated in this release since the strings weren't present ahead of time

PR #2364, opened 2024-01-06
[Gtk4 Prep] IconView: Remove deprecated, cleanup

jeremypw:
For Gtk4 prep:
 - Do not use `Gdk.ButtonEvent`
 - Make rubberbandable areas slightly larger to compensate

Code cleanup:
 - Merge some functions
 - Modernise code style
 - Lose duplicated range check

This is assuming that the initial Gtk4 port will not replace the existing views with dynamic equivalents since that results in unreviewable PRs.  Moving to dynamic views can be done after the initial port with smaller PRs.

Note: `freeze_child_notify` and `thaw_child_notify` will have to be removed for Gtk4 and there is no replacement (?).  However removing those will be left for another PR as it affects several files.

### DIFF
--- a/src/View/IconView.vala
+++ b/src/View/IconView.vala
@@ -62,6 +62,7 @@ namespace Files {
             tree.add_attribute (name_renderer, "text", ListModel.ColumnID.FILENAME);
             tree.add_attribute (name_renderer, "file", ListModel.ColumnID.FILE_COLUMN);
             tree.add_attribute (name_renderer, "background", ListModel.ColumnID.COLOR);
+            tree.bind_property ("item-width", name_renderer, "item-width", DEFAULT);
 
             connect_tree_signals ();
             tree.realize.connect ((w) => {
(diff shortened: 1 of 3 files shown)


src/View/IconView.vala
```diff
@@ -34,69 +34,53 @@ namespace Files {
...
         protected override void set_up_zoom_level () {
+            minimum_zoom = (ZoomLevel)Files.icon_view_settings.get_enum ("minimum-zoom-level");
+            maximum_zoom = (ZoomLevel)Files.icon_view_settings.get_enum ("maximum-zoom-level");
+            zoom_level = (ZoomLevel)Files.icon_view_settings.get_enum ("zoom-level");
+
             Files.icon_view_settings.bind (
                 "zoom-level",
                 this, "zoom-level",
-                GLib.SettingsBindFlags.DEFAULT
+                GLib.SettingsBindFlags.SET
```
danirabbit, line 82: Should this be `SET | SYNC_CREATE`?

jeremypw: That is for property binding  SettingsBindFlags.SYNC_CREATE does not exist.  The previous line initialises `zoom-level` the setting.

danirabbit: Oh sorry I misread my bad!

danirabbit (approved):
Looks good! :rocket:

PR #2365, opened 2024-01-08
[Gtk4Prep] Treeviews cleanup and Gtk4 prep

jeremypw:
Gtk4 related:
 - Avoid use of Gdk.EventButton in some cases

Cleanup:
 - Merge single use functions
 - Avoid repeating some code
 - Modernise code style
 -  Inline namespace

All changes should be functionally equivalent to the original

Hide whitespace when reviewing to ignore inlining of namespace


jeremypw:
I think I'll split this into smaller PRs to reduce the diff.

PR #2366, opened 2024-01-08
[Gtk4Prep] Slots cleanup Gtk4prep

jeremypw:
Gtk4 Prep:
 - No issues found

Cleanup:
 - Merge single use code
 - Modernize construction
 - Fix whitespace and layout

### DIFF
--- a/src/View/Miller.vala
+++ b/src/View/Miller.vala
@@ -47,7 +47,7 @@ namespace Files.View {
             }
         }
 
-        public Miller (GLib.File loc, View.ViewContainer _ctab, ViewMode _mode) {
+        public Miller (GLib.File loc, View.ViewContainer _ctab) {
             Object (
                 location: loc,
                 root_location: loc,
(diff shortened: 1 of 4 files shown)


src/View/Miller.vala
```diff
@@ -47,10 +47,15 @@ namespace Files.View {
             }
         }
 
-        public Miller (GLib.File loc, View.ViewContainer ctab, ViewMode mode) {
-            this.ctab = ctab;
-            this.root_location = loc;
+        public Miller (GLib.File loc, View.ViewContainer _ctab, ViewMode _mode) {
```
danirabbit, line 50: It looks like this takes an argument `_mode` that is never used?

danirabbit (commented):
One comment/question otherwise looks good

src/View/Miller.vala
jeremypw: True - its always been like that I think.  I guess the idea was to allow the possibility of a multi-slot view with the individual slots showing grid or list views instead or column views.  In the event, that was never implemented and the slots are hard-coded as column views so I guess that parameter is  redundant.  I'll remove it.

danirabbit (approved):
:rocket:

PR #2373, opened 2024-01-12
[Gtk4Prep] ButtonWithMenu: Replace GtkMenu with GMenu and GtkPopover

jeremypw:


danirabbit:
I think this is better left to the final GTK4 port branch since it breaks menu styling. The diff in the make compile branch for this class is already only +17 -11 https://github.com/elementary/files/pull/2225/files#diff-083323fd42898651d7338f563d8d5f6bd7d7f28f8b5d729216efadd630bf5fdc

PR #2374, opened 2024-01-12
[Gtk4Prep] ViewSwitcher: Do not use Gtk.RadioButton

jeremypw:


danirabbit (commented):
This doesn't really seem significantly better when we already ported this class in the make compile branch and the diff there is already only +17 -17 https://github.com/elementary/files/pull/2225/files#diff-9b91e87990cec2bca936b271ed70f12fbb9fb085b08cf0c9c948f811c9ef8fd5

Issue #2376, opened 2024-01-16
Tab name does not update on location change after tab dragged to new window


jeremypw:
### What Happened?

After dragging a tab into another window, navigating in the new window does not change the name of the tab (although the view does navigate)

### Steps to Reproduce

1.  Open at least two tabs in a Files window
2. Drag one of the tabs into a new window
3. Change the location of the tab - note that the tab name does not update

### Expected Behavior

The tab behaves normally in the new window

### OS Version

7.x (Horus)

### Software Version

Latest release (I have run all updates)

jeremypw:
This issue does *not* occur if the new window is created in other ways - i.e. though the new-window action or through a menu option.

PR #2377, opened 2024-01-18
[Regression] Reload directory after mode change

jeremypw:
Following the fix of #2376 a further regression has emerged where the view does not reload after changing mode (it loads only on changing location).  This is because on a mode change the associated window does not change.  

This PR fixes the regression by explicitly loading the directory after a mode change.


danirabbit (approved):
Oops, not sure how we missed that :sweat_smile: Can confirm the fix

PR #2379, opened 2024-01-18
[GtkPrep] Cleanup treeviews contd

jeremypw:
Extracted from #2365 

Removes some unnecessary complication.  Should not have any functional effect.


danirabbit (approved):
Looks good, nice cleanup!

PR #2380, opened 2024-01-20
[Gtk4Prep] Views: Use EventControllerKey where possible

jeremypw:
An EventControllerKey is added to AbstractDirectoryView.

It was still necessary to handle `Gdk.EventKey` in some places, in particular to override the native behaviour of Gtk.TreeView.  This will become unnecessary on porting to Gtk4.

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -235,14 +235,14 @@ namespace Files {
                          */
                         disconnect_tree_signals ();
                         clipboard.changed.disconnect (on_clipboard_changed);
-                        view.key_press_event.disconnect (on_view_key_press_event);
                     } else {
-                        view.key_press_event.connect (on_view_key_press_event);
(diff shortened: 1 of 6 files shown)


src/View/AbstractDirectoryView.vala
```diff
@@ -235,14 +235,14 @@ namespace Files {
...
-                        view.key_press_event.disconnect (on_view_key_press_event);
                     } else {
-                        view.key_press_event.connect (on_view_key_press_event);
                         clipboard.changed.connect (on_clipboard_changed);
                         connect_tree_signals ();
 
                         update_menu_actions ();
                     }
+
+                    key_controller.propagation_phase = value ? Gtk.PropagationPhase.NONE : Gtk.PropagationPhase.BUBBLE;
```
danirabbit, line 245: that's clever

danirabbit (approved):
Works as expected afaict. Nice job

PR #2381, opened 2024-01-21
[Gtk4Prep] Window: Use EventControllerKey

jeremypw:
'key-press-event` does not exist in Gtk4.  This replace it with an EventController

### DIFF
--- a/src/View/Window.vala
+++ b/src/View/Window.vala
@@ -96,6 +96,8 @@ public class Files.View.Window : Hdy.ApplicationWindow {
     private int restoring_tabs = 0;
     private bool doing_undo_redo = false;
 
+    private Gtk.EventControllerKey key_controller; //[Gtk3] Does not work unless we keep this ref
+
     public signal void loading_uri (string location);
     public signal void folder_deleted (GLib.File location);
     public signal void free_space_change ();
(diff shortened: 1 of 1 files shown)


danirabbit:
How can I test this? I'm having trouble figuring out from the code what this is supposed to do

jeremypw:
@danirabbit This causes a search to commence (in the search bar) if you just start to type printable characters (i.e. you do not have to press `<Ctrl>F` first.  This should continue to work as normal. It relies on the fact that the native TreeView searching has been suppressed so the key event actually reach the window.

danirabbit:
@jeremypw this doesn't seem to work for me in any view and it also seems to not work in main

jeremypw:
Strange - I thought the merging master might have caused a regression but it still works fine for me in all views. When you type letters in ListView does a native search box appear?  When you press other keys (e.g. Menu and cursor keys do they work as normal?

jeremypw:
@danirabbit Try running the debug version in a terminal - this will show where the key events are going.

jeremypw:
The last commit may fix it if your system produces extra state flags for some reason.

danirabbit:
Works now! This is the Terminal output I get with this branch:
```
** (io.elementary.files:120899): WARNING **: 11:08:44.076: AbstractDirectoryView.vala:2997: ADV on_view_key_press_event

** (io.elementary.files:120899): WARNING **: 11:08:44.076: AbstractDirectoryView.vala:3257: ADV on_view_key_press_event returning false

** (io.elementary.files:120899): WARNING **: 11:08:44.082: Window.vala:339: Window key pressed

** (io.elementary.files:120899): WARNING **: 11:08:44.082: Window.vala:346: find
```

danirabbit (approved):
Approving, just needs debug stuff removed :)

PR #2382, opened 2024-01-21
[Gtk4Prep] Bookmarkrow: Use EventControllerKey

jeremypw:
Also  handles menu key and avoids use of events absent in Gtk4  in `popup_context_menu`


danirabbit (approved):
Works as expected!

PR #2383, opened 2024-01-21
[Gtk4Prep] EditableLabel: Use EventControllerKey, cleanup

jeremypw:
Use EventControllerKey 

Cleanup:
- TextRenderer: Set some EditableLabel properties that do not change after construction in the constructor
 - TextRenderer: Lose alignments that have no effect
 - AbstractEditableLabel: Lose unused or unneeded properties
 - Simplify and DRY construction, lose /merge unneeded functions
 - MultilineEditableLabel - draw_outline is always true


danirabbit (approved):
Changes make sense, nice cleanup, everything works!

PR #2387, opened 2024-01-23
[Gtk4Prep] TreeView & Miller: Avoid remaining Gdk.EventKey

jeremypw:
Avoids the use of special handlers that block the native key event handlers of Gtk.TreeView by putting Window EventControllerKey in CAPTURE phase.

Moves code to call the Miller key event handler into the AbstractDirectoryView handler.


danirabbit (approved):
Great cleanup!

PR #2389, opened 2024-01-23
[Gtk4Prep] BreadcrumbEntry: Use EventControllerKey 

jeremypw:
Behaviours that should be unaffected when the entry is focused are

 - Tab completion (`Tab`)
 - Change entry to parent path (`<Control>Up`)
 - Exit edit mode and focus view (`<Control>Down`)
 - Do not exit if `Enter` pressed in search mode
 

Cleanups
 - Merge single-use or one-line functions
 - Do not use event.is_modifier
 - Do not try to handle shortcut already used as action accelerator by Window


danirabbit (approved):

danirabbit:
Appreciate the notes for things to test!

PR #2391, opened 2024-01-24
Do not accept drop if location is unmounted

### LINKED ISSUE #2390: Drag N Dropping file on a disk that is not mounted makes the file disappear
### What Happened?

### Steps to Reproduce

1.get a usb or whatever

### Expected Behavior

### OS Version

7.x (Horus)

### Software Version

### Log Output

```shell
```

### Hardware Info

### LINKED ISSUE #2465: Dropping onto mountable bookmark does not work correctly if location mounted after Files starts
### What Happened?

See below.

### Steps to Reproduce

1. Start Files
3. Mount it

### Expected Behavior

### OS Version

8.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #2390 
Fixes #2465

 * `file_accepts_drop ()` returns false if location unmounted (fixes first issue)
 * Update the target location if mounted after Files started (fixes second issue)
 *


danirabbit (approved):

danirabbit (changes requested):
Can you resolve conflicts? It looks like maybe this function moved in another commit

jeremypw:
Conflict fixed

danirabbit:
@jeremypw ah this is a +/-0 diff now :sweat_smile: That doesn't seem right

jeremypw:
Looks like `libcore/FileUtils` was moved elsewhere in `main` so need to rewrite this branch if the issue persists.

jeremypw:
Noticed a related issue - the volume bookmarks only have a valid target file if they are already mounted when Files window is opened. In main, the target file is not updated if the volume is subsequently mounted so dropping a file on it still does not work in this branch.  This is now fixed by updating the target file whenever the volume is mounted or unmounted.

jeremypw:
@danirabbit This is now ready for review hopefully.

jeremypw:
Converting to draft to investigate bug where a (copy) file is created in the current working folder when Files is being run in a terminal and the file is dropped onto an unmounted volume.

jeremypw:
For now, I have not implemented auto-mounting of unmounted volumes when dropped on. This could be implemented in a further PR if thought desirable.

jeremypw:
> Converting to draft to investigate bug where a (copy) file is created in the current working folder when Files is being run in a terminal and the file is dropped onto an unmounted volume.

I am no longer sure how to reproduce this bug so a shorter PR has been proposed omitting the related commit.

jeremypw:
Closing in favour of #2466 and #2468

PR #2393, opened 2024-01-28
[RegressionFix] Window: Propagate key events when view not having focus

### LINKED ISSUE #2392: Search resets after pressing /
### What Happened?

When I try to type a path into a search bar it gets reset every time I press `/`

### Steps to Reproduce

1. Open Files
2. Try to navigate using the search bar

### Expected Behavior

Being able to type a path into a search bar

### OS Version

Early Access

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #2392

### DIFF
--- a/src/View/ViewContainer.vala
+++ b/src/View/ViewContainer.vala
@@ -461,6 +461,7 @@ namespace Files.View {
         public void set_active_state (bool is_active, bool animate = true) {
             var aslot = get_current_slot ();
             if (aslot != null) {
+                aslot.grab_focus ();
                 /* Since async loading it may not have been determined whether slot is loadable */
                 aslot.set_active_state (is_active, animate);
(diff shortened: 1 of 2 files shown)


lenemter (commented):
In this branch I have to press Ctrl + F to focus search bar and use it. In `main` it focuses automatically.

jeremypw:
@lenemter Thanks for spotting that!  I think I have fixed it properly now.

jeremypw:
With these changes needed to ensure that the active slot has keyboard focus else switching tabs between an unloadable view (cannot focus) and a loaded view (can focus) did not cause the view to become focussed so key events were not processed.

lenemter (approved):
Fixes the issue for me.

PR #2398, opened 2024-01-29
Window: do not capture keyevents when view is renaming

### LINKED ISSUE #2397: Rename doesn't allow to type any characters
### What Happened?

When renaming a folder I am unable to type anything in the field.

### Steps to Reproduce

1. Select a file
2. Choose rename in the right click menu

### Expected Behavior

Able to rename the files.

### OS Version

Early Access

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #2397 

Fixes another unfortunate side-effect of having the Window override the native key handling of Gtk.TreeView (in order to get desired search behaviour), which requires it to handle events in the `CAPTURE` phase.  The window now checks that the focus is not locked on the view before handling key presses.

### DIFF
--- a/src/View/Window.vala
+++ b/src/View/Window.vala
@@ -336,9 +336,11 @@ public class Files.View.Window : Hdy.ApplicationWindow {
         };
 
         key_controller.key_pressed.connect ((keyval, keycode, state) => {
+            // Handle key press events when directoryview has focus except when it must retain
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
Seems to work here

PR #2399, opened 2024-01-29
[Gtk4Prep] Views:  Use Gtk.GestureMultiPress

jeremypw:
Discontinues support for secondary button dragging for simplicity (avoids a lot of ugly code and probably rarely if ever used - it no longer behaves different from primary anyway).


danirabbit (approved):
Clicking, selecting, launching, rubberband, etc all seems to work in all views. Nice job and great cleanup!

PR #2400, opened 2024-01-30
[Gtk4Prep] ViewContainer: Use GestureMultiPress

jeremypw:
Handle special mouse keys with an eventcontroller.  

It is assumed that it is not necessary to block onward propagation of special mouse buttons as they are not handled by other Files widgets nor, presumably, Gtk widgets.

Needs testing by real special mouse.


danirabbit (approved):
Works :rocket:

PR #2402, opened 2024-01-31
[Gtk4Prep] SearchResults: Use EventControllers, disambiguate view name

jeremypw:


zeebok:
Code looks good, I will give this a test tomorrow!

zeebok (approved):
Code looks good and is what I see is reflected in testing. Not sure we want to address the behaviors I brought up here or save it for another PR. Up to you!

PR #2403, opened 2024-01-31
[Gtk4Prep] Searchresults: Use popover

jeremypw:


jeremypw:
@danirabbit Not sure if this is worth doing in Gtk3 or just do it during the Gtk4 port.

danirabbit (changes requested):
The pathbar is like severely broken in this branch. elements get oddly overlaid and state doesn't get updated and just a ton of `gdk_pixbuf_get_from_surface: assertion 'surface != NULL' failed` warnings

Screenshot from 2024-08-16 13 52 37

jeremypw:
@danirabbit The search also seems broken in `main` , at least on Wayland,  I'll have another look at this.  Using a Popover should be less problematic than a custom POPUP type window I should think, if it can be got working.

jeremypw:
I think I have fixed the issues with the placeholder - it seems to work on OS7.1 and OS8 (Wayland and X).

danirabbit:
@jeremypw I don't really understand why this branch contains any changes to breadcrumbs at all? That seems unrelated to using a Popover instead of a Window here?

jeremypw:
@danirabbit Not sure now - I'll go through it and try to split out any inessential changes.

PR #2404, opened 2024-02-02
[Gtk4Prep] AbstractDirectoryView: Use EventControllerMotion

jeremypw:
This handles changing the cursor depending on what the pointer is over.


danirabbit (approved):
works :+1:

PR #2405, opened 2024-02-02
[Gtk4Prep] Breadcrumbs: Use EventControllerMotion

jeremypw:
This implements tooltips on the breadcrumbs (which are not widgets)


danirabbit (approved):
Looks good! :tada:

PR #2406, opened 2024-02-02
[Gtk4Prep] Bookmarks: Use GestureMultiPress

jeremypw:
This controls bookmark activation and context menu


danirabbit (approved):
Buttons all seem to work as expected still :rocket:

PR #2407, opened 2024-02-02
[Gtk4Prep] StyleContext getters

jeremypw:
In Gtk4 style_context getters must only use the current state.  The colorize function in the IconRenderer was using a modified state and doesnt seem to be working anyway so was removed.


danirabbit (approved):
I don't see a different between this and main :shrug: :rocket:

PR #2408, opened 2024-02-02
AbstractDirectoryView: Use EventControllerScroll

jeremypw:
Scroll handler functions in AbstractDirectoryView merged.
Workaround needed because Gtk.EventControllerScroll swallows all scroll events in CAPTURE mode even if not handled and does not receive them at all in BUBBLE mode.  Therefore it is only active when a modifier is pressed.


danirabbit (approved):
Works here!

PR #2410, opened 2024-02-05
AbstractDirectoryView: claim button press when appropriate

danirabbit:
We accidentally broke selection helpers and maybe a few other things. Fix claiming the button press gesture instead of setting a variable that is never used


jeremypw (approved):
Thanks for fixing this!

PR #2412, opened 2024-02-17
[Regression] Views: Get correct clickzone information when scrolled.

### LINKED ISSUE #2411: Incorrect click coords in a scrolled view
### What Happened?

Clicking in a view that has been scrolled results in a click at the wrong coordinates. This seems to have started after #2399

### Steps to Reproduce

1. Open a folder with a lot of items
2. Scroll down and click on of those items
3. Notice that the selection appears both on the item you selected and on some item above

### Expected Behavior

Clicking an item should select only that item and not a second click in another location :)

### OS Version

Early Access

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #2411


danirabbit (approved):
Works as expected, thanks for the fix!

PR #2413, opened 2024-02-18
RenamerDialog: Use C_ instead of NC_

ryonakano:
The same fix with https://github.com/elementary/switchboard-plug-sound/pull/267


jeremypw (approved):
Thanks for fixing that!

PR #2422, opened 2024-03-10
[Gtk4Prep] Remove Gdk.beep ()

jeremypw:
Gdk.beep () does not exist in Gtk4

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -3141,8 +3141,6 @@ namespace Files {
                             update_selected_files_and_menu ();
                             if (!in_recent && is_writable) {
                                 if (selected_files.first () != null && selected_files.first ().next != null) {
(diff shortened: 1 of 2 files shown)


danirabbit:
Looks like it's just been moved to be per display https://valadoc.org/gtk4/Gdk.Display.beep.html

jeremypw:
@danirabbit  Ah, thanks. If we keep this, should it respect the `org.gnome.desktop.sound event-sounds` setting? I am not keen on in-app beeps when the user's attention doesn't need attracting - they do not give any further information.

danirabbit:
@jeremypw beep is used when there was some kind of input but the task can't be completed. Like when you're in an empty text entry and try backspace. It's acknowledging the input was accepted but there's nothing to do in response to it.

I think it should respect the settings for alert sounds and screen flashing automatically

jeremypw:
OK, thanks.  I'll leave this in then.

PR #2434, opened 2024-05-14
Release 8.0.0

danirabbit:
Todo:
- [x] Update app menu screenshot


danirabbit:
This is still building for Jammy in daily, so need to double check why this was bumped to 8 instead of maybe 6.6? or 7 maybe? Trying to get major version numbers to match the oldest release it builds for

danirabbit:
Ah might have been because of wallpaper portal, but we did release new portals for Jammy

jeremypw:
Are you reserving 8.x version numbers for Gtk4 ports? It is a rather large bump.

danirabbit:
@jeremypw Yeah I'm trying to match the major version number to the oldest release we're still building for. Just makes it easy to know at a glance what I need to build/test on. I think this should be okay to release to 7.1?

danirabbit:
Okay yup as far as I can tell this still works on OS 7.1 so bumped the release version down to 7.0.0

danirabbit:
@jeremypw Is this okay to release or any branches blocking currently?

jeremypw:
@danirabbit There are no blocking branches as far as I am concerned - please go ahead.

PR #2452, opened 2024-08-16
Do not activate "find" with space character

### LINKED ISSUE #2451: pressing spacebar no longer opens selected file
### What Happened?

I prefer to use spacebar to open files, as it's much closer to my laptop's touchpad. After the latest update and reboot (today) it no longer works. The cursor appears in an empty address bar (in place of breadcrumbs) but does nothing there (no search or path entry works). So I expect this to be bug, and would like to have the original behaviour back.

### Steps to Reproduce

1. Open Files and select some file (or even folder). The view doesn't seem to matter, happens in List or Columns view.
2. Press the `space` bar.
3. This is not what I wanted - cursor in useless place and no file / folder open.

### Expected Behavior

Space bar should open files / folders like it did yesterday. The current version reported:

```
$ io.elementary.files --version
io.elementary.files 7.0.0
```

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #2451

### DIFF
--- a/src/View/Window.vala
+++ b/src/View/Window.vala
@@ -347,9 +347,10 @@ public class Files.View.Window : Hdy.ApplicationWindow {
                 var mods = state & Gtk.accelerator_get_default_mod_mask ();
                 /* Use find function instead of view interactive search */
                 if (mods == 0 || mods == Gdk.ModifierType.SHIFT_MASK) {
-                    /* Use printable characters to initiate search */
+                    /* Use printable characters (except space) to initiate search */
+                    /* Space is handled by directory view to open file items */
                     var uc = (unichar)(Gdk.keyval_to_unicode (keyval));
-                    if (uc.isprint ()) {
+                    if (uc.isprint () && !uc.isspace ()) {
                         activate_action ("find", uc.to_string ());
                         return Gdk.EVENT_STOP;
                     }


danirabbit (approved):
Makes sense :+1:

janxkoci:
It does not actually activate the find function - the cursor just appears where find/path/breadcrumbs usually appear, but typing is not doing anything. It is basically just an empty text field that does nothing useful.

Using Ctrl+F to activate search, or Ctrl+L to focus path both work as expected.

jeremypw:
The search function also seems broken on Wayland and after making preparations for Gtk4 but that is a separate issue. I am working on a fix for that.

janxkoci:
I also noticed dragging the window by pathbar (breadcrumbs) no longer works and I have to drag by the remaining (little) parts of the headerbar. Not sure if it's related, but I think it broke around the same time. 

And sorry for late comment, I had some time off.

jeremypw:
@janxkoci Confirmed.  Will need another issue report.  Note you can drag the whole window by holding down the "Super" key before dragging (on OS8).  Normal dragging on the other controls in the header also works, but clicking on the pathbar is now grabbing focus.  Not sure why.

janxkoci:
I may report it after tomorrow's deadline, now I'm too busy, just taking a break for dinner. I know I can drag on the buttons too, like in all GTK apps (I missed that on macOS), but the path bar is still the largest widget, so would be cool to have that back.

jeremypw:
Using the current `main` branch on OS8 It seems that you can drag on the pathbar in some cases:
 - It is showing the "Search or Type Path" placeholder
 - It is showing the breadcrumbs and you click on the breadcrumbs

If you click on the white space after the breadcrumbs the it switches to edit mode and you cannot drag in edit mode.

Can probably fix this by switching on button release instead of button press and only switch to edit mode if no drag occurfed.

janxkoci:
@jeremypw I can confirm it behaves the same on OS 7, thanks for pointing it out.

PR #2458, opened 2024-08-23
Improve thumbnailing Part2: Only generate large thumbnails

jeremypw:
Cherry-picked and adapted from #2340

Results in smoother zooming to large icons.  No discernible performance impact with modern hardware.


danirabbit (approved):
We love a simplification and I can confirm makes zooming smoother

PR #2461, opened 2024-08-28
Improve thumbnaiing code Part1: Explicitly handle all FileMonitorEvents

jeremypw:
Also stop unnecessary signal being emitted when a file changes due to data being streamed into it as this does not currently cause the display to change.  We may want to handle this in future to auto update the file size column in ListView?

Part of #2340  code improvement


danirabbit (approved):
Thumbnails appear to still update when the file is changed. This makes sense to me!

PR #2462, opened 2024-08-28
Lose unused actions in AbstractDirectoryView; simplify

jeremypw:
Showing local and remote thumbnails is no longer controlled by the AbstractDirectoryView context menu so the related actions are no longer needed.

Do not need separate functions for local and remote icons - use one flag to indicate whether to show thumbnails

Extracted from #2340


danirabbit (approved):
Seems like a nice simplification and thumbnails seem to still react to pref changes here. Nice work!

PR #2463, opened 2024-08-28
PropertiesWindow: redesign permisssions

danirabbit:
Attemping to address https://github.com/elementary/stylesheet/issues/345

Uses icons that have `\` states and non-linked buttons to try to make it clearer which permissions are selected

## AFTER

Screenshot from 2024-08-28 13 03 12

## BEFORE

Screenshot from 2024-08-28 13 05 23

### DIFF
--- a/src/Dialogs/PropertiesWindow.vala
+++ b/src/Dialogs/PropertiesWindow.vala
@@ -996,9 +996,9 @@ public class Files.View.PropertiesWindow : AbstractPropertiesDialog {
             perm_grid.attach (owner_user_choice, 1, 0);
             perm_grid.attach (group_combo_label, 0, 1);
             perm_grid.attach (group_combo, 1, 1);
-            perm_grid.attach (l_perm, 0, 2);
-            perm_grid.attach (perm_code, 1, 2);
-            perm_grid.attach (permission_box, 0, 3, 2);
+            perm_grid.attach (permission_box, 0, 2, 2);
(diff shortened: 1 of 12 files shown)


src/Dialogs/PropertiesWindow.vala
```diff
@@ -981,23 +976,29 @@ public class Files.View.PropertiesWindow : AbstractPropertiesDialog {
...
-            perm_grid.attach (perm_button_other, 1, 5, 2, 1);
-            perm_grid.attach (l_perm, 1, 6, 1, 1);
-            perm_grid.attach (perm_code, 2, 6, 1, 1);
+            perm_grid.attach (owner_user_label, 0, 0);
+            perm_grid.attach (owner_user_choice, 1, 0);
+            perm_grid.attach (group_combo_label, 0, 1);
+            perm_grid.attach (group_combo, 1, 1);
+            perm_grid.attach (l_perm, 0, 2);
+            perm_grid.attach (perm_code, 1, 2);
+            perm_grid.attach (permission_box, 0, 3, 2);
```
jeremypw, line 1001: Is there a reason to alter the order of these elements? I think most users would use the buttons.  The permission code and value are kind of extra information and seems out of place in the middle.

jeremypw (commented):
Much better!  Just a query about the ordering.

jeremypw (approved):

PR #2464, opened 2024-08-30
Improve thumbnailing code Part 2: Clean up File.vala

jeremypw:
Extracted from #2340 

* Rewrite and merge some icon related functions
* Make some members into properties
* Sanitize requested dimensions in public functions rather than ignore.
* Assume dimensions valid in private functions
* Always reset thumbstate on full update
* Add explanatory comments to some functions

The naming of some functions and variables could be made less confusing but that will be done is another PR.


danirabbit (approved):
This all makes sense to me afaict and everything still seems to be working correctly. Thanks for doing clean up!

PR #2466, opened 2024-09-01
Fix drop onto unmounted drive bookmark Part 1

### LINKED ISSUE #2390: Drag N Dropping file on a disk that is not mounted makes the file disappear

### LINKED ISSUE #2465: Dropping onto mountable bookmark does not work correctly if location mounted after Files starts

jeremypw:
Fixes #2390 
Fixes #2465

 * `file_accepts_drop ()` returns false if location unmounted (fixes first issue)
 * Update the target location if mounted after Files started (fixes second issue)
 
Extracted from #2391  without last substantive commit to reduce diff.

The last substantive commit does not seem necessary to fix the linked issues.  An unmounted drive bookmark is created with an incorrect target but this PR prevents it from being dropped on and corrects the target when the drive is mounted.   

Correctly creating the bookmark will be fixed separately as it is possible it could cause a bug.


zeebok:
Both before and after trying to test this on OS8 stable, if I try to repro the first issue with a USB and drag a file to it, the file remains and a link to that file is created next to it. Nothing is list to a blackhole.

jeremypw:
@zeebok 
> Nothing is list to a blackhole

Not sure what that means - do you mean the original file does not disappear as reported in the original issue?

> drag a file to it
The issue occurs when dropping a file onto the sidebar bookmark of an unmounted drive/volume - is that what you did?  I just checked again and in `main`, the file disappears from its original location but the drive/volume remains unmounted.  Pressing `<Ctrl>Z` causes the file to reappear.  It is in fact moved to the user home directory instead of the drive/volume.  In this PR the drop just fails (for simplicity).  Automounting may be implemented later.

jeremypw:
I am running OS8 in a VM at the moment so its difficult to reproduce this issue there. I am testing with OS7.1.

jeremypw:
> the file remains and a link to that file is created next to it

This is what happens when a file is dragged and dropped into its own location. The drag icon should show a "link" emblem in that case.

zeebok:
> Not sure what that means - do you mean the original file does not disappear as reported in the original issue?

Sorry, typo, meant nothing was lost* in a blackhole from dragging and dropping.

> The issue occurs when dropping a file onto the sidebar bookmark of an unmounted drive/volume - is that what you did? I just checked again and in `main`, the file disappears from its original location but the drive/volume remains unmounted. Pressing `<Ctrl>Z` causes the file to reappear. It is in fact moved to the user home directory instead of the drive/volume. In this PR the drop just fails (for simplicity). Automounting may be implemented later.

Yep, I plugged in a USB drive, didn't let it mount, went to a source code folder and duplicated a file with a very obvious test name, then tried to drag it to the USB drive. It didn't try to move it and just made a link of the file that sat next to the original test file. I can try and snag a video if you'd like to see!

jeremypw:
That's strange - I'd like to see the video. 

> It didn't try to move it 

 I presume drag and drop is, in general working for you?  i.e. a drag started normally and you could drop the drag icon onto the USB bookmark OK?  But on dropping, a link was created in the original folder?

jeremypw:
Closing in favor of better solution

PR #2467, opened 2024-09-01
Search improvements Part1: Default to search mode when focussed

jeremypw:
TEST IN X ONLY

 Search window is broken in Wayland and needs replacing with (e.g.) Popover to fix.

* Ensures that the search icon shows after pressing `<Ctrl>F`
* Ensures placeholder shows when no text and remains in search mode (or switches to it) 
* Simplifies switching between search and navigate modes

Extracted from #2403 where not related to use of Popover.


danirabbit (approved):
Seems to be a lot simpler and works as expected afaict

PR #2468, opened 2024-09-02
Fix drop onto unmounted volume Part2: Set drop destination according to mount state

jeremypw:
Extracted from #2391 

Review and merge #2466 first please.

The main effect of this is that unmounted volumes are no longer highlighted as a drop destination when a file is dragged over them.

Note: Custom bookmarks pointing to locations on a mountable volume will still appear as drop destinations even if the volume is unmounted.  There does not seem to be a simple fix for this so it will be left for a later PR.


jeremypw:
This no longer seems required with the newer Part 1 branch

PR #2469, opened 2024-09-03
PropertiesWindow: handle permpanel failure case in permpanel widget

danirabbit:
Review with "hide whitespace changes"

* Move the failure to get permissions handling into the function that creates the perm grid. Simplifies packing it into the stack and `perm_grid` can be scoped down to that function
* Simplify making the grid insenstive. Can't use `get_children` in gtk4


jeremypw (approved):
All seems to work and makes sense

PR #2470, opened 2024-09-03
PopupMenuBuilder: use GLib.Action for MountableRow

danirabbit:
Initial work to start porting PopupMenuBuilder to use GLib.Actions. Based on work in https://github.com/elementary/files/pull/2082

Remove unused `add_mount`


jeremypw (approved):
This creates a new warning when about mapping to a non-toplevel window when running under Wayland (but not X11) - but this will probably disappear when converted to GLib.Menu so not worth investigating.  It does not affect functionality.

PR #2473, opened 2024-09-04
AbstractMountableRow: use add_with_action_name directly

danirabbit:
Get rid of a few functions that are only used once. The action names don't exist in the library anyways, so I think this makes more sense?


jeremypw (approved):
Makes sense

PR #2477, opened 2024-09-06
BookmarkRow: delete focused bookmark with delete key

danirabbit:


jeremypw:
These bookmarks only work as expected if the sidebar has keyboard focus, which is not straightforward to achieve at the moment.  The sidebar item may be highlighted but pressing "Del" will rename delete a "real" file in the view if the the view in has focus.   Starting renaming is not a great problem but unexpected deletion is not ideal even though it may be undoable.

The sidebar can be focused by right-clicking on an item to show the menu (in which case you might as well use the menu options) or repeatedly pressing `<Shift>Tab` until the correct widget is focused.

jeremypw:
It may be better to hold on this until #2353 is merged?  I'll set it open for review as the subject has attention.

jeremypw:
@danirabbit What are your thoughts on using modified shortcuts for sidebar actions? e.g. `<Alt>Del`, `<Alt>F2` instead of the same ones as the Directory View?

jeremypw (changes requested):
I do not think it a good idea to have the same accelerator to delete two different things.

jeremypw:
@danirabbit Can this be converted to draft or closed? I think there are some significant issues to be resolved.

danirabbit:
Sorry I haven't been able to follow up on this. Converted to draft

PR #2480, opened 2024-09-09
Use Popover for SearchResults: Part 1

jeremypw:
Extracted from #2403

The focusing of the popover/breadcrumbs is still broken as it is in `main` when running under Wayland.  This will be fixed separately - see #2481


danirabbit (changes requested):
This seems to steal focus on both X11 and Wayland

jeremypw:
@danirabbit The pre-existing focus issue on Wayland was fixed (I hope) in #2481.

jeremypw:
Closing in favour of the combined PR, which has a reasonable diff and separate commits for different parts of the code.

PR #2481, opened 2024-09-09
Use Popover for SearchResults: Part 2

jeremypw:
This fixes breadcrumb behaviour after converting the search window to a popover.

Fixing the issues with `main` with a Gtk.Window search window under Wayland proved problematic so probably not worth the effort if the popover branch is going to be merged.  Using a Popover should be window manager agnostic.


jeremypw:
Closing in favour of the combined PR, which has a reasonable diff and separate commits for different parts of the code.

PR #2482, opened 2024-09-10
Replace PopupMenuBuilder with GLib.Menu

danirabbit:
Replaces the PopupMenuBuilder class with GLib.Menu. Afaict, all the menus are structured the same with separators in the same spot.

Confirmed with ebassi on mastodon that Gtk.PopoverMenu doesn't currently have a way to style items as destructive. So the alternative here is to make our own PopoverMenu class which tbh I don't think is worth it. He said he's open to it upstream though


jeremypw (approved):
Code looks good and menus remain similar to `main` apart from the destructive styling in one case.

PR #2514, opened 2024-11-26
CommonJob: Move scan_dir and this depending functions to Vala

tintou:
Also make use of the use of class-based objects to add virtual accessors.


jeremypw:
Added to https://github.com/jeremypw/files-dogfood-7.1 for further testing.

jeremypw (approved):
Everything still seems to work.  Thanks for your work!

PR #2522, opened 2024-12-31
Fix regression in path entry when in search mode

### LINKED ISSUE #2523: Entering a path into the pathbar no longer possible
### What Happened?

It used to be possible to enter a path or protocol into the pathbar but now as soon as a directory separator is entered, the current entry is overwritten by the current path.

### Steps to Reproduce

1. Start typing a path
2. Observe that the entry is lost when a directory separator is entered

### Expected Behavior

It is possible to enter a path or uri into the pathbar.

### OS Version

7.x (Horus)

### Session Type

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #2523 

 The current entry text is persisted when the mode switches from search to navigate under these circumstances.

The switch now occurs on entry of a colon as well as a directory separator.

It is ensured that the search is inactive after it is cancelled.


src/View/Widgets/LocationBar.vala
```diff
@@ -103,7 +103,6 @@ namespace Files.View.Chrome {
 
         private void on_search_results_exit (bool exit_navigate = true) {
             /* Search result widget ensures it has closed and released grab */
-            search_results.is_active = false;
```
zeebok, line 106: Why remove this? Is there a risk of this being followed by `after_bread_focus_out_event` and causing undefined behavior?

jeremypw: Its because SearchResults.is_active is now readonly.

jeremypw: I'll check the logic wrt focus_out event though - its hard to remember the details now.

jeremypw: SearchResults sets itself inactive *before* emitting the `exit` signal and this is handled before the `after_bread_focus_out_event` so I do not think there is an issue of undefined behaviour.  Tbh the pathbar/searchresults code is very difficult to follow and is likely to be re-written with new widgets for the Gtk4 port.

zeebok (approved):

PR #2526, opened 2025-02-02
Fix drop on mounted after start

### LINKED ISSUE #2390: Drag N Dropping file on a disk that is not mounted makes the file disappear

### LINKED ISSUE #2465: Dropping onto mountable bookmark does not work correctly if location mounted after Files starts

jeremypw:
Fixes #2465 

Sets target file to null when bookmark not mounted and not accepts drop of target_file if null.

NOTE: Ignore first commit - this solution is not reliable as the File property `is_mounted` returns false for local files at the moment. Better to rely on presence of mount object in bookmark.


danirabbit (commented):
I guess this technically fixes the issue but I don't think this is the expected behavior. It seems like the expectation is that the drive would be mounted so that the drop operation succeeds.

If that's significantly more difficult, I can see this as like a hotfix, but I don't think it really solves the problem as far as giving the intended result

jeremypw:
@danirabbit Fair enough - I'll work on automounting.  I note that there is one issue that this does completely fix? Are you willing to merge on that basis and I'll submit a further PR in due course?

danirabbit (approved):
Yeah, I'm okay with merging this for now! Sorry I thought I left an "approve" comment. Should I open an issue report about mounting for a drop operation?

jeremypw:
I've left a comment on #2390 making it clear that the mount should be automounted so I do not think another report is required.

Issue #1794, opened 2021-08-18
Thumbnail generation and large folder loading could be faster in general


abodedis:
## Problem

Thumbnail generation and large folder loading could be faster in general.

## Proposal

As of this writing, Elementary OS generates image thumbnails faster than the default file managers in other distros, but the wish is to load large folders instantaneously, or at the least, fast enough that the user does not notice the wait. This will be useful for artists and those who manage thousands of images in a folder and would rather not wait several seconds or minutes for a large folder to load, and do not want to install another file manager because Files is great!

The image thumbnailer and/or whatever loads the files could utilize multiple threads and/or the GPU. Currently, it appears that Files does most, if not all, of its processing on only one CPU thread.  Most users these days have multi-core, multi-threaded processors and powerful GPUs.

If whatever code that currently loads the files is based on ancient code/libraries, then that code could be rewritten to take advantage of modern hardware.

jeremypw:
The rate limiting step is not the thumbnailing - that is already optimised by only requesting thumbnails when the item is actually visible.  The main problem is Gtk.IconView being very slow and inefficient in layout out and displaying large folders.  I understand the performance in Gtk4 is generally improved so hopefully that will ameliorate the issue when we can move to it.  At the moment working with large folders is better done ListView which is much faster in laying out and drawing.

I did some work on a custom widget here - https://github.com/jeremypw/widget-grid - which is almost unaffected by folder size but there is no real appetite in elementary to develop and support a complicated custom widget.

jeremypw:
I did test on a large folder of files recently and found Files loaded it much more slowly (from a spinning disc) than I remember it used to, even using ListView.  Need to investigate what has caused this regression.

jeremypw:
I think the slow down may be related to recent porting of the custom model used to Vala and basing it on Gtk.TreeStore - it seems that the time take to add files to the model increases quadratically with number of files even with sorting disabled - not sure why yet.  The maximum reasonable usable folder size with a spinning disc is about 1000 - 2000 items now.

PR #2534, opened 2025-02-26
Speed up loading large folders

jeremypw:
Addresses #1794 

This PR uses two strategies for speeding up loading/model creation
 - Add the initially loaded files in one hit in the `done-loaded` signal handler instead of individually
 - Create the file-treerow map asynchronously after the model has loaded.

These strategies decrease the time that the UI is frozen significantly, particularly for ListView.
For IconView the effect is less marked because the majority of the time is taken in actually laying out the view; this largely occurs inside Gtk.IconView and still freezes the UI. There is little that can be done about this until IconView is replaced.

To load and display 20,000 small files on my system
ListView: 22 seconds -> 3 seconds
IconView: 52 seconds -> 29 seconds

TODO:
 - [x] Fix loading subdirectories and dummy rows


jeremypw:
I am inclined to close this for now as although it speeds up loading, it does not address performance related to other operations like sorting updating etc and it is better to address the performance project as a whole after the Gtk4 port and conversion to dynamic views.

PR #2538, opened 2025-03-06
Bookmarkrow cleanup

jeremypw:
See description of individual commits.

### DIFF
--- a/src/View/Sidebar/BookmarkRow.vala
+++ b/src/View/Sidebar/BookmarkRow.vala
@@ -149,7 +149,6 @@ public class Sidebar.BookmarkRow : Gtk.ListBoxRow, SidebarItemInterface {
         label_stack.visible_child_name = "label";
 
         icon = new Gtk.Image.from_gicon (gicon, Gtk.IconSize.MENU);
-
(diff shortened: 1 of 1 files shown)


src/View/Sidebar/BookmarkRow.vala
```diff
@@ -149,7 +149,6 @@ public class Sidebar.BookmarkRow : Gtk.ListBoxRow, SidebarItemInterface {
         label_stack.visible_child_name = "label";
 
         icon = new Gtk.Image.from_gicon (gicon, Gtk.IconSize.MENU);
-
```
danirabbit, line 157: I don't love removing blank space between two different objects personally

danirabbit (approved):
One tiny non-blocking comment but otherwise LGTM

src/View/Sidebar/BookmarkRow.vala
jeremypw: Well, these could be regarded as parts of one compound object (icon_label_grid) but in that case I should have removed the preceding blank line as well. I'll revert this change for now.

PR #2539, opened 2025-03-06
BookmarkRow: Replace Gtk.Menu

jeremypw:


danirabbit (commented):
I really don't like doing this prematurely in Gtk3 because it breaks styles, it breaks items with accels, and it's just such a tiny diff that it feels not really worth it. It's a really easy review in a final GTK4 branch and then we don't have to deal with all the regressions

PR #2540, opened 2025-03-07
Jeremypw/gtk4 prep/bookmark list box cleanup

jeremypw:
Use equivalent code that will work in Gtk4 as well as Gtk3 as far as possible

### DIFF
--- a/src/View/Sidebar/BookmarkListBox.vala
+++ b/src/View/Sidebar/BookmarkListBox.vala
@@ -195,12 +195,11 @@ public class Sidebar.BookmarkListBox : Gtk.Box, Sidebar.SidebarListInterface {
                                        int pos = 0) {
 
         int pinned = 0; // Assume pinned items only at start and end of list
-        foreach (unowned var child in list_box.get_children ()) {
-            if (((BookmarkRow)child).pinned) {
-                pinned++;
-            } else {
-                break;
-            }
+        int index = 0;
+        Gtk.ListBoxRow? row = list_box.get_row_at_index (index++);
+        while (row != null && ((BookmarkRow) row).pinned) {
+            pinned++;
+            row = list_box.get_row_at_index (index++);
         }
 
         if (pos < pinned) {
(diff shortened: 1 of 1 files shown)


src/View/Sidebar/BookmarkListBox.vala
```diff
@@ -195,12 +195,11 @@ public class Sidebar.BookmarkListBox : Gtk.Box, Sidebar.SidebarListInterface {
...
 
         int pinned = 0; // Assume pinned items only at start and end of list
-        foreach (unowned var child in list_box.get_children ()) {
-            if (((BookmarkRow)child).pinned) {
-                pinned++;
-            } else {
-                break;
-            }
+        int index = 0;
+        Gtk.ListBoxRow? row = list_box.get_row_at_index (index++);
```
danirabbit, line 199: You start the index at 0 but then immediately make it 1 here. That seems a bit overcomplicated? Why not just start the index where you need it to start? :sweat_smile:

danirabbit (approved):
Just a tiny comment/question/nitpick otherwise this LGTM

src/View/Sidebar/BookmarkListBox.vala
jeremypw: Unless I'm mistaken, when `++` is post-fixed it operates *after* the variable is used so the current code is equivalent to 
```
Gtk.ListBoxRow? row = list_box.get_row_at_index (index);
index = index + 1;
```

(when the operator is pre-fixed it increments the variable *before* using it)

I'll double-check though and see if there is a clearer, equally concise way.

jeremypw:
@danirabbit 
From Vala doumentation:

> Increment and decrement operations with implicit assignment. These take just one argument, which must be an identifier of a simple data type. The value will be changed and assigned back to the identifier. These operators may be placed in either prefix or postfix positions - with the former the evaluated value of the statement will be the newly calculated value, with the latter the original value is returned.

So with the postfix operator the lookup expression uses the original (pre-incremented) value as I expected.

jeremypw:
Now used the pre-fix operator which requires the same number of lines and is clearer. Not sure why I used the post-fix tbh

PR #2541, opened 2025-03-07
Jeremypw/gtk4prep/devicelistbox cleanup

jeremypw:
Use equivalent code that also compiles in Gtk4 where possible

### DIFF
--- a/src/View/Sidebar/DeviceListBox.vala
+++ b/src/View/Sidebar/DeviceListBox.vala
@@ -173,11 +173,15 @@ public class Sidebar.DeviceListBox : Gtk.Box, Sidebar.SidebarListInterface {
     }
 
     public override void refresh_info () {
-        list_box.get_children ().@foreach ((item) => {
-            if (item is AbstractMountableRow) {
-                ((AbstractMountableRow)item).update_free_space ();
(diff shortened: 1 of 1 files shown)


src/View/Sidebar/DeviceListBox.vala
```diff
@@ -173,11 +173,15 @@ public class Sidebar.DeviceListBox : Gtk.Box, Sidebar.SidebarListInterface {
     }
 
     public override void refresh_info () {
-        list_box.get_children ().@foreach ((item) => {
-            if (item is AbstractMountableRow) {
-                ((AbstractMountableRow)item).update_free_space ();
+        int index = 0;
+        Gtk.ListBoxRow? row = list_box.get_row_at_index (index++);
```
danirabbit, line 177: Same question as another review. It seems odd to me to start at 0 and then immediately increment

danirabbit (approved):
Same little question/nitpick as the other one, otherwise makes sense to me

src/View/Sidebar/DeviceListBox.vala
jeremypw: Now using prefix operator which is clearer.

PR #2550, opened 2025-03-14
Files.IconInfo: Replace Gtk.IconInfo

jeremypw:
This makes Files.IconInfo functional while maintaining the use of pixbufs. 

While there are better solutions in Gtk4 this minimises changes in users.


jeremypw:
As this fixes CI it makes sense to consider this for merging into the skeleton first.

PR #2556, opened 2025-03-19
AbstractDirectoryView: Base off Widget, use stack as main widget, use stack to show empty message

jeremypw:
Cannot subclass scrolled window so embed this in a Widget.  Instead of customized drawing use a label widget and stack to show the empty message.


danirabbit:
Why not do this for merging into main? Feels kind of counterproductive to do this into some kind of mega branch?

PR #2564, opened 2025-03-20
Renamer: Avoid listbox get_children

danirabbit:
There is no more `ListBox.get_children ()` in GTK4

There's one place I didn't replace here and it's the block starting at line 154. I suppose we could iterate over the list and add to a counter to get the number of children? Alternatively if we switched this whole thing to use a ListModel we could get the number of items in the model, but we'd need to add a "new_name" property to `Files.File`. Thoughts @jeremypw?


jeremypw:
I thought I had already rewritten this somewhere avoiding getting the number of children but I can't seem to locate it atm.  If I can't find it I'll do it again.

jeremypw:
As we only add rows in one place and never remove them, the simplest solution is to just count them as they are added.  Using a model would be more elegant if rows are added and removed adhoc.

jeremypw (approved):

PR #2565, opened 2025-03-21
AbstractDirectoryView: use stack to show empty message

jeremypw:


src/View/AbstractDirectoryView.vala
```diff
@@ -282,12 +285,35 @@ namespace Files {
         public signal void path_change_request (GLib.File location, Files.OpenFlag flag, bool new_root);
         public signal void selection_changed (GLib.List<Files.File> gof_file);
 
+        static construct {
```
danirabbit, line 288: You can just use Gtk.Bin here there's Granite.Bin in Gtk4 as a drop-in replacement

src/View/AbstractDirectoryView.vala
```diff
@@ -378,8 +402,7 @@ namespace Files {
         }
 
         private void set_up_directory_view () {
-            set_policy (Gtk.PolicyType.NEVER, Gtk.PolicyType.AUTOMATIC);
-            set_shadow_type (Gtk.ShadowType.NONE);
+            scrolled_window.set_policy (Gtk.PolicyType.NEVER, Gtk.PolicyType.AUTOMATIC);
```
danirabbit, line 405: Can we just set hscrollpolicy here since vscroll is automatic by default and we're only changing the hscroll?

jeremypw:
I've switched to using an Overlay as we need to keep the view in focus in order to maintain the behaviour without a lot of hassle.  I've incorporated @danirabbit 's comments into the overlay branch.

PR #2568, opened 2025-03-23
Gtk4 Prep - Use dialog.response.connect () pattern

jeremypw:
The dialog response is consistently handled in a closure connected to the `response` signal.

Where possible dialog.present () is also used but it seems that in Gtk3, this does not block the main loop even if the dialog is flagged as modal.  In cases where we need the mainloop to wait for a result from the dialog, `dialog.run ()` is still called.  I assume that in Gtk4, this can be changed to `dialog.present ()` and the `modal` flag will be observed.

### DIFF
--- a/libcore/FileUtils.vala
+++ b/libcore/FileUtils.vala
@@ -179,6 +179,7 @@ namespace Files.FileUtils {
             return true;
         }
 
+        bool success = false;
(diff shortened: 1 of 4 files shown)


danirabbit (approved):
I don't think `present` is main loop blocking in GTK4, so we'll need to do something else still in the future for those cases. But this does get us a bit closer I think so :+1:

PR #2575, opened 2025-04-23
Show context menu on all areas of sidebar

### LINKED ISSUE #2574: Context menu for Storage items only shows if secondary click on eject icon
### What Happened?

Tried to show properties of a mounted storage item by right-clicking on the item - nothing appeared.  Found that right-clicking on the eject icon showed the required menu.  

### Steps to Reproduce

1. Secondary-click on a mounted storage item name or levelbar.
2. Nothing happens
3. Secondary-click on the eject icon
4. Context menu shows

### Expected Behavior

The context menu should show where the item is clicked (the same as network items and bookmarks)

### OS Version

7.x (Horus)

### Session Type

### Software Version

Compiled from git

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #2574 

There's some weirdness going after converting to an eventcontroller to handle button clicks.  Only some children of the content grid are responsive and no obvious reason was found.  A simple but temporary solution is to insert an EventBox as the immediate child of the row.  This will need to be removed on porting to Gtk4 but hopefully event controllers will work better in Gtk4 and the fix will no longer be needed :crossed_fingers:

### DIFF
--- a/src/View/Sidebar/BookmarkRow.vala
+++ b/src/View/Sidebar/BookmarkRow.vala
@@ -160,11 +160,13 @@ public class Sidebar.BookmarkRow : Gtk.ListBoxRow, SidebarItemInterface {
         content_grid = new Gtk.Grid ();
         content_grid.attach (icon_label_grid, 0, 0);
 
-        add (content_grid);
+        var event_box = new Gtk.EventBox ();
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
Works here!

PR #2592, opened 2025-06-14
Fix filechooser regression for SAVE and SAVE_AS action

### LINKED ISSUE #2591: File Chooser no longer pre-populates file name when saving
### What Happened?

After the commit to use a validated entry, the file name in save dialogs no longer pre-populates the entry

### Steps to Reproduce

1. Use an app that uses the file chooser portal for its save file dialog. (In my case I was trimming a video with Footage)
2. See that after that commit, the file name entry is no longer pre-populated

### Expected Behavior

File names should still be pre-populated

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Following replacement of the native entry for a Granite.ValidatedEntry, the save and save_as actions no longer work because the chooser is still looking for the original entry for details.

This is fixed by syncing the validated entry content and activate action to the original entry and keeping a reference to the original entry.  Pre-existing or auto-filled text in the gtk entry is mirrored to the validated entry.

Not very elegant but works.  Looking for a better way...

Also fixes #2591

### DIFF
--- a/filechooser-portal/FileChooserDialog.vala
+++ b/filechooser-portal/FileChooserDialog.vala
@@ -380,7 +380,7 @@ public class Files.FileChooserDialog : Hdy.Window, Xdp.Request {
             entry_parent.remove (gtk_entry);
             validated_entry = new Granite.ValidatedEntry ();
             validated_entry.set_placeholder_text (_("Enter new filename"));
-            validated_entry.bind_property ("text", gtk_entry, "text", BindingFlags.SYNC_CREATE);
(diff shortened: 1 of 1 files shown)


jeremypw:
@tintou I tried changing the `get_file` method to use the validated entry but couldn't get it to work. As you wrote the original portal are you able to advise?

jeremypw:
Closing as this relates to a regression for a merge that has now been reverted.

PR #2596, opened 2025-06-16
Prepare for release 7.1.3

jeremypw:
Need another release soon as there was a released regression in drag and drop.

### DIFF
--- a/data/files.metainfo.xml.in.in
+++ b/data/files.metainfo.xml.in.in
@@ -66,7 +66,6 @@
       <description>
         <p>Minor updates:</p>
         <ul>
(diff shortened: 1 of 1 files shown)


danirabbit:
@jeremypw I'd be in favor of reverting #2586 for now so we can maybe get some better testing on it and figure out the bugs before re-merging

jeremypw:
> @jeremypw I'd be in favor of reverting #2586 for now so we can maybe get some better testing on it and figure out the bugs before re-merging

Agreed, even though I fixed the regression you noticed there may be further unexpected issues or a better way of fixing the existing ones.

PR #2601, opened 2025-06-19
Fix regression in copy/paste and DnD

jeremypw:
Fixes regression caused by #2595

### DIFF
--- a/libcore/DndHandler.vala
+++ b/libcore/DndHandler.vala
@@ -330,6 +330,7 @@ namespace Files {
                     target = FileUtils.sanitize_path (target, null, keep_protocol);
                     //Leave it to recipient to quote if required when receiving a uri list by DnD
                     //Terminal app applies quotes but not text entries for example
+                    sb.append (target);
                     sb.append ("\r\n"); /* Drop onto Filezilla does not work without the "\r" */
                 });
             } else {


zeebok:
- [x] Drop image file into Github comment on Chrome works (Not on Firefox which inserts the filename)
- [x] Drop image file onto Code text editor inserts filename (unquoted)
- [x] Drop text file onto Code Welcome screen opens as doc
- [x] Drop text file onto Code doc inserts filename (unquoted)
- [x] Drop image file onto elementary Terminal shell inserts filename (quoted)
- [x] Drop image file onto elementary Terminal tab navigates tab to parent folder 
 > ^^^ This does cause a rapid succession of notifications since Files remains focused; might be nice to bypass the notifications or to pull focus in the future
- [x] Drop folder onto elementary Terminal tab navigates tab to folder
 > ^^^ This does cause a rapid succession of notifications since Files remains focused; might be nice to bypass the notifications or to pull focus in the future
- [x] Drop folder onto elementary Terminal add button opens new tab at folder
- [x] Drag and drop file onto its parent
> How did you test this? If I perform this drag and drop from one Files window to another, nothing happens. Drag and drop to the same parent folder on the side bar makes a copy
- [x] Copy paste file onto its parent creates a copy (renamed)
- [x] Copy paste a file into another folder creates a copy (same name)
- [x] Cut paste a file into another folder moves it

zeebok:
@jeremypw Did the testing, it seems pretty reliable for me on OS8/Wayland

jeremypw:
I'll check what the DnD behaviour on OS8/Wayland/VM was for an earlier (unbroken) release.

jeremypw:
V7.1.0 behaves the same for me when dragging onto its parent using OS8/Wayland in a VM - it fails. So its not due to recent changes and can be investigated later.

jeremypw:
I'll leave unmerged for a little while in case anyone else wants to hammer it. I'll merge tomorrow.

jeremypw:
Drag and drop onto parent works for Classic session on OS8 so there seems to be a bug somewhere in Gtk3/Wayland/VM  in relation to DnD it seems a certain signal is not being generated.

jeremypw:
For some reason on Wayland the Gdk.Context associated with the drag does *not* include `Gdk.DragAction.LINK` whereas with X it does (the drag is set up in Code to allow linking). As linking is the only allowed action when dropping onto the same folder nothing happens in Wayland whereas a link is created in X :shrug:

PR #2610, opened 2025-07-05
Date format: expose setting

jeremypw:
Any setting that is actively supported should be accessible through the UI.  A user could have reason to change the datetime format for a particular folder/purpose.

For now, the existing three options are exposed as radiobuttons appended to the AppMenu.  

Screenshot from 2025-09-19 15 35 04


src/View/Widgets/AppMenu.vala
```diff
@@ -112,6 +113,21 @@ public class Files.AppMenu : Gtk.Popover {
...
         show_remote_thumbnails.get_style_context ().add_class (Gtk.STYLE_CLASS_MENUITEM);
 
+        ///TRANSLATORS The format of the date (possibly with time) shown in the Modified column of the file view
+        var datetimeformat_label = new Gtk.Label (_("Date & Time Format"));
+        datetimeformat_combo = new Gtk.ComboBoxText ();
+        datetimeformat_combo.append_text (DateFormatMode.ISO.to_string ());
+        datetimeformat_combo.append_text (DateFormatMode.LOCALE.to_string ());
+        datetimeformat_combo.append_text (DateFormatMode.INFORMAL.to_string ());
+        //TODO Add a custom format wizard? Or a detailed informat format? Or "days ago" format?
+        datetimeformat_combo.active = (int) Files.Preferences.get_default ().date_format;
```
ryonakano, line 123: I'm not sure if a combobox is appropriate because we're already opening an popover here so resulting a duplicated popover.

What do you think about using a radio button like we do in Maps?

スクリーンショット 2025-09-18 20 55 35

libcore/Enums.vala
```diff
@@ -253,4 +253,41 @@ namespace Files {
...
+                case LOCALE:
+                    return _("Locale");
+                case INFORMAL:
+                    return _("Informal");
+                default:
+                    assert_not_reached ();
+            }
+        }
+
+        public static DateFormatMode from_string (string format) {
```
ryonakano, line 277: It looks like this method is not called anywhere, is this expected?

src/View/Widgets/AppMenu.vala
jeremypw: Thanks - I'll try that.

libcore/Enums.vala
jeremypw: Ah, this is because the setting was originally a string but its now an enum so this is not needed any longer. I'll remove.

src/View/Widgets/AppMenu.vala
```diff
@@ -139,6 +157,37 @@ public class Files.AppMenu : Gtk.Popover {
...
+                app_settings.set_enum ("date-format", DateFormatMode.LOCALE);
+            } else {
+                app_settings.set_enum ("date-format", DateFormatMode.INFORMAL);
+            }
+        }
+
+        iso_button.toggled.connect (() => on_dateformat_changed ());
+        locale_button.toggled.connect (() => on_dateformat_changed () );
+        informal_button.toggled.connect (() => on_dateformat_changed ());
+
```
ryonakano, line 190: I think the following code would be better rather than creating an local method:

```vala
        iso_button.toggled.connect (() => {
            if (iso_button.active) {
                app_settings.set_enum ("date-format", DateFormatMode.ISO);
            }
        });
        locale_button.toggled.connect (() => {
            if (locale_button.active) {
                app_settings.set_enum ("date-format", DateFormatMode.LOCALE);
            }
        });
        informal_button.toggled.connect (() => {
            if (informal_button.active) {
                app_settings.set_enum ("date-format", DateFormatMode.INFORMAL);
            }
        });
```

src/View/Widgets/AppMenu.vala
```diff
@@ -9,6 +9,7 @@ public class Files.AppMenu : Gtk.Popover {
     private Gtk.Button zoom_default_button;
     private Gtk.Button zoom_in_button;
     private Gtk.Button zoom_out_button;
+    private Gtk.ComboBoxText datetimeformat_combo;
```
ryonakano, line 12:
```suggestion
```

Seems like a leftover

libcore/FileUtils.vala
```diff
@@ -615,25 +619,18 @@ namespace Files.FileUtils {
...
-            format = dateformatmode;
-        }
-
-         switch (format) {
-            case "locale":
+        switch (format) {
+            case DateFormatMode.LOCALE:
                 return dt.format ("%c");
-            case "iso" :
+            case DateFormatMode.ISO :
```
ryonakano, line 633:
```suggestion
            case DateFormatMode.ISO:
```

We should unify if we have a space here with `case DateFormatMode.LOCALE:` and `default:`

libcore/Enums.vala
```diff
@@ -253,4 +253,27 @@ namespace Files {
...
+                case INFORMAL:
+                    return _("Informal");
+                default:
+                    assert_not_reached ();
+            }
+        }
+
+    }
+
 }
```
ryonakano, line 279:
```suggestion
        }
    }
}
```

I (possibly personally) prefer not having newlines here.

src/View/Widgets/AppMenu.vala
```diff
@@ -112,6 +112,17 @@ public class Files.AppMenu : Gtk.Popover {
...
+        ///TRANSLATORS The format of the date (possibly with time) shown in the Modified column of the file view
+        var datetimeformat_header = new Granite.HeaderLabel (_("Date & Time Format"));
+
+        var iso_button = new Gtk.RadioButton.with_label (null, DateFormatMode.ISO.to_string ());
+        iso_button.get_style_context ().add_class (Gtk.STYLE_CLASS_MENUITEM);
+        var locale_button = new Gtk.RadioButton.with_label_from_widget (iso_button, DateFormatMode.LOCALE.to_string ());
+        locale_button.get_style_context ().add_class (Gtk.STYLE_CLASS_MENUITEM);
+        var informal_button = new Gtk.RadioButton.with_label_from_widget (iso_button, DateFormatMode.INFORMAL.to_string ());
+        informal_button.get_style_context ().add_class (Gtk.STYLE_CLASS_MENUITEM);
+        //TODO Add a custom format wizard? Or a detailed informat format? Or "days ago" format?
```
ryonakano, line 124: I think we should simply remove this TODO and file an issue if we won't address it in this PR but really want to in the future. TODO comments are often buried in projects that have many source codes and are developed by multiple developers.

jeremypw: OK, I've now limited this PR to exposing the setting.  We can consider other formats later.

ryonakano (approved):
Thank you!

PR #2616, opened 2025-07-10
FileOperations: Re-escape uris from Clipboard or DNDHandler before use in fileoperations

### LINKED ISSUE #2615: Files with names containing `#` characters are not transferred correctly.
### What Happened?

Tried to move a file named "### Test ###" into a subfolder. The operation failed with a message that it not possible to move a folder into one of its children.  The parent folder seems to be moved rather than the file.

### Steps to Reproduce

1. Open a folder containing a subfolder and a file containing `#` in the name
2. Try to move the file in to the subfolder

### Expected Behavior

The file gets moved like any other file

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #2615

After #2268 files being copied or dragged have their name sanitized before being given to the recipient so that they appear correctly when pasted/dropped as text.  However, they were not being re-escaped when used in a file operation and this caused the operation to fail for some files.  This PR fixes that.

After this PR, copy/dragging a file should work both to a text document or terminal and to another folder in Files (or Dolphin or Nautilus at least).

Note that Dolphin does not unescape filenames when transferred as text but Nautilus does.


vjr (dismissed):
The code LGTM but I can't confirm (repro) the fix works...

lenemter:
When dragging a file with '#' in its name into the Terminal, the pasted path gets cut off after '#'

jeremypw:
> When dragging a file with '#' in its name into the Terminal, the pasted path gets cut off after '#'

Nautilus seems to have the same problem (with Pantheon Terminal).

jeremypw:
Cut and paste works correctly.

jeremypw:
This appears to be a bug in Terminal not Files - the "sanitization" of dropped uris is causing this truncation.

jeremypw:
See #877 for Terminal fix.

zeebok:
Hmmm just tested in a VM. It does fix D&Ding the `##test##.txt` to a subfolder, but dragging into Terminal or Code (or an mp3 into Music) still doesn't seem to work. Terminal only shows the folder of the file, Code crashes, and Music sees the mp3 as invalid.

jeremypw:
@zeebok Thanks for the review - it is hard to fix this problem without introducing problems elsewhere hence the need for multiple testers. There is a Terminal PR to fix handling `#` there (https://github.com/elementary/terminal/pull/877). I'll look into the other instances you mentioned to see where the problem lies.

jeremypw:
@zeebok I cannot reproduce the problem with pasting into Code I am afraid (using current master version of Code on OS8).

I notice that pasting from other file managers into Terminal also truncates confirming that it is a Terminal bug.

jeremypw:
I can conform that pasting into Music is problematic.  I suspect it is because Files is including the `files://` protocol for local files.

jeremypw:
Hmm, I now cannot reproduce the Music issue after recompiling Music from source, neither on version 8.0.0 or version 7.0.0 of Music, at least with files not containing `#` so I am not sure what version of Music I was running before.   However, I cannot drop files onto a similar third-party music app (Amberol) but Nautilus e.g. can.  I cannot see any difference in the URI of a file dropped by Files and Nautilus so that is puzzling.

There is a problem similar to Terminal in Music where the URI of files is truncated at the first occurrence of `#` but in that case it does not occur with other file managers.

jeremypw:
@zeebok I'd be grateful for your review after the latest commits.  I noticed that Files was only providing the TargetType TEXT_URI_LIST to other apps - I've fixed that although I don't think it is related to this issue necessarily. I've also tried to clarify the code a little with better names and comments.
Also, I've reverted the stringbuilder function to work as it did in version 6.x i.e. it only sanitizes the path when providing a TargetType.STRING.

I suggest you test using Music compiled from source as the current release has issues with dropped files. I also suggest testing with Terminal compiled from https://github.com/elementary/terminal/pull/877.

I could not get dropping onto Amberol to work - maybe because it is a Gtk4 app and DnD is done differently?  I think any issues with 3rd party apps will have to be left for another PR.

libcore/DndHandler.vala
```diff
@@ -294,42 +294,53 @@ namespace Files {
...
                                  sb.data);
 
         }
 
-        // Used when copying a file item
-        public static void set_selection_text_from_file_list (Gtk.SelectionData selection_data,
+        // Used when TargetType is STRING
+        public static void set_selection_data_as_text (Gtk.SelectionData selection_data,
                                                               GLib.List<Files.File> file_list,
                                                               string prefix = "") {
```
lenemter, line 317:
```suggestion
        public static void set_selection_data_as_text (Gtk.SelectionData selection_data,
                                                       GLib.List<Files.File> file_list,
                                                       string prefix = "") {
```

libcore/DndHandler.vala
```diff
@@ -294,42 +294,53 @@ namespace Files {
             return sb.str;
         }
 
-        // Used when dragging a file item
-        public static void set_selection_data_from_file_list (Gtk.SelectionData selection_data,
+        // Used when TargetType is TEXT_URI_LIST
+        public static void set_selection_data_as_file_list (Gtk.SelectionData selection_data,
                                                               GLib.List<Files.File> file_list,
                                                               string prefix = "") {
```
lenemter, line 300:
```suggestion
        public static void set_selection_data_as_file_list (Gtk.SelectionData selection_data,
                                                            GLib.List<Files.File> file_list,
                                                            string prefix = "") {
```

libcore/DndHandler.vala
```diff
@@ -294,42 +294,53 @@ namespace Files {
...
+                TargetType.STRING
+            );
             sb.truncate (sb.len - 2);  /* Do not want "\r\n" at end when pasting into text*/
             selection_data.set_text (sb.str, (int)(sb.len));
         }
 
         private static void set_stringbuilder_from_file_list (GLib.StringBuilder sb,
                                                               GLib.List<Files.File> file_list,
-                                                              bool keep_protocol) {
+                                                              uint info) {
```
lenemter, line 331:
```suggestion
                                                              TargetType type) {
```

libcore/FileUtils.vala
```diff
@@ -21,14 +21,15 @@ namespace Files.FileUtils {
...
     public GLib.List<GLib.File> files_from_uris (string uris) {
         var result = new GLib.List<GLib.File> ();
         var uri_list = GLib.Uri.list_extract_uris (uris);
-        string unquoted_uri;
         foreach (unowned string uri in uri_list) {
             try {
-                unquoted_uri = Shell.unquote (uri); // Extracted uri may be quoted
-                result.append (GLib.File.new_for_uri (unquoted_uri));
+                var unquoted_uri = Shell.unquote (uri); // Extracted uri may be quoted
+                var escaped_uri = escape_uri (unquoted_uri); // Uris from clipboard or DNDHandler are unescaped/sanitized
```
lenemter, line 31: This function handles uris, they must be escaped if they get to here. And when checking with prints they appear identical:
image

lenemter (changes requested):

libcore/DndHandler.vala
jeremypw: Can we be sure that `info` is a valid TargetType?

libcore/FileUtils.vala
lenemter: Yes, URIs **must** escape special characters as specified by URI syntax -- https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax. If it's not escaped the URI is invalid. I don't think we should try to fix issues in other applications that use URI incorrectly.

jeremypw:
@lenemter I've incorporated your suggested changes and it works correctly with [Terminal](https://github.com/elementary/terminal/pull/894).  I am still getting issue https://github.com/elementary/music/issues/816 with Music though whereas Dolphin and Nautilus work correctly. This is using a file with`#` and spaces in the file name.

May be because the `main` branch is now `Gtk4` I guess.

jeremypw:
@lenemter Its started working with Music now! Not sure what's going on. Must have been running a different version to the one I thought :shrug:

jeremypw:
@lenemter I've made the requested change - sorry I missed that.  It seems to work OK but I would be grateful for more testing.

jeremypw:
Hmm, dropping onto the pathbar seems broken.

Update: This seems to be an issue in `main`

jeremypw:
Also dropping some files (containing hash in name) onto Dolphin is broken but works on Nautilus :shrug:

Update:  Dropping the same file from Nautilus doesn't work either so maybe a Dolphin issue.

lenemter (approved):
LGTM

PR #2624, opened 2025-07-26
Update view after permanent delete of folder containing files

### LINKED ISSUE #2623: File Manager does not refresh state of the current path automatically
### What Happened?

I often delete old directories (with SHIFT + DEL) of old projects of mine, and the directory doesn't always disappear visually; it only disappears after I close and reopen File Manager and enter the same path again. In another words, the File Manager does not refresh state of the path automatically.

That's a critical bug in my opinion, imagine a non-technical user's experience seeing a folder that doesn't exist anymore visually. If the user tries to interact with that deleted folder, obviously it gives erros, but it should not be there visually in the first place.

### Steps to Reproduce

1. Delete a folder with SHIFT + DEL
2. Check if it actually disappears visually on the File Manager, if not close the File Manager and reopen the same path and check again.

### Expected Behavior

After deletion of the directory, the File Manager should refresh automatically and visually remove the deleted directory also, but that is not always the case. Sometimes the deleted directory stays visually. If you try to interact with it, it gives errors. I suspect it does delete the directory underneath the hood, but, visually there is a bug or the File Manager is not refreshing its state automatically.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #2623

### DIFF
--- a/libcore/FileChanges.vala
+++ b/libcore/FileChanges.vala
@@ -23,7 +23,8 @@ namespace Files.FileChanges {
         INITIAL,
         ADDED,
         CHANGED,
-        REMOVED,
+        FILE_REMOVED,
+        FOLDER_REMOVED,
         MOVED
     }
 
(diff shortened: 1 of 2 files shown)


leonardo-lemos (approved):

lenemter:
Permanently deleting multiple of files/directories at once often doesn't refresh the view

https://github.com/user-attachments/assets/b031fa47-3cd0-4b77-9fa7-a0133aac2f2a

jeremypw:
> Permanently deleting multiple of files/directories at once often doesn't refresh the view

I cannot reproduce this with test folders. You imply that the issue is not very reproducible - any idea under what conditions it happens?  I am inclined to wait for an issue report with steps to reproduce before looking into this further.

PR #2638, opened 2025-08-21
Do not unquote dropped and pasted uris from uri_list

### LINKED ISSUE #2637: Issue with folder name containing single quotes
### What Happened?

I encountered an issue while trying to move an image from the "Captures d'écran" folder (print-screen's default folder name in French) to another directory with Files. The operation failed, and I discovered that the problem was related to the folder name containing a single quote.

### Steps to Reproduce

1. Create a folder in Files
2. Place a file in the folder
3. Rename de folder with a ' in the name
4. Try to move the file to another folder

### Expected Behavior

Files move files as expected in folder named with single quotes, like "Captures d'écran".

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #2637

Simplest fix for the linked issue.  

Files not longer quotes uris for dropping into apps like Terminal - the destination must do any quoting themselves. We can assume other apps behave similarly and we do not need to unquote received uris. 

Note that since a single quote is one of the characters in `GLib.RESERVED_CHARS_SUBCOMPONENT_DELIMITERS` it does not get escaped in uris when files are being drag/dropped - but that is not changed by this PR.

Reviewers should confirm that under no circumstances do we need to unquote.

### DIFF
--- a/libcore/FileUtils.vala
+++ b/libcore/FileUtils.vala
@@ -23,13 +23,8 @@ namespace Files.FileUtils {
         var result = new GLib.List<GLib.File> ();
         var uri_list = GLib.Uri.list_extract_uris (escaped_uris);
         foreach (unowned string uri in uri_list) {
-            try {
-                var unquoted_uri = Shell.unquote (uri); // Extracted uri may be quoted
-                // We require that this function be called with escaped uris
-                result.append (GLib.File.new_for_uri (unquoted_uri));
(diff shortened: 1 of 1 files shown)


zeebok (approved):
Seems to work as intended

PR #2644, opened 2025-09-12
Fix freeze when user template dir points at very large directory tree.

### LINKED ISSUE #2643: Right-clicking on the white space in the home folder freezes Files
### What Happened?

I wanted to view the properties for my home folder to see disk usage. However, when I right click on the white space between icons in the folder, Files becomes unresponsive and I have to kill it from the command line.

This doesn't happen when I right-click on a folder or file, only on the blank (white) space in the window.

### Steps to Reproduce

1. Open Files from Dock
2. Right-click on white space between icons (in home folder, but also any other folder)
3. No menu appears, Files becomes unresponsive

### Expected Behavior

A menu should appear on right-click, and the app shouldn't freeze.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #2643 

The MAX_TEMPLATES constant was not being properly honored resulting in a huge menu structure being made under the circumstances of the linked issue.  This is fixed and some other improvements made:

 * Honor MAX_TEMPLATES properly
 * Indicate when too many
 * Do not show empty directories
 * Insert root "Template" item when there are templates

I am uncertain about the "…too many templates" item inserted when MAX_TEMPLATES is reached and this can be removed if desired but in practice it is unlikely to appear I should think as up to 2048 templates are allowed.  When the template folder contains this number or more the context menu appears with a slight but tolerable delay on modern hardware.


ernstki:
Can confirm this works around #2643 as detailed in [this comment](https://github.com/elementary/files/issues/2643#issuecomment-3364315298).

jeremypw:
Thanks @ernstki - hopefully another dev will approve this for merge in due course.

src/View/AbstractDirectoryView.vala
```diff
@@ -2358,6 +2358,7 @@ namespace Files {
         }
 
         private class NewSubMenuItem : Gtk.MenuItem {
+            private uint total_item_count = 0;
             construct {
```
lenemter, line 2362:
```suggestion
            private uint total_item_count = 0;

            construct {
```

PR #2648, opened 2025-09-28
Fix expand row of ListView when in Trash

### LINKED ISSUE #241: Entering a folder in trash loses the trash:// scheme
Further to comments on PR#231, all uris inside a trash folder should start with the trash:// scheme.  This is not always the case e.g. when a subfolder is entered.  It is unclear at the moment whether this is an issue with Files or gvfs.

### LINKED ISSUE #2647: Directory "unrolling" broken in Trash folder
### What Happened?

For example, before unrolling I have this:
Image

After unrolling "gog-galaxy" I have this:
Image

As you can see, the subfolders have the same alignment as their parent and the folders that alphabetically before it are displayed above it, while all files are display below other folders.

### Steps to Reproduce

Open the "trash:///" path and unroll a folder.
This does not happen when we open the "$HOME"/.local/share/Trash/files/ directory.

I tested only in the X session, with the latest updates to elementary 8

### Expected Behavior

Sub-folders and files should be displayed below their parent but above other folders, and have their alignment offset to the right

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #2647 
Fixes #241

It is intentionally only possible to perform trash operations on the top level when within trash but there is still a need to look within folders to what will be operated on so if we allow ListView it must operate correctly in trash.

This PR makes sure that we stay within the `trash//:` hierarchy when navigating so that the ListModel works correctly.


jeremypw:
@lenemter Thanks for the review!

PR #2654, opened 2025-10-30
Jeremypw/fix afc mount

### LINKED ISSUE #1543: When connecting iPhone, only its "Documents" directory is mounted
# Description

When I conntect an iPhone I see this in the sidebar of Files:

image

When I click on "Documents on iPhone" I see this (after waiting for Files to unfreeze, as mentioned in #1541):

image

The contents above is just the "Documents". I can't easily navigate to the DCIM directory to see pictures, access Downloads, etc. In order to access the "root" directory of iPhone I have to manually edit the path at the top:

 
image

I had to remove `:3` from the end of the path. This way I finally get access to the "root" directory:

image

# Why it's wrong

I believe that access to the "root" directory should be mounted by default, instead of the "Document" directory. It is not easy to figure out how the path should be modified to get to what you need.

# Version 
eOS 5.1.7

jeremypw:
Fixes #1543 

Needs a reviewer with ability to mount iPhone for file transfer.  Which locations are mountable/accessible?


jeremypw:
Converting to draft pending some clarification from the OP.

danirabbit:
Totally happy to test this for you. But also, for some added context based on Discord discussion, I think it would be totally fine if you didn't feel like maintaining support for this. It seems like iPhone users are far less likely to want to plug their phone in to their computer with a cable, even among my very heavily skewed technical audience:

iPhone users are multiple times less likely to plug in their phone with a cable than Android users:

https://mastodon.online/@danirabbit/117134290474942538

And overall less likely that Android users to transfer files manually at all (even wirelessly):

https://mastodon.online/@danirabbit/117134315037973351

danirabbit:
With this branch:

The files mount has a nice AFC root with a phone icon:
Screenshot from 2026-08-21 09 36 07

We get the photos mount:
Screenshot from 2026-08-21 09 36 16


afaict the photos mount is empty? But also it seems like this is in the scope of Photos not Files? (fwiw, in Photos its empty too)
Screenshot from 2026-08-21 09 36 28

flodavid:
The same folder and pictures than on Windows are shown in the "iPhone" mount point on my side.
The "Documents on iPhone XXX" mount point now contains the data folder of applications that enabled the feature to import/export files but it is very slow to navigate and the application constantly hangs.
With the current version of Files, I have different folders. I can access pictures uploaded to the phone, downloaded podcasts, and the DCIM contains (some ?) pictures too..., but it is a bit of a mess. Navigation is very fast though and the application does not hang.
~What is strange is that in both versions, the phone does not ask for permission when I mount either folder, which it does on Windows.~ Nevermind, I did ask for permission for my previous iPhone, but it does not anymore.

I have an iPhone 15 (USB 2.0) with iOS 18.7.8.

Current : 
Screenshot from 2026-08-21 19 09 38
This PR : 
Screenshot from 2026-08-21 19 12 52

flodavid:
Here is the view of the "Files" menu inside the "Apple Devices" app on Windows, corresponding to what's now shown with this PR in the documents mount point: 
image

jeremypw:
@flodavid Thanks for testing. As far as I can tell we need to create *two* mounts from the single mount provided by `gvfs` which requires significant "special case" treatment of iPhone AFC mounts.   Ideally this needs to be be fixed in `gvfs` so that when an iPhone is plugged in Files gets notified of the correct mounts just like other protocols.  From what you say, I am not sure whether this PR is an improvement tbh but if it is considered progress and not cause regressions then it could be merged.

It really needs an iPhone owner and dev with an interest in this issue to make further progress I am afraid. As @danirabbit  mentioned it seems to be a niche issue.

It should be possible to transfer files by other means such as bluetooth connection (given appropriate permissions).  A filemanager on your phone can be used to browse files on the phone.

jeremypw:
If we cannot get usable AFC mounts then it may be better to ignore them altogether.

jeremypw:
You could try mounting the iPhone using `ifuse`.  Instructions for doing this can be found e,g, at https://www.hukot.net/community/en/tutorials/mount-iphone-in-debian-ubuntu-linux (use at your own risk)

After mounting you should be able to browse the ~/iPhone folder like any other.  You will need to install ifuse using apt.

PR #2656, opened 2025-10-30
Fix destination icon in file conflict dialog

### LINKED ISSUE #695: No icon of original file in replace dialog
You can see this at the picture

jeremypw:
Fixes #695 (finally)

 - Gets destination image the same way as source icon; in file ready callback.
 - Improves code style by moving code out or create function to construct clause.


libcore/FileConflictDialog.vala
```diff
@@ -36,54 +36,36 @@ public class Files.FileConflictDialog : Granite.MessageDialog {
         NEWEST
     }
 
+    public Files.File source { get; construct; }
+    public Files.File destination { get; construct; }
+    public Files.File dest_dir { get; construct; }
```
lenemter: These should be `private get`, no?

jeremypw: I thought construct properties had to be public get but maybe I'm wrong - I'll check.  There's probably a lot of other instances with this pattern in my code!

PR #2662, opened 2025-11-02
Take presence of hidden files into account when showing empty message

### LINKED ISSUE #1278: Ignore hidden files for 'this folder is empty' message
## Prerequisites

## Describe the bug
The empty folder messages does not display if there are files or folders in a directory, even if these are hidden system files. I found this discrepancy when formatting a drive for additional storage and noticed that the NTFS partition displayed a message that it was empty but the EXT partition did not, only a white background.

EXT partitions by default have a hidden folder called '[lost+found](https://unix.stackexchange.com/a/18157)' for proper operation of fsck. Most users would/should not care about this folder existing. It would be better to report the folder as being empty when only this folder is present.

### To Reproduce
1. Open [gnome-disk-utility](https://packages.ubuntu.com/bionic-updates/gnome-disk-utility)
2. Create two new Disk images
3. Format both images, One with NTFS and one with EXT4
4. Mount both images in files using the sidebar
5. Observe difference, the NTFS disk correctly shows the message 'This folder is empty', the EXT formatted one does not and is just empty.

### Expected behaviour

A folder with only hidden files (such as lost+found) should display the 'Empty folder' message.

When the user explicitly has 'Show Hidden Files' enabled the 'Empty folder' message should go away and show the hidden files and system folders.

#### Caveat
When deleting a folder with (only) hidden files a dialogue _may_ be needed as the user may not be aware that a folder contains hidden files due this message.

  

I've left the descriptions somewhat general as I immediately see two ways to solve this but am unsure which fits better with the elementary philosophy.
1. Explicitly exclude a folder named 'lost+found' from the files count  
2. Exclude all hidden files when the 'Show Hidden Files' is unchecked from the files count  

Both would fix this, I feel nr2 is more inline with elementarys goals but also has the risk of causing confusion. For example when deleting a folder.
 
## Platform Information

elementary OS 5.1.3 Hera  
io.elementary.files 4.4.2 ( [4.4.2+r3786+pkg92~ubuntu5.1.2.1](https://launchpad.net/~elementary-os/+archive/ubuntu/stable/+build/19119459) )

jeremypw:
Fixes #1278 

The chosen solution is to add another label to the existing `empty_label` which shows when there are no visible files but the folder contains undisplayed hidden files.

The original message now displays when the `model` is empty (i.e. the view would be blank). If the model is empty but the folder is not empty then hidden files are present and an addition label is shown with this information.

To reduce the performance hit when pasting or deleting a large number of files (which is already slow) the update function is throttled.  Ideally we would want to implement batch file operations but that will have to wait for Gtk4

### DIFF
--- a/libcore/ListModel.vala
+++ b/libcore/ListModel.vala
@@ -60,11 +60,6 @@ public class Files.ListModel : Gtk.TreeStore, Gtk.TreeModel {
     }
 
     public bool show_hidden_files { get; set; default = false; }
-    public bool is_empty {
-        get {
-            return file_treerow_map.size == 0;
-        }
-    }
 
     private enum PrivColumnID {
         DUMMY = ColumnID.NUM_COLUMNS
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -265,7 +265,6 @@ namespace Files {
         private Gtk.Widget view;
         protected Gtk.ScrolledWindow scrolled_window;
         private Gtk.Label empty_label;
-        private Gtk.Label hidden_label;
         private Gtk.Overlay overlay;
         private unowned ClipboardManager clipboard;
         protected Files.ListModel model;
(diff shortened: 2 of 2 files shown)


jeremypw:
For some reason the diff is not showing all changes - only last commit??

PR #2663, opened 2025-11-02
Indicate presence of undisplayed hidden files with empty message

### LINKED ISSUE #1278: Ignore hidden files for 'this folder is empty' message

jeremypw:
Fixes #1278 

BEFORE

Screenshot from 2025-11-03 09 28 38

AFTER
 - Completely empty folder

Screenshot from 2025-11-08 19 21 45

- Only hidden files

Screenshot from 2026-02-09 17 56 12

- Hidden files temporarily revealed

Screenshot from 2026-05-11 15 13 19


teamcons:
if we mention hidden files we may as well straight up display a button to show them, instead of "lol go find it yourself" ?

jeremypw:
@danirabbit What is your opinion of adding a "Show Hidden" button in this context?

jeremypw:
OK, so should it change the global setting or just reveal the hidden files for the folder in question?

teamcons:
A one-time "display this folder" that lasts only till you change folder could be nice

jeremypw:
That would be possible. My only concern is that there would be no obvious way to revert the step without reloading or navigating away from the folder (the appmenu and context menu would still show hidden files are not shown). Its not a big deal though I guess.

jeremypw:
@teamcons Latest commit implements a basic implementation of you idea.  Instead of having two labels, there are now only one showing at any time - corresponding to the two relevant situations.  I have also avoided the words "Empty Folder" as it is rather ambiguous.

teamcons:
"This folder contains only hidden files" ?
"hidden files are present" sound slightly ominous
That is a nitpick tho. Pretty cool and thank you for adding this. It feels more "aware"

jeremypw:
I wonder whether the button should show below the text? Any styling required?

zeebok:
> I wonder whether the button should show below the text? Any styling required?

Using only mental images I feel like below might look cleaner

jeremypw:
@zeenbok,  Yes, I guess action buttons more usually show below message text.  I'll amend.

jeremypw:
@zeebok Moved button as per your suggestion. Unsure of the best alignment but I have use `CENTER` as the moment  analagous to the Welcome screen

zeebok:
I think this is good to go if you fix the conflicts

jeremypw:
Need to make compatible with recent changes to settings and actions.

jeremypw:
@zeebok Could you take another look when you have time? I have updated this for recent changes to `main` and also added a refinement allowing the hidden files temporarily shown using the override for that folder to be hidden again. This requires the overlay to continue to be shown but I do not think this is likely to be a problem if the user is only taking a peek at the files.

zeebok:
Yeah I can take a look tomorrow!

zeebok:
It all functions as you describe! Code looks good too. I do notice a strange delay of sorts with this new overlay. If I am in a folder with this message and go up to the parent, I see the folder change and the message overlay is still there for a moment. Also see this delay when pressing the Show/Hide button; slight moment after clicking for the button to update.

zeebok (approved):

jeremypw:
Thanks for the review.  I'll see if I can reduce the update delay before merging.

PR #2665, opened 2025-11-03
Only allow setting default app for regular files, not folders

### LINKED ISSUE #1463: Should not show "Open in | Other application ...." option for folders

## Describe the bug
It is possible to set "Terminal" as the default application for folders through the Files interface.  This has adverse effects including the context menu showing "Open in Terminal" as the first option with the accelerator "Enter".  However pressing "Enter" does *not* open the folder in terminal but navigates as usual.

### Expected behavior
Do not allow changing the default application for folders.  (This can still be done through settings if someone wants to use a different filebrowser by default).

## Platform Information
Elementary Odin in VM

Please check what applies:
 I'm using the latest version in Odin

jeremypw:
Fixes #1463 

There does not seem to be a good reason to set a default app for anything other than regular files, so in addition to folders, the "Set as Default" checkbox is not displayed for links, mounts etc.

### DIFF
--- a/libcore/Resources.vala
+++ b/libcore/Resources.vala
@@ -20,7 +20,6 @@ namespace Files {
     public const string APP_ID = "io.elementary.files";
     public const string APP_DESKTOP = APP_ID + ".desktop";
     public const string APP_TITLE = N_("Files");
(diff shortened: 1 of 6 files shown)


jeremypw:
In general, the context menu does not include "Open in …" for inappropriate types such as network items or mounts so it is not necessary to worry about those there

PR #2666, opened 2025-11-03
Pad the formatted datetime

### LINKED ISSUE #1538: Listview Vertical scrollbar is on top of modified date column information.
On the list view vertical scrollbar is on top of modified date column information. Is it possible to make more space on the right side of the modified column?

vertical_scrollbar

jeremypw:
Fixes #1538 

BEFORE
Screenshot from 2025-11-03 11 02 47

AFTER (ISO format)
Screenshot from 2025-11-07 12 36 04
.

### DIFF
--- a/libcore/FileUtils.vala
+++ b/libcore/FileUtils.vala
@@ -700,7 +700,7 @@ namespace Files.FileUtils {
                 break;
         }
 
-        return string.join (" ", dt.format (format_string), " ");
+        return dt.format (format_string);
     }
 
     private bool can_browse_scheme (string scheme) {


tintou:
That's not good because the headers also have a margin

jeremypw:
> That's not good because the headers also have a margin

Ah, that's a pity - it was the simplest solution :disappointed:  I guess I will have to tweak the render instead.

jeremypw:
I know that using spaces is not elegant but bearing in mind it will all have to be re-visited on porting to Gtk4 it does not seem worth spending a lot of time on it.

lenemter (changes requested):
When using ISO format the date is almost touching the scrollbar which looks weird.

image

jeremypw:
Hmm, that's odd - all formats should have one space padding but I can increase it to 2 spaces if you think that looks better. Or maybe there is a fixed-width space char I can use.

jeremypw:
Now using single unichar `emspace` for padding.  With this there is no overlap even when hovered.

libcore/FileUtils.vala
```diff
@@ -689,7 +700,7 @@ namespace Files.FileUtils {
                 break;
         }
 
-        return dt.format (format_string);
+        return string.join (" ", dt.format (format_string), " ");
```
lenemter: `get_formatted_date_time` already adds emspaces so I think we don't need to add spaces here?

lenemter (approved):

PR #2667, opened 2025-11-03
Show search results for current folder first

### LINKED ISSUE #2015: Search results should list folders and files "In this folder" on top
### What Happened?

In search results, files and folders "Below this folder" are shown before results "in this folder", which are likely more relevant. This, combined with the lack of rollover forces the user to press "Down" multiple times to get to the more relevant results. Even the "More Results..." button is more easily accessible with a keyboard. 

### Steps to Reproduce

1. Open files
2. Search for a folder existing in the current folder

image

### Expected Behavior

Results in the root folder should appear first. Something like this:

image

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #2015 

BEFORE
Screenshot from 2025-11-03 15 57 04

AFTER
Screenshot from 2025-11-03 15 56 14

### DIFF
--- a/src/View/Widgets/SearchResults.vala
+++ b/src/View/Widgets/SearchResults.vala
@@ -17,14 +17,14 @@ namespace Files.View.Chrome {
             DEEP_BEGINS,
             DEEP_CONTAINS,
             DEEP_ELLIPSIS,
-            ZEITGEIST_HEADER,
-            ZEITGEIST_BEGINS,
-            ZEITGEIST_CONTAINS,
-            ZEITGEIST_ELLIPSIS,
             BOOKMARK_HEADER,
             BOOKMARK_BEGINS,
             BOOKMARK_CONTAINS,
-            BOOKMARK_ELLIPSIS;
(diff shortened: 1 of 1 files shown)


jeremypw:
@ryonakano Thanks for the review! This works by ensuring that the "In this Folder" header sorts before the other headers.  Without the suffix, the sort key would be the same for all of them.  To be honest, I cannot remember why I used the suffix "4" for the Zeitgeist header. :shrug: Maybe just a typo.  I'll check it still works with "3" and change it if it does,

jeremypw:
Hold on a moment reviewing, I want to check something.

jeremypw:
I can't see any reason now why I did not use the existing header categories, for deep, bookmark and zeitgeist searches which are already in the desired order (apart from zeitgeist). Hopefully there isn't some subtle reason I am missing...

ryonakano (approved):
Thanks for explanation! The current code more makes sense to me and confirmed the linked issue is fixed.

PR #2668, opened 2025-11-03
Intercept  and ignore all vertical scroll events when renaming or frozen

### LINKED ISSUE #1194: Renaming field is misplaced when scrolling
# Prerequisites

# Describe the error

The renaming field of files and folders is misplaced when scrolling.

## To reproduce

1. Open Files.
2. Use the List or Column view.
3. Start renaming the file or folder.
4. Scroll through the list of files and folders.
5. The renaming field follows along, causing a strange appearance.

- This error only happens in the List and Column views.

## Expected behaviour

The renaming field stays where it should in the view of files and folders. Alternatively, the renaming operation is interrupted when one starts scrolling.

## Screenshots or screen recordings

[Screen recording](https://mega.nz/#!ZmZhzaRR!BNnMT7gvUKhiVterqbP_9dWBLtLVhiyWGF4-Zv0KPrU)

# Platform information

[System information](https://mega.nz/#!M752FaDR!8zo2QmFPde_AQNbelPmnM4nNekaQuBeLNciudOjG_zQ)

Language: French
Keyboard: English (US) - English (Dvorak intl., with dead keys)
Location: Paris

The testing system is installed on a small partition on a ThinkPad T470 laptop on which UEFI Secure Boot is disabled. Aside from specifying the language, keyboard, location, initial user, and the specific partition in which to install elementary OS, all of the default settings were used during the installation. No Internet connection was used during the installation. The only system-wide changes which have been made after the installation are that the system connects to the Internet via a wireless access point and that [Screen Recorder](https://appcenter.elementary.io/com.github.mohelm97.screenrecorder/) has been installed via AppCentre.

# Additional context

Before this test was performed, the following steps were followed:

1. All system components were updated via the blue “Update All” button in AppCentre.
2. A new administrator user named “Essai” or “Essai 2” was created with the password “komalleana”.
3. The current user was logged out and the new user was logged in.
4. The old user was deleted, the screen resolution was set to 800x600 at 60 Hz, tap clicking was disabled, and the touchpad pointer speed was set to the maximal value.
5. The computer was rebooted.
6. The test user was logged in.

jeremypw:
Fixes #1194

Scrolling is inhibited by setting the flags for the scroll controller to intercept all scroll events when renaming starts - the handler then ignores the events.

The keypress controller is prevented from modifying the scroll controller flags during a rename.

It is not necessary to intercept horizontal scroll events in Miller view as the renaming entry is tied to the slot and moves horizontally with it.

### DIFF
--- a/libcore/Directory.vala
+++ b/libcore/Directory.vala
@@ -844,17 +844,6 @@ public class Files.Directory : Object {
         }
     }
 
-    public void update_desktop_files () {
-        foreach (unowned Files.File gof in file_hash.get_values ()) {
-            if (gof != null && gof.info != null &&
-                (!gof.is_hidden || Preferences.get_default ().show_hidden_files) &&
-                gof.is_desktop) {
-
-                gof.update_desktop_file ();
-            }
-        }
-    }
-
     public Files.File? file_hash_lookup_location (GLib.File? location) {
         if (location != null && location is GLib.File) {
             Files.File? result = file_hash.lookup (location);
--- a/libcore/File.vala
+++ b/libcore/File.vala
@@ -459,7 +459,7 @@ public class Files.File : GLib.Object {
         return FileUtils.get_formatted_time_attribute_from_info (info, attr);
     }
 
-    //TODO Is it necessary to refetch the icon if have pix at requested size? 
+    //TODO Is it necessary to refetch the icon if have pix at requested size?
     public Gdk.Pixbuf? get_icon_pixbuf (int _size, int scale, IconFlags flags = IconFlags.USE_THUMBNAILS) {
         return get_icon (
             _size.clamp (16, 512),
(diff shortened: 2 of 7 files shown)


lenemter:
I don't think that's the best solution. The renaming field should scroll along with the view and when typing it should scroll back to that item. Not sure if that's possible to implement in Gtk3

jeremypw:
@lenemter I don't think that is possible using the builtin editing function of Gtk.TreeView - possibly not in Gtk4 either while we continue to use Gtk.TreeView editable.  Once we switch to dynamic views it will no longer be an issue but that is some way off at the present rate of progress. I'm not keen on spending a lot of time on this issue or changing the code too much (e.g. by using a dialog) at this stage tbh.

Is there a valid reason to want to scroll the view in the middle of a rename?  This solution is essentially the same as using a modal dialog, preventing interaction with the parent (as far as scrolling is concerned).

PR #2679, opened 2025-12-07
Jeremypw/filechooser portal

jeremypw:
Experimental full function filechooser portal based on Files widgets rather than Gtk.FileChooserDialog. intended to scope out any problems.

Involves creating more stripped down versions of the main Files widgets in `libcore`.  Ultimately some DRYing can be achieved by basing the main widgets on them but during development they are kept separate hence the very large diff.
Also some changes can be split out and applied independently of this PR.

Reinstituted and extended the previous Portal tester.

Pushing for exposure, but may not merge until after Gtk4 port.

 - [x] Entirely replace Gtk.FileChooserDialog with Files.FileChooserDialog using libcore widgets
 - [x] Show filters, choices, read-only check
 - [x] Show text entry in SAVE mode 
 - [x] Allow Copy, Rename (single), Bookmark (with shortcuts or via context menu)
 - [x] Allow changing sort order via context menu)
 - [x] Allow ListView <-> IconView switching 
 - [x] Persist filechooser sidebar-width, window-size
 - [x] Persist filechooser location settings for each filechooser mode (open or save)
 - [x] Persist filechooser zoom level for each viewmode
 - [x] Honor app settings singleclick-select, show-hiddenfiles, show-remote-thumbnails, show-local-thumbnails, date-format
 - [x] Honor system settings: clock-format, sort-directories-first (readonly)
 - [x] Rubberband selection list view
 - [x] Scroll zooming
 - [x] Persist view mode
 - [x] Keyboard zooming
 - [x] Keyboard view mode selection
 - [x] Implement creating and renaming new folder
 - [x] Persist sidebar expander settings
 - [ ] Search
 - [ ] Preview?
 - [ ] Honor system setting: remember-recent-files?
 - [ ] Handle Recent location
 - [ ] Cleanup, DRY, and split up

Some things will not be implemented atm: AppMenu, ConnectionManager, Network view, DND, MillerColumn view

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ build
 *~
 .buildconfig
 *.flatpak-builder*
+*.flatpak*


jeremypw:
A major limitation is that at the moment we implement the freedesktop filechooser interface fully but `Gtk.FileChooserNative" does not (at least in Gtk3).  The Gtk widget does not handle retrieving choices when using the portal. Although it does send the choices through the portal properly it ignores any choice returned.

jeremypw:
Mostly working and CI passing but at the cost of losing some styling and actions. 
Screenshot from 2025-12-11 17 00 59

jeremypw:
So if the libcore widget `Files.FileChooserDialog` is used within the Files app itself then the sidebar styling is correct, however if used from another app either through the portal or by depending on libpantheon-files and using the widget explicitly the styling is lost and this warning is given:

`(io.elementary.files.portal-tester:46320): Gtk-WARNING **: 17:29:01.111: Theme parsing error: <broken file>:1:0: Failed to import: The resource at "io/elementary/files/SidebarExpander.css" does not exist`

Anyone know how to make the gresources compiles by the app from `/data/files.gresource.xml` available both in the app and in the library? I tried somethings but without success so far.

If I can make the app sidebar derive from the basic sidebar used by the filechooser then presumably the styling only need to be available in the library.

danirabbit:
Instead of writing your own portal tester, ASPHD has a comprehensive file chooser portal test page https://flathub.org/en/apps/com.belmoussaoui.ashpd.demo

jeremypw:
@danirabbit Thanks, I'll try it. The problem I have at the moment is debugging what is going wrong when the filechooser is running as a portal service - you don't get terminal error messages and can't use `gdb` afaik.  I run my tester as a native app, not using the portal, to make sure that the widgets themselves are working as expected - as well as in a flatpak.  That's how I identified the `gresource` problem.  The tester doesn't tell you how to fix it though :disappointed:

jeremypw:
Looking at the Granite library (v6.2.0) it doesn't use gresource for its custom styling so perhaps its not possible for libraries in Gtk3?  I see `granite-7` does complicated stuff with gresources though.

jeremypw:
Following @danirabbit's PR moving gresource compilation to the library the portal now has the custom sidebarstyling!
I thought I tried something similar but must have messed it up.

Screenshot from 2025-12-11 19 27 24

jeremypw:
Next job is to get some app-settings honored in a portal context.  Presumably its not a problem if the app and portal share the same preferences?

jeremypw:
Filechooser with Icon view:

Screenshot from 2025-12-16 16 26 31

jeremypw:
Closed in favor of #2711

jeremypw:
Closing in favor of #2711

PR #2682, opened 2025-12-11
LibCore: compile gresource

danirabbit:
builds resources into the library instead of just the files app. Should make resources available to the file chooser portal


jeremypw (approved):
Doesn't break anything in main so lets try it!

PR #2688, opened 2025-12-29
Release version 7.3.0

jeremypw:
Target date April 24, 2026

Todo:
- [x] https://github.com/elementary/files/issues/2714
- [x] https://github.com/elementary/files/pull/2713


danirabbit:
I think the column preview feature needs some more fixing before we should release it

danirabbit:
@jeremypw I'm really suprised you merged it tbh. The code style has issues. There's todo comments. It feels like it wasn't really tested :(

jeremypw:
@danirabbit Are there any other changes to the column preview feature you want before release?

danirabbit:
@jeremypw off the top of my head the scrolledwindow being packed where it is is really janky imo. There's no way to tell where it is and it doesn't run against the edge of the view so it makes it hard to scroll.

I'll make sure to go through tomorrow. I know you've been waiting on release. Sorry 🙏🏼

jeremypw:
Ah yes, there is a slight gap - I hadn't noticed that as I always scroll with a two-finger gesture. Let me know if you would like me to fix it.

danirabbit:
@jeremypw I proposed https://github.com/elementary/files/pull/2713. I'd also like to see https://github.com/elementary/files/issues/2696 fixed

jeremypw:
@danirabbit #2703 needs some more work - I'll get that done as soon as I can.

danirabbit:
@jeremypw to unblock release we could always put the preview pane behind a default off compile time option as well until it's fixed

jeremypw:
@danirabbit May have to do that if you want flawless behaviour - there are some tricky issues and UI decisions. I assume there is no way to target a release at a particular commit (other than HEAD)?

danirabbit:
@jeremypw We can pick a certain commit to merge back to the `horus` branch, but then if you add commits there for metainfo you'll have to deal with diverging histories which is a pain. It would probably be easier to add the compile option

danirabbit (approved):

PR #2694, opened 2026-01-04
Prepare for Files.FileChooserWidget

jeremypw:
Continue to move required classes into library.

 - Browser (with ZeitgeistManager)
 - Cell renderers (with ClipboardManager & editable labels)

Update POTFILES accordingly  (keeping existing files even where they do not appear to contain translatable strings)


danirabbit:
Don't forget to update POTFILES. It looks like the last branch broke the gettext action also

PR #2698, opened 2026-01-05
DetailsColumn: rename button to Properties and align

danirabbit:
* Remove unused box
* Code style
* End align instead of start
* Rename as "Properties" to match the context menu
* Don't make it flat so you can tell it isn't another line of text


jeremypw (approved):
Makes sense

PR #2704, opened 2026-01-06
Do not show overlaybar on top of file detail column

### LINKED ISSUE #2697: Dont' show overlaybar and preview column
### What Happened?

The overlaybar contains duplicate information and covers the properties button

### Steps to Reproduce

1. Use column view
2. Have something previewed
3. Notice that the overlaybar contains the same information as the details column and it can also be covering the button to show properties

### Expected Behavior

Don't show the overlaybar when the details column is also showing

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

Compiled from git

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #2697 

The overlaybar hides itself if there is only one file selected and it is not a folder when in ColumnView and show file details option is on.

### DIFF
--- a/src/View/Widgets/OverlayBar.vala
+++ b/src/View/Widgets/OverlayBar.vala
@@ -90,7 +90,6 @@ namespace Files.View {
             deep_count_cancel ();
 
             if (update_timeout_id > 0) {
-                warning ("remove source %u", update_timeout_id);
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
Works as expected :tada:

PR #2708, opened 2026-01-12
Jeremypw/libcore widgets/basic slot

jeremypw:
Make cut-down versions of views  and related classes based on #2679  and put in `libcore`.

TODO

 - [ ] DRY code in src
 - [ ] Cleanup


jeremypw:
Closed in favor or #2710

PR #2710, opened 2026-01-18
Move most view widgets to libcore

jeremypw:
Rather than create base versions of view widgets that are then subclassed in `src` which proved onerous and introduced a lot of new code, this is an attempt to just move most of the view widgets into libcore and make them able to work in both app and filechooser environments.

Hopefully a lot of the required changes can be split out into separate PRs before the actual move is made, making reviewing much easier.

 - Uses interfaces for widget interactions as the `src` and `libcore` widgets may differ in some cases but provide the same interface
 - References to Window and ViewContainer are replaced by reference to SlotToplevelInterface 
 - The toplevel  is longer be assumed to be an Application Window
 - The settings schemas are not fixed since different schemas may be used for the filechooser.  All settings are now mediated by proxy objects instead of only some of them. The schemas may be changed as long as the required keys are present.

The code now compiling and starting to produce expected loaded tabs. CI tests now pass  Need to test all functionality manually.

### DIFF
--- a/po/extra/fr.po
+++ b/po/extra/fr.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: pantheon-files\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2026-01-09 08:35+0000\n"
-"PO-Revision-Date: 2025-11-12 19:05+0000\n"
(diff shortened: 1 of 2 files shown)


jeremypw:
Blocked until all changes that can be split out into simple PRs have been merged, to reduce diff.

PR #2711, opened 2026-01-21
FileChooserPortal: Use existing view widgets after moving to libcore

### LINKED ISSUE #129: File chooser design should better match files.
The file chooser is a module version of the file browser, and the filebrowser represents the primary way that users interact with the file system, and as such should reflect the design of the file browser in terms of views, and sidebar padding and order.

For image focused tasks list view is so bad that I open files because dragging and dropping is faster than the filechooser, this wouldn't be the case with an icon view because in icon view having meaningful names for images doesn't matter at all.

The sidebar design in filechooser should be copied directly from files, I've spent a long time in files and my bookmarks are muscle memory at this point.

screenshot from 2017-09-10 13 39 13

jeremypw:
Replaces #2679, builds on #2710, 

Fixes #129  and (potentially) many others

This PR replaces the existing Gtk.FileChooser with  a widget that implements most of the Files app functionality and uses mostly the same widgets.

To avoid problems previously encounted with Gtk.FileFilterInfo,  we implement our own simplified  filtering with `Files.FileFilter`. Only mime-type and glob filtering are implemented.

Much of this can be split out into small PRs so it is not expected that this PR will be merged in toto.

Currently partially works but needs further work.  The Miller view awaits transfer to `libcore` so does not work in the portal. Othewise most of the `src` folder has been transferred to `libcore`.

The `tests` subfolder will be lost before merging

Note: Only works if app use the portal (e.g. Flatpaks) .  Native Gtk3 apps must be run with the environment variable `GTK_USE_PORTAL=1`  and use Gtk.FileChooserNative  otherwise the native Gtk.FileChooser will appear.

### DIFF
--- a/filechooser-portal/Main.vala
+++ b/filechooser-portal/Main.vala
@@ -405,7 +405,6 @@ public class Files.FileChooserPortal : Object {
         }
 
         if ("multiple" in options) {
-            warning ("Main setting select multiple");
             filechooser.select_multiple = options["multiple"].get_boolean ();
         }
 
--- a/libcore/FileChooser/FileChooserDialog.vala
+++ b/libcore/FileChooser/FileChooserDialog.vala
@@ -534,7 +534,6 @@ public class Files.FileChooserDialog : Gtk.Dialog, Xdp.Request2 {
     }
 
     public void set_uri (string uri) { // Select file at uri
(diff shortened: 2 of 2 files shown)


jeremypw:
Blocked until #2710 merged

PR #2714, opened 2026-01-29
Introduce side-widget in Abstract Slot and use for preview window

### LINKED ISSUE #2696: Column preview gets cut off
### What Happened?

Column preview can be cut off such that the preview and information is not fully shown

### Steps to Reproduce

Have several columns open such that the preview is the last column

Image

### Expected Behavior

- [ ] The preview should probably be homogeneous in size with the other columns or have some kind of minimum size
- [ ] Text in the column should wrap so that it's not cut off
- [ ] Image previews should shrink to fill the width of the column

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

Compiled from git

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #2696 

Screenshot from 2026-01-29 19 26 19

Instead of inserting preview widget into slot stack in Miller it is added as a side-widget to the main AbstractSlot (will also work for other views).  This minimises impact on the Miller view and makes it easier to fix some issues.

The side-widget is added into a Paned so the space it takes up can be adjusted by the user.

The preview size has been reduced to be more comparable to the directory slots but wide enough not to cause undue wrapping of the property information

The preview size was chosen to minimise the appearance of black bars to the sides (sometimes) when the window height is changed.  I think this may be a window manager issue?

The Miller View now keeps the active slot in view when the space available to it changes either due to window resizing or preview widget space changing.

It would be nice if the width of the active slot were kept above a minimum but that applies to `main` anyway so can be done in a separate PR.


jeremypw:
@danirabbit There's probably some clean-up and refinement to do but I would appreciate some feedback before going any further.

danirabbit:
Yeah this seems like a much better direction. Nice job!

I've got some weird gap here added to the end of the scrollable area:
Screenshot from 2026-01-30 10 45 38

jeremypw:
@danirabbit Thanks for the review! I am encouraged to spend more time on it. That gap has always been there in Miller View as a consequence of the need to have space for the column to expand into (see `calculate_total_width`  function), but before it lay outside the window.  I'll see what I can do to hide it.

jeremypw:
@danirabbit I have got rid of the gap now but you can reveal the expand-into space by reducing the preview width with the paned (or window larger or deselecting) if you do need to make the last slot larger.  Whenever a "scroll to slot" is performed the gap disappears.

I have also made other minor refinements to scroll handling.

I'll dog-food it for a while but if you see anything let me know.

jeremypw:
@danirabbit Further releases of Files are pretty much blocked until the preview facility is acceptable.  I would be  grateful for your input on this PR which, I think, fixes a number of issues with the current situation.

danirabbit:
I feel like we've gone better in a lot of ways and worse in some ways. Can you revert any changes you've made to the design of the contents of the pane? So changes to scrolled areas, adding the separator etc. I'd like to consider those kinds of changes separately

jeremypw:
@danirabbit Not sure exactly what you want but I have reverted all changes made to the DetailsColumn widget since moving to side widget.

jeremypw:
@danirabbit Any chance of having another look at this? It is needed for the next release.

libcore/AbstractSlot.vala
```diff
@@ -46,9 +46,12 @@ public abstract class Files.AbstractSlot : GLib.Object {
 
     public virtual bool is_frozen { get; set; default = true; }
 
+    //TODO Make private so layout fixed?
     protected Gtk.Box extra_location_widgets;
     protected Gtk.Box extra_action_widgets;
-    protected Gtk.Box content_box;
+    protected Gtk.Grid content_box;
+    protected Gtk.Paned side_widget_box;
```
danirabbit, line 53: Can we make sure to name this in a way that indicates what they are. These names imply that they're Gtk.Box. It should instead be `content_grid` and `side_widget_paned` I think

libcore/FileUtils.vala
```diff
@@ -619,32 +620,36 @@ namespace Files.FileUtils {
...
     }
 
     // We add a fixed-width space as both prefix and suffix to ensure date is not obscured by the scrollbar (unless hovered)
     // regardless of the text direction
     // See https://github.com/elementary/files/issues/1538
-    private string emspace_pad (string s) {
+    private string emspace_pad (string s, bool pad) {
+        if (!pad) {
+            return s;
+        }
```
danirabbit, line 651: I don't see where this is ever set false?

src/View/Miller.vala
```diff
@@ -109,7 +120,10 @@ namespace Files.View {
             path_changed ();
             guest.slot_number = (host != null) ? host.slot_number + 1 : 0;
             guest.colpane = new Gtk.Box (Gtk.Orientation.HORIZONTAL, 0);
-            guest.colpane.set_size_request (guest.width, -1);
+            // guest.colpane.set_size_request (guest.width, -1);
```
danirabbit, line 123: Some commented code here

src/View/Miller.vala
```diff
@@ -109,7 +120,10 @@ namespace Files.View {
...
+            guest.colpane.set_size_request (
+                2 * guest.width, -1
+            );
```
danirabbit, line 126: Can we use just `width_request` since the height request is -1?

src/View/Miller.vala
```diff
@@ -188,10 +202,15 @@ namespace Files.View {
...
-            total_width = 100; // Extra space to allow increasing the size of columns by dragging the edge
+             // Extra space to allow increasing the size of columns by dragging the edge
+             // Also without it, the scrolled window page size increases unnecessarily for non-obvious
+             // reasons
+            total_width = END_GAP;
             slot_list.@foreach ((slot) => {
                 total_width += slot.width;
             });
+
+            this.colpane.set_size_request (total_width, -1);
```
danirabbit, line 213: Another place where we can probably use `width_request` instead of `set_size_request`

danirabbit (changes requested):
Thanks for the ping :) Just a few code comments otherwise this looks good. Sorry for taking a while to review!

libcore/AbstractSlot.vala
jeremypw: Better names now used

libcore/FileUtils.vala
jeremypw: Hmm, I clearly intended an option not to pad the datetime but I cannot remember what situation I was thinking of. Anyway it is not used in this PR (now) so I'll remove.

jeremypw: Ahh, I remember now - we need to not pad the datetime in the DetailsColumn.  I will how actually implement that!

src/View/Miller.vala
jeremypw: Now gone

src/View/Miller.vala
jeremypw: Done

src/View/Miller.vala
jeremypw: Done

danirabbit (approved):

PR #2717, opened 2026-02-02
Make GOF.File content_type a property

jeremypw:
The function `get_ftype ()` is called repeatedly from many places but the content type of a file rarely changes so this is inefficient.

A new property `content_type` is introduced and only updates when the backing member `_content_type` is null otherwise the previously determined content type is returned.

If a change occurs to a file that could affect its content-type or icon then so long as the directory is monitored, it should be picked up and trigger a full update of the file's info and a redraw. See line 910 of `Directory.vala`.

These changes produce a small improvement in the loading and reloading of large folders (about 5%).

Reviewing of file updating produced the following related changes which will be the subject of separate PRs:

 - `File:update_desktop_file` is only called by `Directory:update_desktop_files`  which is never used. Both removed.
 - the signal `File:icon-changed` is only listen to internally by File so was removed its handler was changed to a function for clarity.
 - the function `File:update_type ()` was not used and is removed.

### DIFF
--- a/po/extra/sv.po
+++ b/po/extra/sv.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: pantheon-files\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2026-03-28 20:08+0000\n"
-"PO-Revision-Date: 2026-02-24 11:55+0000\n"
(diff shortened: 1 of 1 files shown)


jeremypw:
Need to merge #2720 first, to reduce diff not directly related to content_type

PR #2721, opened 2026-02-05
Introduce SlotToplevelInterface and start to use

jeremypw:
Working towards reworking the view widgets so that they will work in `libcore` in both the app and the filechooser portal.

Although `Window` will remain in `src` it will have to implement an interface that library widgets can use without referring to this specific window.  The FileChooserDialog will also need to implement the same imterface.

This PR implements most of the required interface in `Window` while making minimal changes.  The required methods are already implemented or can be implemented trivially.

Once this PR is merged, then references to `Window` will be replaced by references to the interface in view widgets.

### DIFF
--- a/src/View/Window.vala
+++ b/src/View/Window.vala
@@ -21,7 +21,7 @@
 *              ammonkey <am.monkeyd@gmail.com>
 */
 
-public class Files.View.Window : Hdy.ApplicationWindow {
+public class Files.View.Window : Hdy.ApplicationWindow, SlotToplevelInterface {
     static uint window_id = 0;
 
     const GLib.ActionEntry [] WIN_ENTRIES = {
@@ -102,7 +102,6 @@ public class Files.View.Window : Hdy.ApplicationWindow {
 
     public signal void loading_uri (string location);
     public signal void folder_deleted (GLib.File location);
-    public signal void free_space_change ();
 
     public Window (Files.Application _application) {
         Object (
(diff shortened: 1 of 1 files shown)


libcore/Interfaces/SlotToplevelInterface.vala
```diff
@@ -0,0 +1,42 @@
...
+public interface Files.SlotToplevelInterface : Gtk.Widget {
+    public signal void free_space_change ();
+
+    public abstract unowned AbstractSlot? get_view (); // Should return current slot
+    public abstract AbstractSlot? prepare_reload ();
+    public abstract void refresh (); // Reloads the current slot
+    public abstract bool can_bookmark_uri (string uri);
+    public abstract void change_state_show_hidden (SimpleAction action);
+    public abstract void bookmark_uri (string uri, string custom_name = "");
+    public abstract Gtk.Application? get_files_application ();
```
danirabbit: GLIb.Application already has a `get_default ()`

libcore/Interfaces/SlotToplevelInterface.vala
```diff
@@ -0,0 +1,42 @@
...
+/* Interface to be implemented by toplevel of both app (MainWindow) and
+* filechooser (FilechooserDialog)
+*/
+
+public interface Files.SlotToplevelInterface : Gtk.Widget {
+    public signal void free_space_change ();
+
+    public abstract unowned AbstractSlot? get_view (); // Should return current slot
+    public abstract AbstractSlot? prepare_reload ();
+    public abstract void refresh (); // Reloads the current slot
```
danirabbit: This should probably be an action so that you can just use the action name `win.refresh`?

libcore/Interfaces/SlotToplevelInterface.vala
```diff
@@ -0,0 +1,42 @@
...
+    public abstract bool can_bookmark_uri (string uri);
+    public abstract void change_state_show_hidden (SimpleAction action);
+    public abstract void bookmark_uri (string uri, string custom_name = "");
+    public abstract Gtk.Application? get_files_application ();
+    public abstract Gtk.Window? get_gtk_window ();
```
danirabbit: I'm not sure what this is about? Seems like you already have `get_toplevel` and in Gtk4 `get_root`?

jeremypw:
Looks like I need to back up and make some substantial changes in another direction before seeing if there remains a need for this interface.  I was starting from where we are now tbh to keep diff small.

PR #2722, opened 2026-02-05
Introduce and use ViewPreferences object

jeremypw:
Working towards eliminating use of static settings objects inside the app and using Preferences objects in the library instead.  

This is so that the FileChooser can use the same objects but where required can initialise them with a different schema.

Before the app static settings objects can be removed it will be necessary to add more settings to `Files.Preferences` as well. This will be done in another PR.

To simplify proxying the window dimensions, the `window-size` variant setting is split into two simple dimensions.

The two preferences objects may be merged at some point.

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -474,17 +474,6 @@ namespace Files {
             }
         }
 
-        public void zoom_normal () {
-            var view_settings = get_view_settings ();
-            if (view_settings == null) {
-                zoom_level = ZoomLevel.NORMAL;
-            } else {
-                zoom_level = (ZoomLevel)view_settings.get_enum ("default-zoom-level"); // syncs to settings
-            }
-
-
-        }
-
         private uint set_cursor_timeout_id = 0;
         public void focus_first_for_empty_selection (bool select) {
             if (selected_files == null) {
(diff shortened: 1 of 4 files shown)


jeremypw:
@danirabbit This makes it simpler in the long run to use view widgets both in an app context and in a filechooser context where they use a different setting schema.  The widgets themselves need not be aware of the context - they access the same preferences object.  Note that the widgets cannot access global settings objects created in `src/Applications` after they have been moved to `libcore` so the alternative would involve passing multiple settings objects (or schemas) around which seems no less complicated.

The current situation of partly using Preference objects and partly using multiple global settings objects is confusing - I found it easier to move to preference objects to achieve the long-term aim of Files.FileChooserDialog. Even though it introduces an intermediate object it is more concise to use once initialised.  It also makes it easier to keep code related to settings handling in one place.

:+1: on making more use of settings actions where appropriate but will still have to cope with the same action being linked with different setting schema depending on context.

danirabbit:
I don't understand why you're trying to have global settings objects at all instead of creating a glib settings object in the class that uses it when you need one?

jeremypw:
@danirabbit That's just a preference - it avoids having hard-coded schema paths all over the place. But if you do create them on demand in view widgets then when the widgets are moved to libcore and are used in two different contexts them they need to be context aware and have double the number of different paths (or be passed them on creation).  Its doable but I don't see the advantage.

danirabbit:
I'm not sure why there would be separate settings for the file chooser?

jeremypw:
Because they are performing different functions. At least the preferred view mode and zoom-level is likely to be different depending on whether it is acting as a filechooser or the app.   I guess the two schemas could be merged if its a problem but I note that Gnome/Gtk have separate schemas for the file-chooser and the file-manager at the moment.

jeremypw:
Closing while investigating feasability of using on demand settings objects.

PR #2729, opened 2026-02-10
AppMenu: use settings action for date format

danirabbit:
* It's not recommended to connect to "toggled", we should use actions instead
* Also use `group` property instead of `from_widget` since this will go away in GTK4


jeremypw (changes requested):
If you change the settings directly with e.g. `dconf-editor` then you get an infinite loop either on the first attempt or subsequent one.  Not sure if this is due to these changes (which otherwise look very good) but be nice to fix it anyway.

danirabbit:
Hm yeah. This appears to have something to do with setting the group. It looks like it's not necessary to do in GTK4, but for some reason we don't get the right styles in GTK3. Might have to wait until after the port

PR #2732, opened 2026-02-10
Window: set view mode action enabled

danirabbit:
We're currently setting the view switcher sensitive, but leaving the action enabled. The action handles keyboard shortcuts so we want to make sure to set its enabled state as well


jeremypw (approved):
Nice catch!

PR #2733, opened 2026-02-10
Window: organize and scope properties

danirabbit:
So we know what our API is here
* public signals and properties at the top
* Explicitly scope things (no implicit private)
* make `tab_view` private set
* Scope down widgets that don't need public access


jeremypw (approved):
Is this ordering mentioned in the developer docs? I was never sure what the approved order was.

danirabbit:
@jeremypw i don't think it is documented. More just convention. I'll put together a proposal!

PR #2734, opened 2026-02-10
Window: use settings action for show-hidden

danirabbit:
Use a settings action instead of creating an action and syncing it to gsettings manually

### DIFF
--- a/src/View/Widgets/AppMenu.vala
+++ b/src/View/Widgets/AppMenu.vala
@@ -13,6 +13,16 @@ public class Files.AppMenu : Gtk.Popover {
     private string[] undo_accels;
     private unowned UndoManager undo_manager;
 
+    private static Settings column_view_settings;
+    private static Settings icon_view_settings;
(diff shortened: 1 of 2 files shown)


danirabbit:
Not sure what that extra action in abstract directory view was about? Any insights @jeremypw? I think it would clean up things a lot to use Settings Actions

jeremypw:
> Not sure what that extra action in abstract directory view was about? Any insights @jeremypw? I think it would clean up things a lot to use Settings Actions

Presumably at the time it was thought easier to keep the same pattern as the other context menu linked actions whereas the toplevel one was for dealing with the accelerator.  I certainly agree Settings Actions, created where needed, should be used now.  Not sure when they came in, but they seem to have passed me by up to now :disappointed: .

danirabbit:
I learned about this recently as well!

danirabbit:
Not sure what I've done here, this was working yesterday I thought but now it's not :sweat_smile:

PR #2735, opened 2026-02-11
SidebarWindow: Use class static settings object

jeremypw:
Continuing towards losing Application static settings objects and replacing with either class static objects or creating temporarily on demand.

Used a different name in order to easily distinguish from previous global object.


danirabbit (approved):
🚀

PR #2738, opened 2026-02-11
Window: use settings action for show file preview

danirabbit:
Another simple settings action change


jeremypw (approved):
No regressions found

PR #2742, opened 2026-02-11
AppMenu: Fix regression and another issue

jeremypw:
Changing to temporary settings object introduced a regression as they have signal handlers attached.   The default button label did not update when the zoom change buttons were used.

 - Persist settings as static objects
 
It was also noticed that the zoom buttons were not correct if the menu was re-opened after switch to a different tab with a different viewmode/zoom-level.  This bug was found to be pre-existing but was fixed here anyway.

 - Set up the buttons every time the popup menu is opened.

It was noticed that it is not now necessary for the toplevel to update the view-mode even if the menu is not open and the relevant code was moved inside the AppMenu. This is pretty much the same as #2731

### DIFF
--- a/src/View/Widgets/AppMenu.vala
+++ b/src/View/Widgets/AppMenu.vala
@@ -4,10 +4,6 @@
  */
 
 public class Files.AppMenu : Gtk.Popover {
-    private static Settings icon_view_preferences;
-    private static Settings list_view_preferences;
-    private static Settings column_view_preferences;
-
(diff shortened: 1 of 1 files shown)


jeremypw:
Closing as this now only fixes a very corner issue - the dateformat buttons updating when the setting is changed externally (e.g. with `dconf-editor` or another Files Window) and the AppMenu does not play well with multiple windows anyway.

PR #2743, opened 2026-02-11
Add zoom actions to abstract directory view

danirabbit:
This should theoretically allow us to clean up a bunch of the zoom handling and let abstract directory view do all of the heavy lifting

### DIFF
--- a/src/View/AbstractDirectoryView.vala
+++ b/src/View/AbstractDirectoryView.vala
@@ -338,10 +338,7 @@ namespace Files {
                 draw_when_idle ();
             });
 
-            set_should_thumbnail ();
-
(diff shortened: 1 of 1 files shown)


jeremypw:
@danirabbit There were some typos that prevented the zoom actions working as expected - I fixed.  Now you just need to get the accelerators working as before.

danirabbit:
I fixed accels, but its now very complicated and feels icky. Maybe this isn't the right direction for where these should live :thinking:

jeremypw:
I agree.  Maybe the toplevel window is the best place after all? Or somewhere you do not need different actions for different viewmodes.

PR #2744, opened 2026-02-11
get show hidden files from gsettings

danirabbit:
get settings directly from gsettings instead of an intermediary object


jeremypw (approved):
Even though `ensure_sizer ()` is called repeatedly during loading of files, I did see of significant change in loading time for folders containing 5000 items so presumably the Settings backend caches the setting anyway.

jeremypw:
The CI passed locally, but failed on GitHub because the schema is not installed.

danirabbit:
Hm I'm not sure how to resolve this. I'll revisit with more coffee maybe. At least this exposes a problem with FilesPreferences being in core but the actual gsettings being set elsewhere :sweat_smile:

jeremypw:
@danirabbit  I proposed one way of working around the CI problem but not sure its acceptable. As there are only a couple of places where the app_settings are used in Files I have allowed for the settings to be `null` (not found in the default schemasource).  There may be a better way to search elsewhere under test conditions.  The settings are not actually used in test as they relate to the view appearance.

jeremypw (changes requested):
Needs conflict resolution and fixing the CI issue.

PR #2745, opened 2026-02-12
Preferences: remove show_file_preview

danirabbit:
Can rebase merge if desired. Make sure we're using gsettings as the source of truth


jeremypw:
I don't think this will be needed if we are going to move the file preview out of Column view.  The overlay will no longer interfere with the preview and the whole of this clause will go away.

Issue #2341, opened 2023-10-12
[Regression] Show thumbnails settings no respected


jeremypw:
### What Happened?

Turned the "Show local thumbnails" and "Show remote thumbnails" off, and cleared the thumbnails under ~/.cache/thumbnails.  After navigating to local and remote folders containing images, thumbnails were still loaded.

### Steps to Reproduce

1. Turn off showing thumbnails
2. Clear the thumbnailer cache
3. Navigate to image containing folders
4. Observe thumbnails still appear

### Expected Behavior

Thumbnails should not be generated (or shown) when the relevant setting is off.

### OS Version

7.x (Horus)

### Software Version

Compiled from git

PR #2747, opened 2026-02-12
AbstractDirectoryView: get thumbnail settings from gsettings

danirabbit:


danirabbit:
@jeremypw should be fixed, sorry!

jeremypw:
@danirabbit Could you resolve conflict please?

danirabbit:
@jeremypw done!

jeremypw (commented):
This mostly works now although there is a strange issue with remote thumbnails on an FTP server (maybe pre-existing - I'll check).  When you navigate to a folder containing thumbnailable files immediately after logging on to the server, then whether thumbnails are shown depends on the **local** setting not the remote setting.  However, if either the settings are then toggled on and off then the correct view is obtained.

jeremypw:
Just confirmed that bug is present in `main` so will approve this.

jeremypw (approved):

jeremypw:
Looks like a similar issue was reported some time ago - #2341

PR #2748, opened 2026-02-12
AbstractDirectoryView: actions use BookmarkList directly

danirabbit:
Instead of pushing things through several layers of UI, actions interact with BookmarkList directly


danirabbit:
@jeremypw should be fixed. I tried to handle checking if a bookmark already exists in `update_menu_actions ()` but it didn't work for some reason. Probably worth cleaning up later, but for now I'll just leave that as it was

PR #2754, opened 2026-02-12
Control: policykit-1 → pkexec

danirabbit:
Goes with https://github.com/elementary/files/pull/2753

fixes Resolute build


danirabbit:
comments in Code's packaging make me think maybe we actually want `libpolkit-gobject-1-dev`? https://github.com/elementary/code/blob/deb-packaging/debian/control

Any thoughts @tintou @ryonakano ?

tintou:
I feel that we should just remove this and use the admin:// scheme instead of pkexec…

danirabbit:
@tintou I agree, I was trying to fix the build in the short term 😅

PR #2760, opened 2026-02-16
AbstractDirectoryView: use GLib.Menu for open in submenu

danirabbit:
No Gtk.Menu/MenuItem in GTK4. Port this submenu to using GLib.Menu


jeremypw:
@danirabbit CI has started failing in the last few days - something to do with policykit.  Any idea why?  I notice you have proposed some related changes but I do not know enough to approve them.

danirabbit:
@jeremypw `development-target` was updated to point to `resolute` containers. I've marked it as not required for now. I do have that branch for policykit that uses the newer package name instead of the old one that was removed. But it's also not completely building because of something about missing systemd

zeebok (approved):

Issue #2658, opened 2025-10-31
Drag and drop files into a folder doesn't always work


o-alquimista:
### What Happened?

Drag and drop **more than one file** will fail to move them 50% of the time.

Drag and drop **one file** into a folder ~always works.~ EDIT: actually, it's failing too.

https://github.com/user-attachments/assets/d15a185b-f6be-4886-b54e-2666d12bbc66

### Steps to Reproduce

1. Select some files.
2. Drag them over a folder icon and immediately drop them there (the first attempt should often fail).
3. Select the files again.
4. Drag them over a folder icon and hold still until that folder is opened, then drop the files in there (it should often fail too).

### Expected Behavior

Drag and drop ~for multiple~ selected files should work. ~the same as it does for a single file~

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell

```

teamcons:
Can reproduce. Has been that way since AFAIK

jeremypw:
Sounds like a recurrence of #2419 except I think that applied to all occasions.  If it only happens sometimes it would help to identify the conditions.  For example, are there any unusual characters in the filenames?  Make sure you have v7.1.6 as some DnD related fixes were applied there.

o-alquimista:
I have 7.1.6.

In the screen recording I used some empty files named "test file" and "test file 2" — only spaces, no special characters or accented letters.

I repeat the same operation on the same files, and it works 50% of the time.

jeremypw:
Hmm, cannot reproduce with `main` at the moment. Could you try running `io.elementary.files` in a terminal to see whether and error messages are produced when the move fails?

jeremypw:
Are you copying or moving the files? They should move by default but I cannot see how the moved single file is returning to the original position after you select it- are you pressing `<Control>Z` or Cut/paste?

jeremypw:
Does the issue still occur if you allow the destination to auto-open before dropping?

teamcons:
> Are you copying or moving the files? They should move by default but I cannot see how the moved single file is returning to the original position after you select it- are you pressing `<Control>Z` or Cut/paste?

it happens for us without pressing anything

> Does the issue still occur if you allow the destination to auto-open before dropping?

Yes. It just has the icon do the phantomatic-woosh-back-to-initial-spot all the same but nothing is transferred

[...]

PR #2765, opened 2026-03-01
Add warnings during check file actions

jeremypw:
To be used to investigate #2658 

Not for merging.


teamcons:
Quick question, how can i revert from this to the "standard" files once done?

In the past i had some issues, apt install --reinstall 'ing the official package would not seem to be enough to go back to an "official" version

jeremypw:
@teamcons This branch only contains additional warning output added to the current 'main' branch.  If you want to return to the latest daily version then checkout the `main` branch then rebuilld and install that (which is what I do as I am on the daily version of elementary)

If you want the current stable release version then first uninstall this branch with `sudo ninja uninstall`.  You can then install the released version using `sudo apt reinstall pantheon-files`.  Instead of using reinstall it may be better to use `sudo apt purge pantheon-files` followed by `sudo apt install pantheon-files`.

Alternatively you can build and install the released version yourself with

`git fetch --tags`
`git checkout 7.2.0`

Delete the build directory and then run the usual build and install commands.

PR #2768, opened 2026-03-02
Systemd optional

jeremypw:
Systemd dependency is not required so build should not fail if it is absent.

In` main` the systemd dependency is marked as optional but is followed by an assert that it must be found which seems contradictory.

With this PR absence of the systemd dependency has the same effect as building with the option `'systemduserunitdir'` set to `'no'` .

For some reason the `development-target` CI was failing to find the `systemd` dependency.  Is this expected?  With this PR the development target CI passes.

### DIFF
--- a/filechooser-portal/meson.build
+++ b/filechooser-portal/meson.build
@@ -22,13 +22,12 @@ conf_data = configuration_data()
 conf_data.set('libexecdir', libexec_dir)
 
 systemd_systemduserunitdir = get_option('systemduserunitdir')
-if systemd_systemduserunitdir != 'no'
+systemd_dep = dependency('systemd', version: '>= 206', required: false)
 
-  if systemd_systemduserunitdir == ''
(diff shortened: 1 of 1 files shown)


danirabbit:
I think this is fine but we should also add `systemd-dev` to CI. It seems like it's because of package renames that wasn't building for Resolute.

Also worth noting that we require systemd for the session itself now, so if it's a burden for you to maintain optional systemd support, I don't think it would be unreasonable to just require it

danirabbit (approved):

jeremypw:
@danirabbit Yes, this was just a quick fix.  I am not entirely sure what the systemd stuff does but since it was marked "not required" and there is a build option to disable it I assumed it wasn't essential.  Presumably the FileChooser portal will still work without it? It won't affect elementary anyway since we ship systemd.   I doubt Files is used on non-systemd systems that much.

danirabbit:
Ah that's probably what it is, the portal probably requires systemd to run

PR #2771, opened 2026-03-11
Jeremypw/dnd/fallback suggested action

### LINKED ISSUE #2658: Drag and drop files into a folder doesn't always work

jeremypw:
Fixes #2658 - hopefully.

The first four commits are basically cleanup and can be ignored or split out into a separate PR. 

The final commit reworks and slightly simplifies deciding whether a file can be dropped and now ignores the selected action from  `DragContext.get_selected_action ()`  as it seems that this is now unreliable on Wayland and may return 'null'  or some other value which is not a valid DragAction.  

As a consequence of this, dropping a file onto its parent now results in a menu allowing choice between COPY and LINK actions.  This is parallel to the paste options on the context menu when using the keyboard. If desired this can be easily reverted to defaulting to refusing the drop.

TODO
 - [ ] Fix #2772 - the second operation behaves like dropping on parent folder (fails in main, triggers menu here).


danirabbit:
Can we break out the refactoring or fix the commit history so this can be rebase merged? Whichever is easier :)

jeremypw:
Splitting into smaller PRs

jeremypw:
Closing - now superceded by separate PRs

PR #2779, opened 2026-04-07
DnDHandler: use GLib.Menu

danirabbit:


danirabbit:
@jeremypw how do you activate this menu? I'm trying to test this but I'm not sure how to invoke it

jeremypw:
@danirabbit You might like to look at or merge #2771 first.  This fixes an issue with DnD on Wayland and reinstates the previous behaviour of showing the menu asking to choose between possible actions when dropping a file onto its parent.

PR #2783, opened 2026-04-27
Fix regression: Crash if too many templates

jeremypw:
After commit 1d6672a08d606a32b5c55c602cdec0dbabc575a2 Files crashes if there are too many templates. The fix is simple and implemented here.

Marked as critical but should not affect most users who have a templates folder with less than 2048 templates.


danirabbit (approved):
Ah good catch!

PR #2784, opened 2026-04-27
Workflows: update release to noble

danirabbit:
It looks like daily only builds for Noble and Resolute now? I'm not sure when this changed. The version number is 7.x which seems to indicate we didn't break API yet, but maybe that just didn't get bumped?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,4 +16,4 @@ jobs:
          GIT_USER_NAME: "elementaryBot"
          GIT_USER_EMAIL: "builds@elementary.io"
        with:
-         release_branch: 'horus'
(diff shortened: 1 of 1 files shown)


danirabbit:
@jeremypw any idea why we might have stopped building daily for Jammy?

jeremypw:
@danirabbit I just succesfully built `main` on OS7.1.  The only thing I had to do was install `libpoppler-glib-dev` by hand.  Running `sudo apt build-dep pantheon-files` did not install it.  The preview column feature added a dependency on lib-poppler.

danirabbit:
Ah I think it's because we added `systemd-dev` to fix the resolute builds. I'll push a separate jammy package and close this

Issue #2788, opened 2026-05-02
Miller columns: the right-hand column appears blank


nawan95:
### What Happened?

When I try to open a folder in Miller columns view, the folder's contents don't appear in the right column.

### Steps to Reproduce

1. Open Files app
2. Change to miller columns view
3. Try open any folder

### Expected Behavior

When a folder is opened in Miller columns, its contents should be displayed in the right-hand column.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland, This is the default)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
** (io.elementary.files:10155): DEBUG: 11:16:09.492: plugin.vala:162: Error opening git repository at file:///home/nawan: could not find repository at '/home/nawan'
** (io.elementary.files:10155): DEBUG: 11:16:09.674: plugin.vala:162: Error opening git repository at file:///home/nawan: could not find repository at '/home/nawan'
** (io.elementary.files:10155): DEBUG: 11:16:09.675: Slot.vala:373: SLOT close file:///home/nawan/Dokumen

(io.elementary.files:10155): GLib-GObject-CRITICAL **: 11:16:09.675: invalid (NULL) pointer instance

(io.elementary.files:10155): GLib-GObject-CRITICAL **: 11:16:09.675: g_signal_handlers_disconnect_matched: assertion 'G_TYPE_CHECK_INSTANCE (instance)' failed

(io.elementary.files:10155): Gtk-CRITICAL **: 11:16:09.675: gtk_tree_selection_unselect_all: assertion 'GTK_IS_TREE_SELECTION (selection)' failed
** (io.elementary.files:10155): DEBUG: 11:16:09.676: Slot.vala:119: Slot 1 destruct
** (io.elementary.files:10155): DEBUG: 11:16:09.676: AbstractTreeView.vala:30: ATV destruct
** (io.elementary.files:10155): DEBUG: 11:16:09.676: AbstractDirectoryView.vala:417: ADV destruct
** (io.elementary.files:10155): DEBUG: 11:16:09.677: plugin.vala:162: Error opening git repository at file:///home/nawan: could not find repository at '/home/nawan'
```

jeremypw:
@nawan95 Thanks for the report. I could not reproduce this issue using the `main` branch but noticed the critical messages on closing a slot and pushed a PR to fix those.

[...]

PR #2789, opened 2026-05-03
ColumnView: fix some critical messages

jeremypw:
- Wait to ensure slot container window is not null when activating a slot
 - Do not unnecessarily truncate miller view on closing (which triggers `active` signal and causes an error messages)

Not sure if this is related to #2788 or not but the log supplied there indicated one of the critical messages this PR fixes.


ryonakano:
Is there any reason why it's not `content_type != null` but `_content_type != null`?

I'm not familiar with the code here, but this looks a bit strange for me:

```vala
        if (content_type != UNKNOWN_CONTENT && _content_type != null) {
```

jeremypw:
@ryonakano I think it was because after referencing the property `content_type` once it is quicker to reference the backing variable `_content_type` as we know it is up to date. If the first reference returned `null` we don't need to try to evaluate the content_type again.

Issue #1474, opened 2020-09-13
Landing page like Code when all tabs are closed


hanaral:
It would be a lot more user friendly if there was a landing page for files. Right now the tab reopening loop is quite jarring and inconsistent with Code or Epiphany. 
I think having a landing like code should be the default, and maybe have a toggle to close the window when all tabs are closed (but that would also require that code gets a similar toggle). The page could say something like 'Pick a folder to get started', the first button could open the user's home folder, the second to open the most recently closed one. It could look like this:
![Screenshot from 2020-09-13 11-01-04](https://user-images.githubusercontent.com/58219504/93015332-80130200-f5b0-11ea-82e7-d80f99f14962.png)

There could also be a cogwheel next to the recent tabs icon in the tab bar where users can toggle whether closing the last tab closes the window. 
![Screenshot from 2020-09-13 11-03-53](https://user-images.githubusercontent.com/58219504/93015378-d718d700-f5b0-11ea-963b-c10d1a4d3e71.png)

Like this tickbox in Ephemeral:
![Screenshot from 2020-09-13 10-47-58](https://user-images.githubusercontent.com/58219504/93015307-56f27180-f5b0-11ea-9c4d-5727d9ac9977.png)

jeremypw:
The discussion at https://github.com/elementary/files/issues/327 is relevant, which is essentially the same issue but with a different proposed solution.

jeremypw:
I am struggling to think of the utility of closing the last tab in Files anyway tbh.  You can either close the app or navigate somewhere in the existing tab. The proposed entries in the landing page can already be done directly without closing the last tab. Maybe a better solution would be to stop the last tab being closed.

danrabbit:
It could be interesting to explore an alternative home page.

Something I've personally encountered is that the XDG directories are not usually very useful to me since they are:
1. Present in the sidebar by default
2. Usually better managed and navigated from within applications. Example: pictures, music, videos, etc.

It might be more useful to make Recent the new tab page or to introduce a most-frequently-used page or (as in Finder), a new view that shows both of these things

hanaral:
I understand that file browsers are not friendly things, but it would be nice to have a simple landing page with maybe a recents view like you suggested (many users don't have recents enabled, though if its designed nicely, maybe it'll change).

I don't really like the idea of having the user be presented with their home folder directly, as there aren't any user friendly things anyone would ever do there (apart from opening the folders already present in the sidebar). 

[...]

Issue #327, opened 2018-02-11
Ctrl+W is not consistent with Epiphany & Terminal


jlnr:
When only one tab is open in Files, Ctrl+W does not close the window, but instead opens the home folder again. I think that Epiphany's/Terminal's behavior is more intuitive here - Ctrl+W with only one open tab closes the window.

jeremypw:
If the design team is in agreement this is easily changed.

danrabbit:
We previously decided against this because `Ctrl + W` is the close tab shortcut, not the close window shortcut. If your intent is the close the window there is another shortcut for that. But changing this would mean that we're trying to be clever and assume your intent instead of following what you told the computer to do.

For example, if your intent is simple to close all of your open tabs so that you have a clear "workspace" and you mash `Ctrl + W` and then suddenly Files quits, this looks like a crash and does something different from what you told Files to do (Close tabs).

cassidyjames:
So, this is actually a bug with Terminal and Epiphany. 😉

jlnr:
There's some predictable consistency in the Epiphany-like behavior that I enjoy: If you press Ctrl+W when only one tab is open, the window closes. Likewise, if you press Alt+F4 when only one window is open, the application quits.

Other apps like Scratch/Code and most "workbench apps" (where mashing Ctrl+W is especially common) don't close the window when the last tab is closed (or dragged out from the tab bar). Instead, they display an empty state screen with an empty tab bar, which ignores further Ctrl+W presses and typically makes it easy to get started with the next project.

Files is an outlier because it neither closes the window, nor does it display an empty state screen. It immediately re-opens a new tab with the home folder, and in a way, this is also a guess at what I want to do next; it could also keep the user's last tab open, or show a list of recent locations below an empty tab bar.

I don't think any app's behavior is unreasonable, but I feel that changing Files to match Epiphany/Terminal would make the OS both more internally consistent and more compatible with muscle memory from third-party apps (all browsers seem to work like Epiphany now).

I guess this is the best case I can make for this :)

jeremypw:
Elementary apps are supposed to restore their state at close when reopened.  If the last tab has been closed what is intended to be restored?  In fact Terminal restores the tab that was closed which seems illogical.  It seems more logical for closing the last tab to put the app in a sensible "initial" state.  In the case of Code then there is a sensible "no document" state but that is not the case for Files or Terminal. Showing the user home directory is not an unreasonable default state, however.

artemanufrij:
[...]

PR #2795, opened 2026-05-19
Close window when closing the only remaining tab 

stan-janssen:
This closes the Files window when the only tab in that window is closed:

https://github.com/user-attachments/assets/7fbab8c1-c65e-4e2c-9b72-8093bdee33e8

This makes the behavior consistent with other default apps like Terminal and Web, which also close their window when the last tab closes.

This also makes it possible to close a Files window using just the keyboard. Previously, one could only close tabs using `ctrl + W`, or close all windows using `ctrl + Q`, but not close just one window.

### DIFF
--- a/src/View/Window.vala
+++ b/src/View/Window.vala
@@ -407,7 +407,7 @@ public class Files.View.Window : Hdy.ApplicationWindow {
         tab_view.close_page_finish (page, true);
 
         if (tab_view.n_pages == 0) {
-            add_tab.begin (default_location, PREFERRED, false);
+            quit ();
         }
 
         return Gdk.EVENT_STOP;


jeremypw:
There has been some discussion as to the appropriate behaviour for a file manager when closing the last tab with different solutions proposed. See #327 and #1474.   I do not personally see the utility of closing an app by closing a tab even though browsers and Terminal do.  I understand the possible need for a way of closing just the current window with the keyboard but that is a different issue.

@danirabbit Would you like to see consistency between Files, Terminal, Code (and Browser)? If so what should be the common behaviour? Or is the correct behaviour dependent on the app?

stan-janssen:
Ah, I did not know that this was intended behavior, my bad.

I see that some excellent points were made in #327, so I can only add my intent here: My intent is: I'm done with these tabs, I don't want them restored at any point in the future, and I want them all to go away because I'm done with my file operation for now. Files is not really an application that I'm "doing work in", so to speak, so more often than not, if I close all my tabs, I don't want to "continue using Files". I press `<Ctrl>+W` the exact number of times that there are tabs open, and I expect the last one to also close the last window, just like Terminal and Web. Right now it feels like the app is fighting me, I'm startled that closing the last tab does note make the thing "go away", so to speak.

I think this is also what used to happen on Mac OS (it's been quite a few years since I've last used it), and I think it's also what happens in browsers on Windows.

The current re-opening-a-new-tab behavior has a bug by the way: pressing `<Ctrl>W` twice in succession with only one tab open reliable crashes the app (see attached gdb log of opening a window and quickly pressing `<Ctrl>W` twice). This might seem like a contrived thing, but in a window with a bunch of tabs open, if you press `<Ctrl>W` one time too many, it crashes the entire app and takes down all the other windows as well.

(Ironically, that was the exact scenario that danirabbit [described here](https://github.com/elementary/files/issues/327#issuecomment-366540402), except right now Files *actually* crashes when you mash `<Ctrl>W` one time too many, instead of just closing the last window)

[files-crash.txt](https://github.com/user-attachments/files/28023104/files-crash.txt)


If we don't want to change the behavior in Files, I can make a fix to this crashing bug. I'll follow the discussion here and see where we land. Thanks for taking the time anyway!

jeremypw:
Although elementaryos tries to avoid unnecessary settings, if there are really two equally valid behaviours then adding another switch to the app menu is a possibility.  Thanks for pointing out that crash - it certainly needs fixing.

danirabbit:
My personal opinion is that I don't usually love when one button does two different things, especially if one of those things is destructive.

We do however run into an issue where we don't have full control over the behavior of the web browser, so if our top priority is consistent behavior we might just have to follow that unless we can get upstream to carry a special behavior for us. 

It sounds like what we might be hearing from folks though is that they're using the keyboard shortcut here and not the tab close button. So perhaps we should consider that the keyboard shortcut would close the window if there's only one tab but keep the current behavior for clicking the tab close button? Would that solve this issue for you @stan-janssen?

jeremypw:
I agree that having two actions from the same button feels wrong, but the button itself always closes the tab - what happens afterwards may vary.

At the moment there are three different behaviours across (at least) four system apps - Files, Terminal, Code, Browser.  I notice that Gnome Text Editor has yet another behaviour - that is to hide the tabbar when there is only one tab essentially preventing closing the last tab with the mouse (although the shortcut still works).  If the tab is closed with the shortcut then it is replaced by a landing page, not a new document - mirroring Code.  The same happens with Gnome File Manager (nautilus) except the window closes  if the tab is closed with the shortcut - as is proposed here.

On reflection, I could live with closing the Files window on closing the last tab as, like Terminal, there is no sensible "landing page" for this type of app that I can think of.  Then 3 out of 4 of the system apps with a tabbed view would behave the same way. Code can be regarded as different as an editor. Of course, the "no tabs" state cannot be restored - you will still end up with a home directory tab on restarting Files if it is the primary window that closes.

The home directory would still be only one click (bookmark) or shortcut (Alt+Home) away so nothing is really lost and we gain a shortcut for closing just the current window (indirectly)

jeremypw:
Will need to consider the text for the "Close Tab" option in the tab context menu (which is linked to the `Ctrl+W` shortcut).  I guess this should change to "Close Tab and this Window" or something when there is only tab?

jeremypw:
I think it is safe enough to merge this to see what, if any, feedback there is from the daily users.   Users only have to reopen a Files window if it closes unexpectedly. I would not have thought there would be much of a learning curve to using other ways of getting a Home tab than closing the last tab (bookmark or `<Alt>Home`).  We can iterate on the finer details before the next release.

Issue #1719, opened 2021-05-22
"Empty the Trash" deletes all files without any sort of warning.


Blast-City:
## What Happened

As soon as the "Empty the Trash" button is pressed the files are deleted without any sort of warning/confirmation that this will be permanent.

## Expected Behavior

Clicking "Empty the Trash" should warn the user that this operation is irreversible.

## Steps to Reproduce

1. Create a file
2. Right click on the file and select "Move to Trash"
3. Go to Trash and click on "Empty the Trash"

## Platform Information

elementary OS 5.1.7 Hera

jeremypw:
I am not sure we need a confirmation dialog but the text should be e.g. "Permanently delete all trash" maybe with destructive action styling.

Blast-City:
I disagree. Accidents happen, especially with "tap to click" enabled so it's always good to have a safe net for that cases.

What are the advantages of not having confirmation dialog?

danrabbit:
The entire trash is a bit of an undo/confirm flow isn't it? If you've moved something to the trash, why wouldn't you expect it to be deleted?

jeremypw:
I don't think there is a way of permanently deleting anything with only one action (or there shouldn't be).  The second action should always be preceded by some indication that the operation is irreversible.  Sometimes it is a dialog (e.g. after using the shortcut) but otherwise it is text/color on the button/menuitem. A third required action seems excessive.

jeremypw:
May I ask whether you have actually lost files through accidentally deleting them from the trash?

Blast-City:
> The entire trash is a bit of an undo/confirm flow isn't it? If you've moved something to the trash, why wouldn't you expect it to be deleted?

That makes sense, but since we're talking about an irreversible action I don't see why not just give an extra warning. Empty the Trash is something that people use sporadically so it's not that bothersome to have to do an extra click.

> May I ask whether you have actually lost files through accidentally deleting them from the trash?

No, but I was testing something and then I moved the test files to Trash. When I went to empty the Trash I was really surprised that it permanently removed everything without any confirmation. I'm used to see a confirmation on other Apps so it felt like something was off.

Blast-City:
Another issue that can occur it when you're trying to restore one file, the "restore form Trash" context menu entry is so close to the "delete permanently" that it takes just a twitch and your file is removed forever.

jeremypw:
[...]

PR #2796, opened 2026-05-19
Change the shortcut for Focus Sidebar to not interfere with text editing

### LINKED ISSUE #2677: When renaming a file pressing CTRL+LEFT makes the focus move to the left column of Files
### What Happened?

When renaming a file, pressing CTRL+LEFT makes the focus of the window move to the left column of Files (bookmarks, storage, network, etc.) instead of jumping the cursor to the word on the left. Pressing CTRL+RIGHT seems to work fine.

See the video below where I first select the file to rename, press F2 and then press CTRL+RIGHT which works and then press CTRL+LEFT which switches the focus to the left column.

https://github.com/user-attachments/assets/c78d2bfb-e352-4ea6-9962-a7bf6d3ee640

### Steps to Reproduce

1. Select a file to rename and press F2
2. Press CTRL+LEFT and notice it switches the focus to the left column of Files

### Expected Behavior

Pressing CTRL+LEFT should act like pressing CTRL+RIGHT while renaming a file and moving/jumping between words while renaming a file.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

stan-janssen:
The existing `<Ctrl> + Left` would interfere when editing a filename.

Closes #2677.

### DIFF
--- a/src/View/Window.vala
+++ b/src/View/Window.vala
@@ -68,7 +68,8 @@ public class Files.View.Window : Hdy.ApplicationWindow {
         {"tabhistory-restore", action_tabhistory_restore, "s" },
         {"forward", action_forward, "i"},
         {"back", action_back, "i"},
-        {"focus-sidebar", action_focus_sidebar}
+        {"focus-sidebar", action_focus_sidebar},
+        {"focus-current-container", action_focus_current_container}
     };
 
     private static uint window_id = 0;
@@ -162,6 +163,7 @@ public class Files.View.Window : Hdy.ApplicationWindow {
             marlin_app.set_accels_for_action ("win.tab::TAB", {"<Shift><Ctrl>K"});
             marlin_app.set_accels_for_action ("win.tab::WINDOW", {"<Ctrl><Alt>N"});
             marlin_app.set_accels_for_action ("win.focus-sidebar", {"<Ctrl><Alt>Left"});
+            marlin_app.set_accels_for_action ("win.focus-current-container", {"<Ctrl><Alt>Right"});
         }
 
         build_window ();
@@ -1004,6 +1006,10 @@ public class Files.View.Window : Hdy.ApplicationWindow {
 
     private void action_focus_sidebar () {
         sidebar.focus ();
+    }
+
+    private void action_focus_current_container () {
+        grab_focus ();
     }
 
     private void before_undo_redo () {
--- a/src/View/Window.vala
+++ b/src/View/Window.vala
@@ -161,7 +161,7 @@ public class Files.View.Window : Hdy.ApplicationWindow {
             marlin_app.set_accels_for_action ("win.back(1)", {"<Alt>Left", "XF86Back"});
             marlin_app.set_accels_for_action ("win.tab::TAB", {"<Shift><Ctrl>K"});
             marlin_app.set_accels_for_action ("win.tab::WINDOW", {"<Ctrl><Alt>N"});
-            marlin_app.set_accels_for_action ("win.focus-sidebar", {"<Ctrl>Left"});
+            marlin_app.set_accels_for_action ("win.focus-sidebar", {"<Ctrl><Alt>Left"});
         }
 
         build_window ();


jeremypw:
See also https://github.com/elementary/code/pull/1719.  It would be good to agree on a common shortcut to switch to and from the sidebar. I think `<Ctrl><Alt>Left` `<Ctrl><Alt>Right` is more memorable on balance. Not sure how to make these shortcuts discoverable though.

stan-janssen:
Thanks, I will update this PR to change the keyboard shortcut to `<Ctrl><Alt>Left`.

stan-janssen:
(A thing that comes to mind: do people with a RTL (Right-To-Left) setting see the sidebar on the right side? And isn't a fixed directional-key shortcut confusing in that case? I assume that this won't be the only place this is encountered...)

Dipanshusinghh:
Heiy  everyone.
It’s great to see we are keeping the keyboard shortcuts consistent across the elementary ecosystem. Using <Ctrl><Alt>Left and <Ctrl><Alt>Right has felt quite natural while implementing it in Code (#1719)..
Regarding the RTL layout question, that's a very valid and interesting point! I imagine GTK might handle flipping directional keybindings automatically based on the system's text direction, but it's definitely something good to keep an eye on for the future..
Thanks for working on this

stan-janssen:
I installed the Arabic language and took a look, and indeed the sidebar is on the right, and I still have to press `<Ctrl><Alt>Left` to get to it, which is not great. The same goes for navigating in column view by the way, the right arrow opens a new column (on the left).

I don't think this should block this PR, because it is not a regression to the current status, but I wonder if there is something in some of the underlying libraries that might help with abstracting this. If there isn't, perhaps directional arrow keys should only be used very cautiously in shortcuts.

jeremypw:
I agree that dealing with RTL locales should be dealt with but can be done in another PR.  It should be simple enough to alter the actions depending on the StateFlagsof the window (DIR_RTL or not).

jeremypw (changes requested):
This works as expected, thanks. I think that we should change the corresponding `<Ctrl>Right` shortcut as well though to maintain consistency?

stan-janssen:
> I think that we should change the corresponding `<Ctrl>Right` shortcut as well 

I can't find an existing `<Ctrl>Right` shortcut. Where is it?

Should I implement a `focus_tab_view` action (which I can't find either) and connect a `<Ctrl><Shift>Right` shortcut to it? Or am I not looking correctly?

I'd be happy to add it.

jeremypw:
Hmm, both `<Ctrl>Right` `Right` do work in `main` in LTR mode - not sure why. In RTL mode. `Left` returns to the view (as expected) but not `<Ctrl>Left`. Within the view the movement is the wrong way round so that needs fixing (elsewhere) as well. I guess this behaviour must be built into Gtk somewhere because I cannot find in the Files code either.

I don't think adding `<Ctrl><Alt>Right` as a paired action has any downside though so may as well do it.

stan-janssen:
I added an explicit accelerator for focusing on the current container (this re-uses the existing focus logic that was already present).

I'll add a separate PR (that builds on this one) that ensures proper directionality of accelerators in RTL languages.

Dipanshusinghh:
@stan-janssen 
Thanks for adding the paired shortcut. reusing the existing focus logic for returning to the current container seems like a clean solution, & handling the RTL directionality in a follow-up PR makes sense to me.

Looks good from my side🎉

src/View/Window.vala
```diff
@@ -161,7 +162,8 @@ public class Files.View.Window : Hdy.ApplicationWindow {
             marlin_app.set_accels_for_action ("win.back(1)", {"<Alt>Left", "XF86Back"});
             marlin_app.set_accels_for_action ("win.tab::TAB", {"<Shift><Ctrl>K"});
             marlin_app.set_accels_for_action ("win.tab::WINDOW", {"<Ctrl><Alt>N"});
-            marlin_app.set_accels_for_action ("win.focus-sidebar", {"<Ctrl>Left"});
+            marlin_app.set_accels_for_action ("win.focus-sidebar", {"<Ctrl><Alt>Left"});
+            marlin_app.set_accels_for_action ("win.focus-tabview", {"<Ctrl><Alt>Right"});
```
jeremypw: The action must be `win.focus-current-container` otherwise it doesn't work.

jeremypw (changes requested):
Not working at the moment because of wrong action name in shortcut definition.

jeremypw:
Might be better to focus on the follow-up PR which superceded this one anyway.

stan-janssen:
Sorry about that, I see that it was updated in the other branch but I forgot to rename it her. I updated and rebased both of the PRs!

src/View/Window.vala
stan-janssen: Sorry about that, I fixed it now (and rebased the other branch)

jeremypw (approved):
Works well now, thanks.

PR #2797, opened 2026-05-24
Correct RTL accelerator arrow keys

stan-janssen:
This corrects accelerators that include a Left or Right arrow key when using a Right-to-Left (RTL) language, where UI elements are draw from right to left.

This includes:

- switching focus between the sidebar and the main tab view (the sidebar is on the right side in an RTL situation); 
- the 'drill down' and 'drill up' keys when navigating in list or Miller views;
- the arrow keys for selecting the previous and next icon in icon view (the icons are also drawn right to left, with the 'next' icon being on the left of the 'previous' icon in an RTL situation;
- the 'back' and 'forward' actions that map to the buttons in the navigation bar (including the tooltips for these buttons)

### DIFF
--- a/src/View/IconView.vala
+++ b/src/View/IconView.vala
@@ -243,11 +243,21 @@ public class Files.IconView : Files.AbstractDirectoryView {
 
     /* Override native Gtk.IconView cursor handling */
     protected override bool move_cursor (uint keyval, bool only_shift_pressed, bool control_pressed) {
+        uint prev_key;
+        uint next_key;
+        if (Gtk.StateFlags.DIR_RTL in get_style_context ().get_state ()) {
+            prev_key = Gdk.Key.Right;
+            next_key = Gdk.Key.Left;
+        } else {
+            prev_key = Gdk.Key.Left;
+            next_key = Gdk.Key.Right;
+        }
+
         Gtk.TreePath? path = get_path_at_cursor ();
         if (path != null) {
-            if (keyval == Gdk.Key.Right) {
+            if (keyval == next_key) {
                 path.next (); /* Does not check if path is valid */
-            } else if (keyval == Gdk.Key.Left) {
+            } else if (keyval == prev_key) {
                 path.prev ();
             } else if (keyval == Gdk.Key.Up) {
                 path = up (path);
--- a/src/View/ListView.vala
+++ b/src/View/ListView.vala
@@ -160,36 +160,41 @@ namespace Files {
         }
 
         protected override bool on_view_key_press_event (uint keyval, uint keycode, Gdk.ModifierType state) {
+            uint up_key;
+            uint down_key;
+            if (Gtk.StateFlags.DIR_RTL in get_style_context ().get_state ()) {
+                up_key = Gdk.Key.Right;
+                down_key = Gdk.Key.Left;
+            } else {
+                up_key = Gdk.Key.Left;
+                down_key = Gdk.Key.Right;
+            }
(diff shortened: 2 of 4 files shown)


src/View/IconView.vala
```diff
@@ -244,10 +244,22 @@ public class Files.IconView : Files.AbstractDirectoryView {
     /* Override native Gtk.IconView cursor handling */
     protected override bool move_cursor (uint keyval, bool only_shift_pressed, bool control_pressed) {
         Gtk.TreePath? path = get_path_at_cursor ();
+
+        uint prev_key;
+        uint next_key;
+
```
jeremypw: Unnecessary whitespace

src/View/IconView.vala
```diff
@@ -244,10 +244,22 @@ public class Files.IconView : Files.AbstractDirectoryView {
     /* Override native Gtk.IconView cursor handling */
     protected override bool move_cursor (uint keyval, bool only_shift_pressed, bool control_pressed) {
         Gtk.TreePath? path = get_path_at_cursor ();
```
jeremypw: I would move this next to the clause that uses it

src/View/IconView.vala
```diff
@@ -244,10 +244,22 @@ public class Files.IconView : Files.AbstractDirectoryView {
...
+
```
jeremypw: Unnecessary whitespace

src/View/ListView.vala
```diff
@@ -160,36 +160,43 @@ namespace Files {
         }
 
         protected override bool on_view_key_press_event (uint keyval, uint keycode, Gdk.ModifierType state) {
+            uint up_key;
+            uint down_key;
+
```
jeremypw: Unnecessary whitespace

jeremypw (commented):
There is some whitespace I would consider superfluous.  I try not to leave space between a variable declaration and where it used.  Not all existing code may comply!

I have not tested this yet but looks like it does not fully cope with a change in LTR/RTL state "on the fly" as the shortcuts are set up on construction.  May be better to have fixed shortcuts (e.g. for `action-focus-left` and `action-focus-right`) and deal with the state in the handlers.  Not sure how important this is in practice though.

stan-janssen:
Thanks for the review. I'll remove the superfluous whitespace (I'm not completely "fluent" yet in the code style of the project, so thanks for pointing these things out).

>  it does not fully cope with a change in LTR/RTL state "on the fly" as the shortcuts are set up on construction

The left/right arrow keys for ascending and descending through the hierarchy and for navigating through Icon view do change "on the fly", but the ones for focusing different areas and the ones for Forward / Back do not.

I'm not aware of a use case for changing these things on the fly (I think it can only be done using the interactive debugger?) so I don't think it's a big issue myself.

jeremypw:
Yes, you are probably right that its not a problem in practice.  I tried switching to an RTL language and although some widgets (e.g. TextView) start typing RTL, the layout of widgets does not change.  I guess you either need to use the interactive debugger or change the system locale in settings (needing a new session). I'll test tomorrow.

stan-janssen:
I tested this using:

```
LANGUAGE=ar_SA.UTF-8 io.elementary.files
```

afbeelding


(This requires the arabic language to be installed from System Settings)

jeremypw:
@stan-janssen Could you resolve the conflicts please?

jeremypw (approved):
Let's go!  Thanks.

PR #2798, opened 2026-05-25
Get drop target file info on demand

jeremypw:
Fixes #2772 
May also addess #2658 (tbc) 

extracted from #2771


ryonakano (approved):
Confirmed this fixes #2772.

PR #2799, opened 2026-05-26
Do not use deprecated DragAction.PRIVATE

jeremypw:
Gdk.DragAction.PRIVATE does not exist in Gtk4 and does not seem to be doing anything here so lose it.

### DIFF
--- a/libcore/DndHandler.vala
+++ b/libcore/DndHandler.vala
@@ -406,7 +406,7 @@ namespace Files {
             } else if (dest.is_executable ()) {
                 actions |= (Gdk.DragAction.COPY |
                            Gdk.DragAction.MOVE |
-                           Gdk.DragAction.LINK
+                           Gdk.DragAction.LINK);
             } else {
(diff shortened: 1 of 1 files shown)


libcore/DndHandler.vala
```diff
@@ -406,8 +406,7 @@ namespace Files {
             } else if (dest.is_executable ()) {
                 actions |= (Gdk.DragAction.COPY |
                            Gdk.DragAction.MOVE |
-                           Gdk.DragAction.LINK |
-                           Gdk.DragAction.PRIVATE);
+                           Gdk.DragAction.LINK
```
stsdc: Missing `;`

stsdc (changes requested):

libcore/DndHandler.vala
jeremypw: Oops!  Now fixed.

stsdc (approved):

PR #2806, opened 2026-06-16
Fix IconView blank after cut/paste between parent-child folders

### LINKED ISSUE #2801: IconView fails to refresh after reversing a cut/paste operation between parent and child folder
### What Happened?

Cut/pasted a file between two folders in separate tabs.  The file appeared in the second folder OK.  Moved the cut/pasted the file back in to the first tab.  The view blanked and did not refresh until the view was resized or the refresh button was pressed. The file was successfully moved.  No terminal warnings were given.  Only happens with IconView.  Moreover, it only happens between two tabs in the same window.  If the second tab is in a separate window then the view redraws correctly.

Maybe related to some other issues involving refreshing but this is the easiest to reproduce

### Steps to Reproduce
Using IconView:
1. Create a folder and within it a child folder and open them in separate tabs A & B
2. Create a file in A
3. Cut/paste the file from A to B
4. Cut/paste the file from B to A or undo (`<Ctrl>Z`) while displaying tab A
5. Notice tab A appears blank until refreshed

### Expected Behavior

The view always updates when a file is pasted into it.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

Early Access or Compiled from git

### Log Output

```shell

```

### Hardware Info

_No response_

Dipanshusinghh:
Fixes #2801

Added `child_notify_frozen` boolean flag in `IconView.vala` and 
`AbstractTreeView.vala` to ensure `freeze_child_notify()` and 
`thaw_child_notify()` calls are always balanced, preventing IconView 
from going permanently blank after cut/paste operations.

### DIFF
--- a/src/View/AbstractTreeView.vala
+++ b/src/View/AbstractTreeView.vala
@@ -20,6 +20,7 @@ namespace Files {
     /* Implement common features of ColumnView and ListView */
     public abstract class AbstractTreeView : AbstractDirectoryView {
         protected Files.TreeView tree;
+        private bool child_notify_frozen = false;
         protected Gtk.TreeViewColumn name_column;
 
         protected AbstractTreeView (View.Slot _slot) {
@@ -325,15 +326,17 @@ namespace Files {
 
         // For scrolling
         protected override void freeze_child_notify () {
-            tree.freeze_child_notify ();
+            if (!child_notify_frozen) {
+                tree.freeze_child_notify ();
+                child_notify_frozen = true;
+            }
         }
 
         protected override void thaw_child_notify () {
-            // Do not prematurely thaw tree when loading
-            if (!tree_frozen) {
+            if (child_notify_frozen) {
                 tree.thaw_child_notify ();
+                child_notify_frozen = false;
             }
-
         }
     }
 
--- a/src/View/IconView.vala
+++ b/src/View/IconView.vala
@@ -18,6 +18,7 @@
 
 public class Files.IconView : Files.AbstractDirectoryView {
     protected new Gtk.IconView tree;
+    private bool child_notify_frozen = false;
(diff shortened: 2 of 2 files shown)


jeremypw (changes requested):
This does seem to fix the issue for me, I am afraid.  According to the documentation I have found there is no need to pair `freeze_child_notify` and `thaw_child_notify` calls (unlike object.freeze_notify and object.thaw_notify).  When testing, make sure that you move a file between a parent folder and a child folder and then undo.

Dipanshusinghh:
Understod ! i will update this by today☺️

jeremypw:
You are right that the calls to the child_notify functions are unbalanced though.  This whole area of the code is rather ugly and is a hack to improve the performance of IconView.  Whether it is worth spending a lot of time making it more elegant is questionable as it will go away once we switch to Gtk4 dynamic views (unless it causes an issue).

Dipanshusinghh:
Thnks.
 for the clarification since the documentation says pairing is not strictly necessary should I revert the boolean flag approach and instead focus on fixing the root cause of why thaw_child_notify() is being called without a prior freeze_child_notify()? Or would you prefer a different approach? Happy to update as per your suggestion

jeremypw:
@Dipanshusinghh You need to identify and fix the root cause of #2801 to complete this PR.  If you want to improve the code for other reasons it would be better to open a separate PR for that.   

The unpaired calls do not seem to be causing any actual issue do they?

Note that the issue only occurs with parent - child source and destination folders so the cause must have something to do with that.  The cause could be quite obscure so I would try to determine (for example) whether the model for the view is being filled and not displayed by adding debugging messages or otherwise. Or maybe its a race between the view being unthawed and being drawn.

PR #2810, opened 2026-08-21
Window: cleanup update_location_bar ()

jeremypw:
- Lose unrequired calls to `update_location_bar ()
 - Absorb it into update_headbar ()

Part of #2661


danirabbit (approved):
Seems to work for me

PR #2815, opened 2026-08-23
Add latte color tag and simplify

### LINKED ISSUE #2814: Color tags: Latte is missing
### What Happened?

You can choose any color of the elementary palette as color tag on right click on a file. Except latte.

### Steps to Reproduce

Right click on a file or folder

### Expected Behavior

can also tag latte

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #2814

In order to add a color while not affect users' previously assigned color tags and keeping the context menu widget colors in a sensible order it was necessary to distinguish the button index from the color index.

 - Add a `index` member to color button pointing to the required color in COLOR_TAGS
 - Add a `latte` color to the end of COLOR_TAGS and document origin
 - Simplify identifying which button was pressed and remove magic constants
 - SImplify handling RTL

The simplification could be done in a separate PR if adding another color is not approved.

### DIFF
--- a/libcore/Preferences.vala
+++ b/libcore/Preferences.vala
@@ -20,9 +20,21 @@ namespace Files {
     public static Preferences? preferences = null;
 
     public class Preferences : Object {
-        /* First element set to null in order that the text renderer background is not set */
+        /* We have to hard code the colors while we are using renderers - cannot use css */
         public const string?[] TAGS_COLORS = {
-            null, "#64baff", "#43d6b5", "#9bdb4d", "#ffe16b", "#ffc27d", "#ff8c82", "#f4679d", "#cd9ef7", "#a3907c", "#95a3ab", null
+            null,       // No color set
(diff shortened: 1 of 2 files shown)


teamcons:
Pantheon uses the palette of colors in a few places, and granite apps too.
I did solve it in granite apps by making them an enum, maybe we could granite it up?

teamcons:
(Thank you <3 )

jeremypw:
> Pantheon uses the palette of colors in a few places, and granite apps too.
> I did solve it in granite apps by making them an enum, maybe we could granite it up?

The recommended way is to use CSS with the elementary stylesheet but that does not work with cell renderers unfortunately. Once we have ported to Gtk4 we should be able to use dynamic views with the file items being widgets which we can then style without having to hardcode the color values.

PR #2817, opened 2026-08-27
Better preview behaviour with small window

### LINKED ISSUE #2816: preview panel takes over window
### What Happened?

When clicking a file with a small Files window, the preview panel shoves the rest of the folder's contents to an unusable sliver

### Steps to Reproduce

https://github.com/user-attachments/assets/a036a65e-e3d2-4dfb-ad70-659c960083a5

### Expected Behavior

It should take up half width at most

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #2816 

With a window narrower than the miller column width and preview column width combined we have to compromise.  The paned pack parameters are tweaked so the available area is divided equally between the two (subject to manual positioning of divider) and horizontal scrolling is enabled in the preview window. 

The drawback of this is that it is true even for larger windows so the user may still need to adjust the divider or window size for best results.

Optimising the layout automatically will require more complicated code afaict.

### DIFF
--- a/src/View/DetailsColumn.vala
+++ b/src/View/DetailsColumn.vala
@@ -246,7 +246,7 @@ public class Files.View.DetailsColumn : Gtk.Bin {
         var scrolled = new Gtk.ScrolledWindow (null, null) {
             child = box,
             propagate_natural_height = true,
-            hscrollbar_policy = NEVER
+            hscrollbar_policy = AUTOMATIC
         };
 
         child = scrolled;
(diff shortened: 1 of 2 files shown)


jeremypw:
Still not sure this is optimal.

jeremypw:
I think this is the best *simple* solution so marking ready.

Issue #137, opened 2017-09-19
No progress indication when trashing many small files after a restoring a large number of files


jeremypw:
If "Empty the Rubbish Bin" button is pressed when the trash contains many (1000's) small(ish) files, the interface freezes for seconds and no progress indication is given.  However, the trash is successfully emptied.

jeremypw:
The same thing happens when trashing very many (1000's) of smallish files.  Eventually the app unfreezes and shows an updated view but this takes many seconds.

jeremypw:
Added to improve performance project

jeremypw:
Transferred to more specific file transfer/progress indication project.

jeremypw:
Using the latest version with 5000 small files, trashing showed a progress indicator as expected on the first attempt but on subsequent attempts (after undoing) it did not appear.  There seemed to be some background activity slowing Files perhaps related to the undo manager.

Emptying the files from trash showed a "Preparing ..." dialog but this never changed to a progress dialog even though the view behind showed that files were being emptied from trash.

As the behaviour is different for moving to trash and deleting from trash, this report should be split into two.

jeremypw:
Edited description to focus one of the issues.

jeremypw:
Issue still present September 2026

vjr:
I might give this one a shot see if I can tackle it/

PR #2819, opened 2026-08-28
Sync transfers for realistic progress bar updates

### LINKED ISSUE #2818: Transfer progress bar finishes too fast
### What Happened?

Even if I copy a relatively large file to a relatively slow (usb stick) destination, the progress bar appears to finish unrealistically quickly.

The actual transfer is still in progress (kernel cache?) and if I (a average Joe user) unplug the usb it might result in data loss? Certainly if I (a little more savvy average Joe) click the unmount button it just spins for a long while until the transfer completes during which time I might wonder whether something has hung.

### Steps to Reproduce

See:

https://github.com/user-attachments/assets/1c264998-69c7-4bba-993a-635e35de5d36

### Expected Behavior

Progress bar should be realistic.

### OS Version

Early Access

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

vjr:
Fixes #2818 

edit: See https://github.com/elementary/files/pull/2819#issuecomment-5561154542 for a potential way simpler approach.

BEFORE:

[files-progress-before.webm](https://github.com/user-attachments/assets/1c264998-69c7-4bba-993a-635e35de5d36)

AFTER:

[files-progress-after.webm](https://github.com/user-attachments/assets/e8ba1bb4-4c19-4405-ba6f-d5a306b8137d)


vjr:
I see the progress bar dialog is not "modal" so if I click the unmount button while the transfer is in progress Files crashes - but this is a different issue that already exists in the current main branch.

jeremypw:
The value of `15` was set in 2011 by the original author (!)  Certainly not appropriate for modern hardware especially SSD.  I vaguely remember thinking that the linked issue was something to do with caching the write but if it fixes the issue for you there seems little harm in reducing this value.  I'll check the code where this value is used though.

jeremypw:
> I see the progress bar dialog is not "modal" so if I click the unmount button while the transfer is in progress Files crashes - but this is a different issue that already exists in the current main branch.

The dialog is intended to be modal so we will have to block unmounting of any volume being written to (I am surprised the system does not do that) and show a warning if necessary.  Doesn't the system show a "this volume is busy" warning?

jeremypw:
I wonder whether this constant can be removed altogether or reduced to a small value :thinking:  How fast can a reasonably accurate transfer rate be known?  Maybe better to use a minimum number of bytes transferred rather than a fixed time?

vjr:
> > I see the progress bar dialog is not "modal" so if I click the unmount button while the transfer is in progress Files crashes - but this is a different issue that already exists in the current main branch.
> 
> The dialog is intended to be modal so we will have to block unmounting of any volume being written to (I am surprised the system does not do that) and show a warning if necessary. Doesn't the system show a "this volume is busy" warning?

No warning, the app quits (crashes) if I click unmount while transfer is in progress. Should I open another PR to force the transfer dialog to be modal and not allow interacting with the app?

vjr:
> I wonder whether this constant can be removed altogether or reduced to a small value 🤔 How fast can a reasonably accurate transfer rate be known? Maybe better to use a minimum number of bytes transferred rather than a fixed time?

I've reduced the const down from 15 seconds to 3 seconds - seems reasonable for both slow and fast disks/hardware... if the hardware is fast the copy dialog completes and closes quickly anyway, if the hardware is slow then about 3 seconds seems okay for the transfer speed indication to show up?

vjr:
> The value of `15` was set in 2011 by the original author (!) Certainly not appropriate for modern hardware especially SSD. I vaguely remember thinking that the linked issue was something to do with caching the write but if it fixes the issue for you there seems little harm in reducing this value. I'll check the code where this value is used though.

Yep I thought as much haha. Thanks for looking at this PR. I think we can leave this reduced value in place for now, dont want to mess with further code changes to use a different way of calculating the duration before transfer speeds show up.

flodavid:
> Doesn't the system show a "this volume is busy" warning?

I have had the same issue. Another app failed to unmount one of my hard drive volume saying “This volume is busy”. So I tried with Files and it just crashed.

jeremypw:
@flodavid Could you open an issue regarding your crash please? I'll try and look into file transfers again soon.

vjr:
> @flodavid Could you open an issue regarding your crash please? I'll try and look into file transfers again soon.

@jeremypw do you want me to make the transfer dialog modal thus preventing user interaction with the main app while it's visible? i can do that in this PR itself presuming it's a "one liner" change?

jeremypw:
@vishal No, I think how Files deals with trying to unmount a busy volume is better addressed with a separate issue/PR.  File transfers should be non-modal and occur in the background if possible.  Not sure how to stop people pulling out a USB stick prematurely We need to get a true (i.e. unaffected by caching) indication of whether a file transfer is ongoing/complete and show a warning and/or style the sidebar item approriately.

vjr:
> @Vishal No, I think how Files deals with trying to unmount a busy volume is better addressed with a separate issue/PR. File transfers should be non-modal and occur in the background if possible. Not sure how to stop people pulling out a USB stick prematurely We need to get a true (i.e. unaffected by caching) indication of whether a file transfer is ongoing/complete and show a warning and/or style the sidebar item approriately.

@jeremypw okay fair enough - so is this PR acceptable as an interim solution to make it more evident to users that a transfer is still in progress and to not unplug or unmount while so? further fixes/improvements in future PRs?

jeremypw:
@vjr Yes. I wasn't able to test this today.  Should do it tomorrow.

jeremypw (commented):
This appears to slow file transfers significantly. Maybe because the progress is more realistic?  But it also seems to slow internal transfers between SSDs noticeably.  I wonder if the syncing can be done less often and/or only when transferring to an external drive?  Or should that wait for a future PR?

I noticed several areas that need cleaning up and reviewing in the file operations area (and it is still awaiting complete transfer to Vala) anyway.

I experimented with only syncing when a certain number of bytes had been transferred which seemed to reduce the impact on speed.

vjr:
Yes I see the impact of the `sync ()` calls slowing down transfers on **_fast_** media , it was not noticeable on the very slow media.

Plus, it might be that the global posix `sync ()` call is not a good thing especially if there are multiple transfers in progress.

Marking this PR as draft while I come up with a better approach.

jeremypw:
> it was not noticeable on the very slow media.

I tested on a slow antique MTP player tranferring many small (100 byte) files and the *reported* transfer rate was about half with this PR.  I must admit I did not manually time the complete transfer but cancelled before it finished. The situation is complicated by the fact that when transferring many small files the transfer slows down so the predicted time left is not very helpful - it often goes up instead of down.  This was reported long ago but has never really been diagnosed.  I think it may be to do with the "undo" processing taking longer and longer but I haven't dug into file operations nitty-gritty for a long time.

vjr:
@jeremypw I've added a (WIP) custom `copy_move_with_sync ()` intended to be a drop-in replacement for the `g_file_copy ()` and `g_file_move ()` glib/gio methods.

Yes, the periodic `g_fsync ()` calls seem to be slowing down the transfers compared to the `main` branch so I'll boot up live sessions of ubuntu and kubuntu to compare against gnome and kde file managers, see what they do and how fast or slow they are.

I just tested individual file transfers of various sizes on two usb storage I have, a fast ssd and a slow flash stick, also tried multiple small files.

At the moment this new approach needs work to fix crashes/errors trying to copy directories, haven't yet tried other copy/move styles, folders, overwrite, recurse whatever.

Will continue working on this, see if the original goal of the issue (avoid users inadvertently unplugging their usb devices) can be resolved.

Thanks for looking into this.

flodavid:
@jeremypw here it is: #2821. I assigned you to the issue.

jeremypw:
@vjr Thanks for working on this!  At the end of the day a (modest) reduction in speed is worth it to fix serious issues like crashes and data loss.  I'll have another look at this tomorrow.

vjr:
Tried the gnome and kde file managers - they for some reason seem to be slower than this branch with my fast usb drive - it's rated for 1 GBps and this branch shows about 700 to 900 MBps read or write for large files to/from internal SSD. My slow usb stick shows the same transfer speeds on all file managers - about 7 MBps write and 20 MBps read. So this PR looks pretty good to me. I tested various scenarios/cases - copying/moving files/folders to/from internal/external disks - seems ok to me but I'm looking forward to others testing and any code review/changes if any. Thanks again!

PS: if i click eject the usb in gnome file manager it shows a notification "do not unplug device" - then after the transfer has completed (with NO indication of it being in progress) - it shows another notification "you can unplug the device now"... this is not very helpful with my very slow 8 GB usb stick which can maybe a full 15 minutes to complete the read (transfer out) and a user can still forget about the notification and unplug mid transfer!

jeremypw:
@vjr This is indeed looking good from a functional point of view.  The only slight issue I found (and it is not a regression on `main` anyway) is that my old MTP device , which plugs in a USB port, is not detected as unpluggable.  Not sure what can be done about that other than assuming all devices with that protocol are unpluggable.

Code wise, my C skills are rudimentary at best and rusty so probably the code should be reviewed by another.  I am not sure how easy it would be to implement the new function in Vala? Some of the file operation stuff has already been ported by @tintou and it would be nice to make some more progress.  

When transferring small files I am not sure it is needed to fsync twice for every file but it does not seem to slow things down significantly so perhaps the system throttles fsync calls?

vjr:
OK I'll see if I can port this PR to Vala. I suspect the MTP device does not show as unpluggable but the PR triggers syncing (this part of the code does detect it as unpluggable) is because I've added 3 conditions if any of them true, I'll see about DRY-ing that part so that your MTP device also shows up as ejecctable. Working on it.

jeremypw:
@vjr Thanks!  Sorry to make extra work but we might as well continue to port to Vala when adding new code at least.  I look forward to testing the new version; it looks like a big improvement.

vjr:
Marking back to draft status while I see if I can port this PR to Vala.

vjr:
@jeremypw while i'm working on porting this PR to vala could you take a quick look at https://github.com/elementary/files/pull/2819/commits/a737e58ab4c67e09e8123d47010d932917e2bdeb to see if it now marks your old MTP drive as `can_umount` or `can_eject` and actually does so when you click the unmount/eject icon that hopefully shows up?

jeremypw:
@vjr Your `mount_can_unplug_drive` function returns `false` for the old MTP drive.  This is because there is no drive detected.  In the sidebar it just appears as a mount (i.e. there is an unmount button but the context menu has no "Eject " or "Safely Remove" option).  I wouldn't worry too much about this tbh.  It is probably more an issue with `gvfs`.  Also I tried  forcing it to use sync and it crashed because it was unable to create a stream (your code should guard against that possibility by the way)

jeremypw:
I guess you could just assume that anything other than a local file system could, in principle, be disconnected at any time.

vjr:
Only quickly tested for copying large/small files to/from fast/slow usb drives, it works but some work left to do, will return tomorrow. Also overwriting fails right now, will look into it. Port looks promising!

vjr:
@jeremypw I've marked it ready for (your) review. Can we tackle the MTP not being detected as unpluggable issue in a later PR/branch? I have another idea to fix that, but I'd rather you try making changes since you can test quicker. It may just need adding the following somewhere in either the `file_can_unplug_drive ()` or `mount_can_unplug_drive ()` new util methods before checking `drive == null` or maybe inside the check before returning false:

```
if (mount.can_unmount () || mount.can_eject ()) {
    return true;
}
```

vjr:
@jeremypw what is the filesystem on your MTP device? Maybe that's the issue?

vjr:
@jeremypw another thing to try if you would look for missing packages with "mtp" in the name such as libmtp-xxx and mtp-tools and sudo apt install them?

jeremypw:
> what is the filesystem on your MTP device? Maybe that's the issue?

The filesystem is `mtpfs`.  To be honest I am not too bothered about it - I only use it for testing these days and I can move files back and forth OK.

jeremypw:
@vjr This seems to be working well afaict! The progressbar now stays open until the transfer onto a unpluggable device completes and the progress bar actually reaches 100%.

There are code style issues but I don't mind fixing those after merging if you are not familiar with elementary code styling.  

In the longer run I was envisaging this method being an instance method of CopyMoveJob and being a Vala async method perhaps but for an initial port its best to stick close to the original C code as you have done.

There are only a couple of (old) issues that mention MTP related to connecting mobile phones and I think these are more related to permissions than Files itself. Modern phones make direct USB connections difficullt anyway and most people make transfers another way so I don't think this is a priority.

I raised a couple of unrelated issues I noticed while testing this PR which I will address at some point.

jeremypw:
If you are interested in doing more work on file operations there is a long outstanding issue #137 which it would be good to fix!  No pressure though.

jeremypw:
One thing I am not too sure about: what happens when an error is thrown e.g. at line 127 in CopyMoveJob.vala? Are streams left open?  Would it be better to handle errors internally and have `finally` clauses to ensure cleanup? We can always throw our own error afterwards.

vjr:
> One thing I am not too sure about: what happens when an error is thrown e.g. at line 127 in CopyMoveJob.vala? Are streams left open? Would it be better to handle errors internally and have `finally` clauses to ensure cleanup? We can always throw our own error afterwards.

IINM (might have read this on the vala.dev website) it auto closes the stream when going out of scope aka final ref released, regardless of success or failure or error thrown/returned?

vjr:
> If you are interested in doing more work on file operations there is a long outstanding issue #137 which it would be good to fix! No pressure though.

Yep I think I can give this a shot, see if I can fix.

vjr:
> @vjr This seems to be working well afaict! The progressbar now stays open until the transfer onto a unpluggable device completes and the progress bar actually reaches 100%.
> 
> There are code style issues but I don't mind fixing those after merging if you are not familiar with elementary code styling.
> 
> In the longer run I was envisaging this method being an instance method of CopyMoveJob and being a Vala async method perhaps but for an initial port its best to stick close to the original C code as you have done.
> 
> There are only a couple of (old) issues that mention MTP related to connecting mobile phones and I think these are more related to permissions than Files itself. Modern phones make direct USB connections difficullt anyway and most people make transfers another way so I don't think this is a priority.
> 
> I raised a couple of unrelated issues I noticed while testing this PR which I will address at some point.

If you point out the code style issues (hopefully not too many) I'm happy to address in this PR itself, is it the initial new method return value/signature being on separate line? Also, maybe the first few lines where I check `do_sync` then the `is_move ? return...` the way they are on separate lines? I assumed since the linter was passing I was good hah.

If you don't mind retesting I would like to fix/address the MTP issue as I mentioned in https://github.com/elementary/files/pull/2819#issuecomment-5536226799 I think that might work.

Other than that, please point out any remaining issues in this PR that I can address to get it approved and merged... I'm really motivated to have this fix done!

Thanks!

vjr:
Marking as draft while I push some more pending commits and address any review feedback...

jeremypw:
@vjr fyi elementary code style info here https://docs.elementary.io/develop/writing-apps/code-style. I'll go through and point out any remaining issues later today.  Note that old code (particularly C) does not necessary fully comply but new code should.

vjr:
@jeremypw apologies for the repeated drafting/undrafting but could you take another look at the code and maybe redo a few of your tests, especially with your MTP device? thank you!

jeremypw:
A convention I use (which I do not think is mentioned in the guide) is to use only one indent for the parameters of functions if there are too many to fit on one line and all parameters are indented the same.  Closing bracket groups should be on the next line:

``` 
private void some_function (
    param1,
    param2,
    ....
) {
}

```

The same applies to function calls that do not fit within the 120 char line liimit except one more indent that the current line indent is used.

The scope and type of functions etc are always inline with the name.

In general I do not like using whitespace between lines that are logically related e.g. initializing parameters for the same clause.

But whitespace is put after clauses ending in `}` but not necessarily after statements ending in `};`

vjr:
> A convention I use (which I do not think is mentioned in the guide) is to use only one indent for the parameters of functions if there are too many to fit on one line and all parameters are indented the same. Closing bracket groups should be on the next line:
> 
> ```
> private void some_function (
>     param1,
>     param2,
>     ....
> ) {
> }
> ```
> 
> The same applies to function calls that do not fit within the 120 char line liimit except one more indent that the current line indent is used.
> 
> The scope and type of functions etc are always inline with the name.
> 
> In general I do not like using whitespace between lines that are logically related e.g. initializing parameters for the same clause.
> 
> But whitespace is put after clauses ending in `}` but not necessarily after statements ending in `};`

@jeremypw i've hopefully understood and properly applied your advice in https://github.com/elementary/files/pull/2819/commits/cc5418086866e22ef70c2eb4d29ae9305d85d353 ?

libcore/FileOperations/CopyMoveJob.vala
```diff
@@ -43,6 +45,120 @@ public class Files.FileOperations.CopyMoveJob : CommonJob {
...
+    ) throws Error {
+        if (!do_syncs) {
+            if (is_move) {
+                return src.move (dest, flags, cancellable, progress_callback);
+            } else {
+                return src.copy (dest, flags, cancellable, progress_callback);
+            }
+        }
+
+        var overwrite = (flags & FileCopyFlags.OVERWRITE) != 0;
```
jeremypw: This variable is not use 'til later so can be moved to before the clause in which it is used, and whitespace reduced to a single line.

libcore/FileOperations/CopyMoveJob.vala
```diff
@@ -43,6 +45,120 @@ public class Files.FileOperations.CopyMoveJob : CommonJob {
...
+        if (is_move) {
+            var atomic_move_success = false;
+            try {
+                atomic_move_success = src.move (
+                    dest,
+                    flags | FileCopyFlags.NO_FALLBACK_FOR_MOVE,
+                    cancellable,
+                    progress_callback
+                );
+            } finally {
```
jeremypw: Shouldn't need `finally` here as we did not setup anything that needs to be undone at this point.

libcore/FileOperations/CopyMoveJob.vala
```diff
@@ -43,6 +45,120 @@ public class Files.FileOperations.CopyMoveJob : CommonJob {
...
+
+        var src_is_dir = FileUtils.file_is_dir (src);
+        var dest_is_dir = FileUtils.file_is_dir (dest);
+        var dest_exists = dest.query_exists ();
+
+        var error = -1;
+
+        if (src_is_dir) {
+            if (dest_is_dir) {
+                error = overwrite ? IOError.WOULD_MERGE : IOError.EXISTS;
```
jeremypw, line 89: Can't we throw an error straightaway here and in similar other places in this clause?  You can use the simpler form:

`throw new IO.Error.EXISTS ("Put a suitable error message here if needed");`

for example.

libcore/FileOperations/CopyMoveJob.vala
```diff
@@ -43,6 +45,120 @@ public class Files.FileOperations.CopyMoveJob : CommonJob {
...
+            }
+        } else {
+            if (dest_is_dir) {
+                error = overwrite ? IOError.IS_DIRECTORY : IOError.EXISTS;
+            } else if (!overwrite && dest_exists) {
+                error = IOError.EXISTS;
+            }
+        }
+
+        if (error >= 0) {
```
jeremypw, line 103: If errors are thrown straightaway this clauses would not be needed.

libcore/FileOperations/CopyMoveJob.vala
```diff
@@ -43,6 +45,120 @@ public class Files.FileOperations.CopyMoveJob : CommonJob {
...
+            if (atomic_move_success) {
+                return true;
+            }
+        }
+
+        var src_is_dir = FileUtils.file_is_dir (src);
+        var dest_is_dir = FileUtils.file_is_dir (dest);
+        var dest_exists = dest.query_exists ();
+
+        var error = -1;
```
jeremypw, line 85: If errors are thrown straightaway this variable would not be needed here.

libcore/FileOperations/CopyMoveJob.vala
```diff
@@ -43,6 +45,120 @@ public class Files.FileOperations.CopyMoveJob : CommonJob {
...
+        var info = in.query_info (FileAttribute.STANDARD_SIZE, cancellable);
+        var total_size = info.get_size ();
+        var out = dest.replace (null, false, FileCreateFlags.NONE, cancellable);
+
+        var fd = -1;
+        if (out is FileDescriptorBased) {
+            fd = out.get_fd ();
+        }
+
+        var success = false;
```
jeremypw, line 117: Whitespace between here and the related `while` clause can be removed

libcore/FileOperations/CopyMoveJob.vala
```diff
@@ -43,6 +45,120 @@ public class Files.FileOperations.CopyMoveJob : CommonJob {
...
+
+        uint8 buffer[COPY_MOVE_CHUNK_SIZE];
+        size_t copied = 0;
+
+        int64 last_sync_time = 0;
+
+        while (true) {
+            var read = in.read (buffer, cancellable);
```
jeremypw, line 123: This can throw and error and immediately return without closing streams.  Do we need to put this in a try/catch/finally clause to ensure streams are closed?

libcore/FileOperations/CopyMoveJob.vala
```diff
@@ -43,6 +45,120 @@ public class Files.FileOperations.CopyMoveJob : CommonJob {
...
+
+            size_t written = 0;
+            out.write_all (buffer[0:read], out written, cancellable);
+
+            copied += written;
+
+            var now = get_monotonic_time ();
+            if (last_sync_time == 0 || (now - last_sync_time).abs () >= SYNC_INTERVAL_MICROS) {
+                if (fd >= 0) {
+                    Posix.fsync (fd);
```
jeremypw, line 137: Have you considered using `flush` method of `GLib.OutputStream` instead of the Posix function.  We would then not need `fd`. I am not sure whether they are equivalent but its worth a try.

libcore/FileOperations/CopyMoveJob.vala
```diff
@@ -43,6 +45,120 @@ public class Files.FileOperations.CopyMoveJob : CommonJob {
...
+                }
+                last_sync_time = now;
+            }
+
+            if (progress_callback != null) {
+                progress_callback (copied, total_size);
+            }
+        }
+
+        if (fd >= 0) {
```
jeremypw, line 147: Closing the stream will flush its data.

Apparently streams are automatically closed when the last reference is dropped but it might be better to close them explicitly and handle any errors.

jeremypw (changes requested):
With an unpluggable MTP device, it is identified as unpluggable but fails to copy or move files onto it.  This is because creating the output stream fails and the code does not handle that - it should fall back to a "normal" copy/move.

vjr:
@jeremypw hmm i wonder if https://github.com/elementary/files/pull/2828 might be a better (way better) approach to fix the issue? Thought to attempt this while working on https://github.com/elementary/files/pull/2827 to fix #137

jeremypw:
@vjr May as well close this as the simpler alternative was merged?

PR #2823, opened 2026-08-31
Fix crash on unmount when busy with other processes

### LINKED ISSUE #2821: App crashes when trying to unmount a volume being used by another app
### What Happened?

The app crashes when I try to unmount a busy volume. Comparing with my own app (Espace Libre) which shows a warning.
Here Byte (with PID 9254) is playing music from BIG :

https://github.com/user-attachments/assets/a54d735e-34d4-4a1b-b998-78a94175b39c

For information, here is the code I use to try to unmount the volume, which uses the `GLib.Mount` class, automatically showing the warning dialog : 
```vala
var mount_op = new Gtk.MountOperation (null);
try {
    yield mount.unmount_with_operation (
            GLib.MountUnmountFlags.NONE,
            mount_op,
            null
    );
    return true;
} catch (GLib.Error e) {
    warning ("Unable to unmount '%s': %s", mount.get_name (), e.message);
    return false;
}
```

### Steps to Reproduce

1. Make a volume busy. For example, play music from the volume using the Byte music player
2. Try to unmount the volume
3. Enter user password
4. App crashes

### Expected Behavior

A warning saying that the volume is busing after entering the password. Or even better, do not ask password and show the warning immediately when trying to unmount the busy volume.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

Nvidia RTX 2060, Intel i5-9400F

jeremypw:
Fixes #2821 

 - Use a custom MountOperation
 - Show our own dialog for showing other processes
 - Always abort for now

For simplicity a Granite.MessageDialog is used for now.
Screenshot from 2026-08-31 18 03 13

A problem is that if the competing process is FlatPak'd which is likely, the name of the competing process may well be that of a portal, not the original process. Not sure how to get round that.

In principle we could offer additional choices such forcing the unmount or ending the competing processes but that is left for design input.

### DIFF
--- a/libcore/FileOperations/MountUtils.vala
+++ b/libcore/FileOperations/MountUtils.vala
@@ -273,7 +273,7 @@ namespace Files.FileOperations {
             primary_text = _("The resource '%s' is in use by other processes").printf (mount_name);
             secondary_text = _("Unmounting now might cause a process to fail or to lose data");
             var sb = new StringBuilder ("");
-            sb.append (_("Other processes using '%s' ...\n").printf (mount_name));
+            sb.append (_("Other processes using '%s'… \n").printf (mount_name));
             foreach (var pid in processes) {
                 sb.append (get_process_name_from_pid (pid));
                 sb.append ("\n");
@@ -284,7 +284,7 @@ namespace Files.FileOperations {
         }
 
         private string? get_process_name_from_pid (int pid) {
-            string process_path = "/proc/%d/exe".printf(pid);
+            string process_path = "/proc/%d/exe".printf (pid);
 
             try {
                 string path = GLib.FileUtils.read_link (process_path);
--- a/libcore/FileOperations/MountUtils.vala
+++ b/libcore/FileOperations/MountUtils.vala
@@ -274,11 +274,22 @@ namespace Files.FileOperations {
             secondary_text = _("Unmounting now might cause a process to fail or to lose data");
             var sb = new StringBuilder ("");
             foreach (var pid in processes) {
-                sb.append (pid.to_string ());
+                sb.append (get_process_name_from_pid (pid));
                 sb.append ("\n");
             }
             show_error_details (sb.str);
             show_all ();
+        }
+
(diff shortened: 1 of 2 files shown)


vjr (approved):
Works for me in a couple of tests except when using the Code app.

Fixes (prevents) crash-on-unmount while transferring files and while editing a file in VIM but there's maybe something in the way the Code app works I can still unmount and remount while editing a file in Code but that seems to be an issue in Code not this fix.

This can be merged for sure, any further fixes if required can be left for future PRs.

jeremypw:
@vjr I think Code closes the document file after every update (and keeps a copy of its content in the text buffer) so mostly does not appear to be using it.  Maybe we should mark open document files as in use?

vjr:
> @vjr I think Code closes the document file after every update (and keeps a copy of its content in the text buffer) so mostly does not appear to be using it. Maybe we should mark open document files as in use?

Where does this "mark open files as in use" change need to happen? Would you mind filing an issue ticket in the proper repo? I'm not sure whether it's supposed to be (likely?) Code or (unlikely?) Files.

jeremypw:
@vjr At the moment, Code handles the situation of unmounting the source of a document  while there are unsaved changes by showing a dialog:

Screenshot from 2026-09-03 17 16 57

This gives the user the chance to remount the source or save changes elsewhere. 

If there are no unsaved changes, the document just closes.

 I presume that if the unmount was requested *during* a lengthy save from Code (normally documents are not large enough to take a significant time to save) then a "busy" dialog would be shown (after this PR).  I'll check what other editors do.

jeremypw:
@vjr I checked Gnome Text Editor (FlatPak as native install gave weird behaviour on Wayland) and this did not inhibit unmounting a USB stick containing the file being edited.  The dialog when saving after that was less helpful than Code - it simply gave an error that the file did not exist.

jeremypw:
Nevertheless I'll raise an issue to see what comments arise.

PR #2826, opened 2026-09-04
Some unmount operation busy-dialog fixes

vjr:
Following changes proposed:

- Automatically close the busy warning dialog when the unmount operation finishes. Previously the dialog would not close when the cancel button was clicked although pressing ESC worked.
- Special-case to show different dialog messages if the only process is ourself.
- Exclude ourself from the list of other processes when shown.

### DIFF
--- a/libcore/FileOperations/CommonJob.vala
+++ b/libcore/FileOperations/CommonJob.vala
@@ -48,6 +48,8 @@ public class Files.FileOperations.CommonJob {
     protected unowned Gtk.Window? parent_window;
     protected uint inhibit_cookie;
     protected unowned GLib.Cancellable? cancellable;
+    public unowned GLib.File? src_mount = null;
+    public unowned GLib.File? dest_mount = null;
     protected PF.Progress.Info progress;
     protected Files.UndoActionData? undo_redo_data;
     protected GLib.Timer time;
@@ -59,6 +61,7 @@ public class Files.FileOperations.CommonJob {
         inhibit_cookie = 0;
         progress = new PF.Progress.Info ();
         cancellable = progress.cancellable;
+        cancellable.cancelled.connect (cancel_mount_operation);
         undo_redo_data = null;
         time = new GLib.Timer ();
     }
@@ -68,6 +71,28 @@ public class Files.FileOperations.CommonJob {
         uninhibit_power_manager ();
         if (undo_redo_data != null) {
             Files.UndoManager.instance ().add_action ((owned) undo_redo_data);
+        }
+    }
+
+    private void cancel_mount_operation () {
+        try {
+            if (src_mount != null) {
+                Mount mount = src_mount.find_enclosing_mount (null);
+                UnmountOperation.cancel_mount_operation (mount);
+            }
+        } catch (Error e) {
+        } finally {
+            src_mount = null;
+        }
+
+        try {
+            if (dest_mount != null) {
+                Mount mount = dest_mount.find_enclosing_mount (null);
+                UnmountOperation.cancel_mount_operation (mount);
+            }
+        } catch (Error e) {
+        } finally {
+            dest_mount = null;
         }
     }
 
--- a/libcore/FileOperations/MountUtils.vala
+++ b/libcore/FileOperations/MountUtils.vala
@@ -279,7 +279,7 @@ namespace Files.FileOperations {
             Pid self = Posix.getpid ();
             if (processes.length == 1 && processes.index (0) == self) {
                 primary_text = _("The resource '%s' is in use").printf (mount_name);
-                secondary_text = _("Please wait for it to unmount or you may cancel it");
+                secondary_text = _("Please wait or cancel unmounting it");
             } else {
                 primary_text = _("The resource '%s' is in use by other processes").printf (mount_name);
                 secondary_text = _("Unmounting now might cause a process to fail or to lose data");
--- a/libcore/FileOperations/MountUtils.vala
+++ b/libcore/FileOperations/MountUtils.vala
@@ -235,6 +235,11 @@ namespace Files.FileOperations {
             );
         }
 
+        ~UnmountOperation () {
+            dialog.close ();
+            dialog.destroy ();
+        }
+
         public override void show_processes (string message, Array<Pid> processes, string[] choices) {
             if (dialog != null) {
                 return;
@@ -242,6 +247,7 @@ namespace Files.FileOperations {
 
             dialog = new BusyDialog (mount_name, processes);
             dialog.response.connect (() => {
+                dialog.close ();
                 dialog.destroy ();
                 dialog = null;
                 reply (MountOperationResult.ABORTED); // Results in IOError.FAILED_HANDLED
--- a/libcore/FileOperations/MountUtils.vala
+++ b/libcore/FileOperations/MountUtils.vala
@@ -241,6 +241,11 @@ namespace Files.FileOperations {
         }
 
         public override void show_processes (string message, Array<Pid> processes, string[] choices) {
+            Pid self = Posix.getpid ();
+            if (processes.length == 1 && processes.index (0) == self) {
+                return;
+            }
+
             if (dialog != null) {
                 return;
             }
--- a/libcore/FileOperations/MountUtils.vala
+++ b/libcore/FileOperations/MountUtils.vala
@@ -286,6 +286,9 @@ namespace Files.FileOperations {
                 var sb = new StringBuilder ("");
                 sb.append (_("Other processes using '%s'… \n").printf (mount_name));
                 foreach (var pid in processes) {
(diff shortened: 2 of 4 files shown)


vjr:
This is a follow-up to https://github.com/elementary/files/pull/2823 to propose some fixes and usability improvements.

vjr:
@jeremypw another one of my drive-by proposed contributions - hopefully this is useful.

jeremypw (commented):
Thanks for this - I agree with the intention.  Just a couple of suggestions for further clarification.

 - Label the "Cancel" button "Do not unmount" or something similar to clarify the exact consequence of pressing it.
 - When it is Files only using the resource, we could give more information regarding the file operation - or better, disable the sidebar eject button altogether?
 - Expand "Please wait" to clarify what it is the user is waiting for.  e.g. "<resource> will unmount when the operation finishes."

vjr:
> Thanks for this - I agree with the intention. Just a couple of suggestions for further clarification.
> 
> * Label the "Cancel" button "Do not unmount" or something similar to clarify the exact consequence of pressing it.
> * When it is Files only using the resource, we could give more information regarding the file operation - or better, disable the sidebar eject button altogether?
> * Expand "Please wait" to clarify what it is the user is waiting for.  e.g. " will unmount when the operation finishes."

I've updated the text in https://github.com/elementary/files/pull/2826/commits/6ee3d9f50c10acab2ec9f26efae0caed52bef475 how does it look?

I don't know how to set custom text for the CANCEL button - how do I do that?

Not sure how to do item 2 - when Files is the only app using it - how to detect this without the `UnmountOperation` - or gather more info about whats being done?

jeremypw:
> I don't know how to set custom text for the CANCEL button - how do I do that?

Use `ButtonsType.NONE` in the dialog new function and then use `add_button` in the construct clause.  It doesn't matter what response id you choose since this is ignored in this case.


> I've updated the text, how does it look?

Not sure whether the balance between brevity and exactitude is right yet.  Here is something I mocked up:

Screenshot from 2026-09-08 11 37 27

This probably goes too far into verbosity but it is up to the user to decide whether to wait or cancel unmounting.

I found an issue if the operation in question is a Files transfer.  If the transfer is cancelled while the dialog is open, then the dialog does not close and, if cancelled, further attempts to unmount throw the dialog again even though there are no Files operations in progress.   This is probably an issue with Files operations code rather than this PR.

> Not sure how to do item 2 - when Files is the only app using it - how to detect this without the UnmountOperation - or gather more info about whats being done?

I'll look into it.

vjr:
> I found an issue if the operation in question is a Files transfer. If the transfer is cancelled while the dialog is open, then the dialog does not close and, if cancelled, further attempts to unmount throw the dialog again even though there are no Files operations in progress. This is probably an issue with Files operations code rather than this PR.

This should be fixed by https://github.com/elementary/files/pull/2826/commits/d3abac0dc5a7a8dae39c6155ffa2306a81354aa7

Please re-test the overall PR when you can, thanks!

libcore/FileOperations/CommonJob.vala
```diff
@@ -48,6 +48,8 @@ public class Files.FileOperations.CommonJob {
     protected unowned Gtk.Window? parent_window;
     protected uint inhibit_cookie;
     protected unowned GLib.Cancellable? cancellable;
+    public unowned GLib.File? src_mount = null;
+    public unowned GLib.File? dest_mount = null;
```
jeremypw, line 52: A little confusing that these a called `*_mount` but are not Mount objects. Could these not be called `*_folder` perhaps.

jeremypw:
@vjr 
> Not sure how to do item 2 - when Files is the only app using it - how to detect this without the UnmountOperation - or gather more info about whats being done?

In principle some information can be obtained from the ProgressInfo objects listed in ProgressInfoManager, however at the moment I think some changes would be required to get this to work which are outside the scope of this PR.  

I am drafting some more ported Vala code which includes a FileOperationsManager that maintains a list of ongoing jobs. If this is made available to the UI then it should be straightforward to display the required information.

jeremypw:
> This should be fixed by https://github.com/elementary/files/commit/d3abac0dc5a7a8dae39c6155ffa2306a81354aa7

This commit does fix the issue - thanks! The only comment I would make is that if the ongoing Files operation is cancelled then the dialog closes without unmounting the mount which is in contrast to what happens if the operation is allowed to proceed to the end and contradicts the text in the dialog somewhat.

libcore/FileOperations/CommonJob.vala
vjr: ah yes, maybe i'll rename to something like `unmount_op_src` and `unmount_op_dest` ...

vjr:
> > This should be fixed by [d3abac0](https://github.com/elementary/files/commit/d3abac0dc5a7a8dae39c6155ffa2306a81354aa7)
> 
> This commit does fix the issue - thanks! The only comment I would make is that if the ongoing Files operation is cancelled then the dialog closes without unmounting the mount which is in contrast to what happens if the operation is allowed to proceed to the end and contradicts the text in the dialog somewhat.

Ah yes, my brain fart, should NOT cancel the unmount when the file op is cancelled, it should proceed to unmount, i'll see about fixing, but yes, probably best to wait on your porting/file-operations PR(s) to get merged first, then can revisit my PRs one at a time.

vjr:
> @vjr
> 
> > Not sure how to do item 2 - when Files is the only app using it - how to detect this without the UnmountOperation - or gather more info about whats being done?
> 
> In principle some information can be obtained from the ProgressInfo objects listed in ProgressInfoManager, however at the moment I think some changes would be required to get this to work which are outside the scope of this PR.
> 
> I am drafting some more ported Vala code which includes a FileOperationsManager that maintains a list of ongoing jobs. If this is made available to the UI then it should be straightforward to display the required information.

@jeremypw sounds good, i think it's best to wait for your new PR(s) to get merged so i can revisit mine to update them to take advantage of the ported code and the new file operations manager.

jeremypw:
@vjr It should be OK to finish and merge this PR before my work - it doesn't touch the same code.  The thing about showing more info about Files operations is not essential and can be left for future PR.

vjr:
> I found an issue if the operation in question is a Files transfer. If the transfer is cancelled while the dialog is open, then the dialog does not close and, if cancelled, further attempts to unmount throw the dialog again even though there are no Files operations in progress. This is probably an issue with Files operations code rather than this PR.

I've reverted https://github.com/elementary/files/commit/d3abac0dc5a7a8dae39c6155ffa2306a81354aa7 in https://github.com/elementary/files/pull/2826/commits/7f276ef3b89aecf89f4b9e2817778b4fd3e91f0e because this issue you observed seems to be just appearances due to the way kernel buffer cache is happening. Meaning, if you cancel a transfer, the unmount dialog does not close because the transfer dialog disappears but under the hood the cancel is still in progress, and the unmount dialog will in fact close if you wait some time. Could you try that?

Other scenario you mentioned, if you cancel the transfer and then immediately cancel the unmount, the transfer cancel is still in progress, so when you retry unmount, the busy dialog reappears, but it will eventually close and the unmount will happen when the underlying kernel buffer cache completes.

Could you re-try and see if what I've observed is also happening at your end?

libcore/FileOperations/CommonJob.vala
vjr: I've reverted this commit anyway: https://github.com/elementary/files/pull/2826#issuecomment-5653065368

vjr:
If this PR is ready (enough) to approve/merge we might just go ahead with it and see if the other earlier PR(s) to sync and the progress dialog fixes will help to make this issue clearer in that the cancelled transfer dialog stays open until actually finished the cancel.

jeremypw:
@vjr I noticed you were translating a constant and a verbatim string which I do not think works properly.  Also a missing blank line. I have corrected this and marking as approved.  Please check my corrections before merging.

jeremypw (approved):
Looks good to go.

vjr:
@jeremypw the whitespace commit seems to add it under a `foreach` instead of an `if` block? Not a big deal either way.

Also, I'm very unfamiliar with translation concepts (the vala/gtk/glib APIs anyway) so I'll take your commit at face value there :-)

So going ahead and merging this, looking forward to your WIP on the porting/FileOperationsManager code changes before I revisit my other PRs for the syncing and progress bar dialog fixes, thanks a lot for your patience with my work!

jeremypw:
> the whitespace commit seems to add it under a `foreach` instead of an `if` block?

No, the foreach block is inside an if - else block? The guidance is a bit inconsistent anyway - I would prefer there to be white space under all blocks that are not statements (i.e. end in `}` not `;`).

jeremypw:
Just noticed another one missing inside the `foreach`  block but no matter :shrug:

PR #2827, opened 2026-09-04
Fix progress bar dialog for trash, restore and empty trash file operations

vjr:
Fixes #137 

- Sync files for trash, restore from trash and empty trash.
- Trash files progress dialog working as before.
- Restore from trash progress dialog now shows up.

PR #2828, opened 2026-09-06
Sync transfers (redux) for realistic progress bar updates

### LINKED ISSUE #2818: Transfer progress bar finishes too fast

vjr:
Fixes #2818 

Alternative (simpler) approach to #2819

### DIFF
--- a/libcore/FileOperations/CopyMoveJob.vala
+++ b/libcore/FileOperations/CopyMoveJob.vala
@@ -102,6 +102,7 @@ public class Files.FileOperations.CopyMoveJob : CommonJob {
         int64 now = GLib.get_monotonic_time () * 1000; // in ns
 
         if (transfer_info.last_report_time != 0 &&
+            (source_info.num_files - transfer_info.num_files) > 1 &&
             ((int64)transfer_info.last_report_time - now).abs () < 100 * CommonJob.NSEC_PER_MSEC) {
             return;
         }
@@ -199,6 +200,7 @@ public class Files.FileOperations.CopyMoveJob : CommonJob {
         }
 
         var total_size = int64.max (source_info.num_bytes, transfer_info.num_bytes);
+        var size_left = total_size - transfer_info.num_bytes;
 
         double elapsed = time.elapsed ();
         double transfer_rate = 0;
@@ -207,11 +209,18 @@ public class Files.FileOperations.CopyMoveJob : CommonJob {
         }
 
         if (elapsed < CommonJob.SECONDS_NEEDED_FOR_RELIABLE_TRANSFER_RATE &&
+            size_left > 0 &&
             transfer_rate > 0) {
             var num_bytes_format = GLib.format_size (transfer_info.num_bytes);
             var total_size_format = GLib.format_size (total_size);
             /// TRANSLATORS: %s is a placeholder for a size like "2 bytes" or "3 MB".  It must not be translated or removed. So this represents something like "4 kb of 4 MB".
             progress.take_details (_("%s of %s").printf (num_bytes_format, total_size_format));
+        } else if (size_left == 0) {
+            progress.take_details (
+                is_move ?
+                _("Please wait, finishing move\u2026") :
+                _("Please wait, finishing copy\u2026")
+            );
         } else {
             var num_bytes_format = GLib.format_size (transfer_info.num_bytes);
             var total_size_format = GLib.format_size (total_size);
--- a/libcore/FileUtils.vala
+++ b/libcore/FileUtils.vala
@@ -433,6 +433,22 @@ namespace Files.FileUtils {
         return new_path;
     }
 
+    public void sync (GLib.File? file) {
+        if (file == null) {
+            return;
+        }
+        var path = file.get_path ();
+        if (path == null) {
+            return;
+        }
+        int fd = Posix.open (path, Posix.O_RDONLY);
+        if (fd < 0) {
+            return;
+        }
+        Posix.fsync (fd);
+        Posix.close (fd);
+    }
+
     /** Splits the path into a protocol ending in '://"  and a path beginning "/". **/
     public void split_protocol_from_path (string path, out string protocol, out string new_path) {
         protocol = "";
--- a/libcore/marlin-file-operations.c
+++ b/libcore/marlin-file-operations.c
@@ -1409,6 +1409,15 @@ copy_file_progress_callback (goffset current_num_bytes,
     }
 }
 
+static void
+sync_file_callback (
+    goffset current_num_bytes,
+    goffset total_num_bytes,
+    gpointer file_to_sync
+) {
+    files_file_utils_sync (*(GFile **)file_to_sync);
+}
+
 static gboolean
 test_dir_is_parent (GFile *child, GFile *root)
 {
@@ -2230,9 +2239,11 @@ retry:
     if (g_file_move (src, dest,
                      flags,
                      job->cancellable,
-                     NULL,
-                     NULL,
+                     sync_file_callback,
+                     &dest,
                      &error)) {
+
+        files_file_utils_sync (dest);
 
         if (debuting_files) {
             g_hash_table_replace (debuting_files, g_object_ref (dest), GINT_TO_POINTER (TRUE));
@@ -3165,10 +3176,12 @@ retry:
                                dest,
                                G_FILE_COPY_NONE,
                                common->cancellable,
-                               NULL, NULL,
+                               sync_file_callback,
+                               &dest,
                                &error);
             // Start UNDO-REDO
             if (res) {
+                files_file_utils_sync (dest);
                 files_undo_action_data_set_create_data(common->undo_redo_data,
                                                          g_file_get_uri(dest),
                                                          g_file_get_uri(job->src));
--- a/libcore/marlin-file-operations.c
+++ b/libcore/marlin-file-operations.c
@@ -41,8 +41,6 @@
 
 #include "pantheon-files-core.h"
 
-#define SECONDS_NEEDED_FOR_RELIABLE_TRANSFER_RATE 15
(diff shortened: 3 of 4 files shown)


vjr:
@jeremypw hmm i wonder if this might be a better (way better) approach to fix the issue? Thought to attempt this while working on https://github.com/elementary/files/pull/2827 to fix #137

libcore/marlin-file-operations.c
```diff
@@ -2227,10 +2239,12 @@ move_file_prepare (FilesFileOperationsCopyMoveJob *move_job,
     if (g_file_move (src, dest,
                      flags,
                      job->cancellable,
-                     NULL,
-                     NULL,
+                     sync_file_callback,
+                     &dest,
                      &error)) {
 
+        files_file_utils_sync (dest);
```
jeremypw: This is not required as sync_file_callback is guaranteed to be called by g_file_move when data transfer is complete.

vjr: Done -> https://github.com/elementary/files/pull/2828/commits/5826976b2c5057dec55945901d7fa1daf2caf7fd

jeremypw (commented):
Certainly simpler! But as far as I can see , when moving files your sync function is only called for local moves not for moves e.g. onto USB stick which is the opposite of what is needed?  

I must admit it is hard to get one's head around all the C code 😞 

As far as I can see `move_files_prepare` actually moves files in some cases (g_file_move succeeds without fallback flag) so the name is misleading.  When this fails it calls `copy_move_files` and eventually `g_file_move` again with different flags allowing fallback.  In the latter case the callback is still `copy_file_progress_callback`.

jeremypw:
It would be nice to get the progress window to update in the *middle* of transferring a large file.  At the moment it just freezes.

vjr:
> It would be nice to get the progress window to update in the _middle_ of transferring a large file. At the moment it just freezes.

This appears to be the `g_file_copy ()` invoking the callbacks very quickly likely due to the kernel buffer cache? So the "report copy move progress" function returns early because 100ms has not elapsed yet.

Should be addressed in https://github.com/elementary/files/pull/2828/commits/f2a101752cf3fc420ac35bddd48864bb892f695d

It always reports the final callback/progress invocation and added a special case progress details text like "Please wait, finishing copy..."

vjr:
> Certainly simpler! But as far as I can see , when moving files your sync function is only called for local moves not for moves e.g. onto USB stick which is the opposite of what is needed?
> 
> I must admit it is hard to get one's head around all the C code 😞
> 
> As far as I can see `move_files_prepare` actually moves files in some cases (g_file_move succeeds without fallback flag) so the name is misleading. When this fails it calls `copy_move_files` and eventually `g_file_move` again with different flags allowing fallback. In the latter case the callback is still `copy_file_progress_callback`.

What are you doing that you do not see the syncs/progressbar updates happening for USB stick? Can you tell me the file size(s) and what steps you are doing? Seems to be working in whatever scenarios I've been testing.

jeremypw:
@vjr  The freezing mid-bar I previously noted does not seem to be happening now, thanks.  I noticed the new "finishing moving ..." message when moving a large file onto the USB stick.   Copying the same file onto the same stick was *much* slower when replacing an existing copy of the file and behaved differently in that the progress bar completed quickly (a few seconds) but then remained on "Please wait finishing copy ... " for a very long time even though the file transfer seemed complete as judged by the progress bar and the file size reported on the destination (after refresh).  It eventually completed after 10 - 15 minutes.  So the realistic progress updates do not seem to be working in that case.

vjr:
> @vjr The freezing mid-bar I previously noted does not seem to be happening now, thanks. I noticed the new "finishing moving ..." message when moving a large file onto the USB stick. Copying the same file onto the same stick was _much_ slower when replacing an existing copy of the file and behaved differently in that the progress bar completed quickly (a few seconds) but then remained on "Please wait finishing copy ... " for a very long time even though the file transfer seemed complete as judged by the progress bar and the file size reported on the destination (after refresh). It eventually completed after 10 - 15 minutes. So the realistic progress updates do not seem to be working in that case.

@jeremypw not sure how this can be addressed. with my slow 8gb usb stick if i transfer (copy/move/overwrite) a file of say 6gb it previously **_appeared_** to finish quickly in some cases, but with this branch it will also take 10-15 minutes because the real transfer (write) speed of the usb stick is only 7-8 MBps so the duration is expected - previously it would give the user a false sense that the operation has safely compeleted.

ProgramCrafter:
Should the code use `fdatasync` rather than `fsync` in the middle of the transfer, possibly?

vjr:
> Should the code use `fdatasync` rather than `fsync` in the middle of the transfer, possibly?

Hi yes thanks I did see the manpage earlier fdatasync seems a little better, fewer writes, especially reduced wear and tear on flash media, it excludes metadata I understand, until the final sync and close. 

I guess we can just switch to it, do you have any additional insight into pros and cons beyond the docs?

jeremypw:
@vjr I would try to resist the temptation to tinker with things not essential to the PR such as the wording of the progress dialog - it just clouds the issue.  Just concentrate on fixing the linked issue(s). We can polish other features in subsequent PRs.

jeremypw:
Could you add some comments to explain the difference (if any) between `destination` and `destination_for_progress_dialog` and their relation to `target_dir` in `marlin_file_operations`.  Its best to add too many comments than not enough when the changes are not immediately obvious.

libcore/FileUtils.vala
```diff
@@ -433,6 +433,22 @@ namespace Files.FileUtils {
...
+        }
+        var path = file.get_path ();
+        if (path == null) {
+            return;
+        }
+        int fd = Posix.open (path, Posix.O_RDONLY);
+        if (fd < 0) {
+            return;
+        }
+        Posix.fsync (fd);
```
jeremypw, line 451: Missing whitespace after if clause.

vjr: Addressed in https://github.com/elementary/files/pull/2828/commits/96dea33e1e2e35e7ffc2f9eb4af1380c2e1df107

vjr:
> @vjr I would try to resist the temptation to tinker with things not essential to the PR such as the wording of the progress dialog - it just clouds the issue. Just concentrate on fixing the linked issue(s). We can polish other features in subsequent PRs.

OK agreed, I've reverted what I think are all the relevant commits you refer to in these new commits: https://github.com/elementary/files/pull/2828/commits/055dc3e04777867534c5074b7c0645844c3d421f, https://github.com/elementary/files/pull/2828/commits/3a2678256f0c8f5a62ef777d64d02e32326b1c7f and https://github.com/elementary/files/pull/2828/commits/e518d4532d4d6d38eb1632fed0670f099c45d4d5.

vjr:
> Could you add some comments to explain the difference (if any) between `destination` and `destination_for_progress_dialog` and their relation to `target_dir` in `marlin_file_operations`. Its best to add too many comments than not enough when the changes are not immediately obvious.

@jeremypw comment added in https://github.com/elementary/files/pull/2828/commits/3a929f0f589add27a3b656e6d5cc625cad6322ed and also addressed your other feedback to avoid modifying the progess dialog texts.

libcore/FileOperations/CopyMoveJob.vala
```diff
@@ -19,7 +19,11 @@
 public class Files.FileOperations.CopyMoveJob : CommonJob {
     protected bool is_move = false;
     protected GLib.List<GLib.File> files;
-    protected GLib.File? destination;
+    protected unowned GLib.File destination;
```
jeremypw, line 22: Shouldn't this still be nullable?  It is set to null in marlin-file-operations when duplicating i.e. copying into same directory.

libcore/FileOperations/CopyMoveJob.vala
```diff
@@ -102,18 +106,22 @@ public class Files.FileOperations.CopyMoveJob : CommonJob {
...
         /* See https://github.com/elementary/files/issues/464. The job data may become invalid, possibly
          * due to a race. */
-        if (files.data == null || destination == null) {
+        if (files.data == null) {
             return;
         }
 
         var srcname = FileUtils.custom_basename_from_file (files.data);
-        var destname = FileUtils.custom_basename_from_file (destination);
+        var destname = "";
```
jeremypw, line 121: Can't we set destname once at job creation since the destination is not going to change for the lifetime of the job?

libcore/marlin-file-operations.c
```diff
@@ -1639,6 +1653,10 @@ copy_move_file (FilesFileOperationsCopyMoveJob *copy_job,
 
 retry:
 
+    if (!g_file_equal(g_file_get_parent(src), dest_dir)) {
```
jeremypw, line 1656: Can't we do this inside the job creation Vala code?

jeremypw (approved):
There are still some questions about coding but it seems to work well nevertheless.  I am inclined to get it merged and we can push clean up / simplification PRs later as required.  I hope to move some more C code into the CopyMoveJob class soon anyway.

flodavid:
Just testing this and it is very nice, thanks @vjr

PR #2832, opened 2026-09-14
Disambiguate "%u folder"

jeremypw:
Running `io.elementary.files-update-po` gives the following warning
```
xgettext: warning: msgid '%u folder' is used without plural and with plural.
                   src/View/Widgets/OverlayBar.vala:180: Here is the occurrence without plural.
                   src/Dialogs/PropertiesWindow.vala:1283: Here is the occurrence with plural.
                   Workaround: If the msgid is a sentence, change the wording of the sentence; otherwise, use contexts for disambiguation.
```

This fixes that.


jeremypw:
On reflection this might cause issues with languages that use the singular form for other numbers so returning to draft.

jeremypw:
Using `ngettext` in both places should be better.

src/View/Widgets/OverlayBar.vala
```diff
@@ -167,8 +167,8 @@ namespace Files.View {
                 }
             } else { /* Multiple selection. */
                 var fsize = format_size (files_size);
+                str = (ngettext ("%u folder", "%u folders", folders_count)).printf (folders_count);
```
ryonakano:
```suggestion
                str = ngettext ("%u folder", "%u folders", folders_count).printf (folders_count);
```

I don't think we need this set of parentheses.

tintou:
I think that there is still the issue of the whole sentence being made out of concatenation not working for all languages I think

jeremypw:
@tintou Agreed.  I may as well try and fix that here as well.

jeremypw:
Keeping everything translatable into all languages is something of a challenge given the variablity of the status messages.  I have tried to minimise the different msgids while maintaining flexibility for the translators.  Fairly extensive comments are added. I assume that translators are familiar with positional arguments.

We may have to more radically change the layout of the status message in English to fully achieve translatability but I have kept fairly close to the existing layout for now.

src/View/Widgets/OverlayBar.vala
```diff
@@ -151,9 +151,14 @@ namespace Files.View {
                     if (goffile.format_size == "" ) { /* No need to keep recalculating the formatted size. */
                         goffile.format_size = format_size (PropertiesWindow.file_real_size (goffile));
                     }
-                    str = "%s - %s (%s)".printf (goffile.info.get_name (),
-                                                 goffile.formated_type,
-                                                 goffile.format_size);
+
+                    ///TRANSLATORS positional arguments are filename, formatted filetype, formatted filesize
+                    /// filetype is formatted by GLib.ContentType.get_description ()
+                    str = _("%1$s - %2$s (%3$s)").printf (
```
ryonakano: In general we don't need to use `n$` in the source string; translators uses them if necessary.

jeremypw: OK, but it has the advantage of being able to refer to arguments by number in the comments so the translator does not need to inspect the code.

ryonakano: I think we can achieve that by writing something like "the first argument means foobarbaz……" even if we don't use `n$`.

jeremypw: Yes, but it would mean longer comments and less easy for many arguments.

I am not sure what the advantage of removing the positional arguments or the disadvantage of keeping them is but if you feel strongly about it I will remove.

src/View/Widgets/OverlayBar.vala
```diff
@@ -162,32 +167,38 @@ namespace Files.View {
...
+                        files_count,
+                        fsize
+                    );
+                } else if (files_count == 0) {
+                    ///TRANSLATORS arguments are in order: folder count, diskspace used
+                    /// The count is in the form "<number> folders" (translated with ngettext)
+                    /// The diskspace is formatted with GLib.format_size
+                  str = ngettext ("%u folder selected", "%u folders selected", folders_count).printf (
+                        folders_count
+                    );
```
ryonakano, line 188:
```suggestion
                    str = ngettext ("%u folder selected", "%u folders selected", folders_count).printf (
                        folders_count
                    );
```

Malformed indentation.

src/View/Widgets/OverlayBar.vala
```diff
@@ -162,32 +167,38 @@ namespace Files.View {
...
+                if (folders_count == 0) {
+                    ///TRANSLATORS arguments are in order: file count, diskspace used
+                    /// The count is in the form "<number> files" (translated with ngettext)
+                    /// The diskspace is formatted with GLib.format_size
+                    str = ngettext ("%u file selected (%s)", "%u files selected (%s)", files_count).printf (
+                        files_count,
+                        fsize
+                    );
+                } else if (files_count == 0) {
+                    ///TRANSLATORS arguments are in order: folder count, diskspace used
```
ryonakano: This comment says the string contains diskspace used, but it's actually not: `%u folder selected`. Which is correct?

src/View/Widgets/OverlayBar.vala
```diff
@@ -162,32 +167,38 @@ namespace Files.View {
...
+                    /// The count is in the form "<number> folders" (translated with ngettext)
+                    /// The diskspace is formatted with GLib.format_size
+                  str = ngettext ("%u folder selected", "%u folders selected", folders_count).printf (
+                        folders_count
+                    );
+                } else {
+                    ///TRANSLATOR arguments are folder count, file count
+                    /// The counts are in the form "<number> files" (translated with ngettext)
+                    str = _("%s and %s selected").printf (
+                        ngettext ("%u folder ", "%u folders", folders_count).printf (folders_count),
```
ryonakano:
```suggestion
                        ngettext ("%u folder", "%u folders", folders_count).printf (folders_count),
```

Extra whitespace.

src/View/Widgets/OverlayBar.vala
jeremypw: Oops, don't know where those comments came from!  Now removed.

jeremypw:
@ryonakano Thanks for the review - sorry about the careless mistakes. Hopefully now fixed.

src/View/Widgets/OverlayBar.vala
```diff
@@ -162,32 +167,30 @@ namespace Files.View {
...
-                    str = _("%u items selected (%s)").printf (files_count, fsize);
+                if (folders_count == 0) {
+                    str = ngettext ("%u file selected (%s)", "%u files selected (%s)", files_count).printf (
+                        files_count,
+                        fsize
+                    );
+                } else if (files_count == 0) {
+                  str = ngettext ("%u folder selected", "%u folders selected", folders_count).printf (
+                        folders_count
+                    );
```
ryonakano:
```suggestion
                    str = ngettext ("%u folder selected", "%u folders selected", folders_count).printf (
                        folders_count
                    );
```

The malformed indentation still presents.

jeremypw:
@ryonakano Sorry, I thought I applied your suggestion last time.  Something must have gone wrong.  I have applied it this time for sure.

ryonakano (approved):

PR #2834, opened 2026-09-16
Fix unmount criticals

vjr:
Fix the following during unmount:

```
(io.elementary.files:371186): GLib-GObject-CRITICAL **: 23:48:11.625: invalid (NULL) pointer instance

(io.elementary.files:371186): GLib-GObject-CRITICAL **: 23:48:11.625: g_signal_emit_by_name: assertion 'G_TYPE_CHECK_INSTANCE (instance)' failed

(io.elementary.files:371186): Gtk-CRITICAL **: 23:48:11.625: gtk_widget_destroy: assertion 'GTK_IS_WIDGET (widget)' failed
```


jeremypw (dismissed):
You don't really need an early return here, you can save a line by testing for `dialog != null`

jeremypw (approved):
Thanks!

PR #2838, opened 2026-09-19
Update free space after emptying trash

### LINKED ISSUE #2836: The disk space bar doesn't update after permanently deleting trash
### What Happened?

I right click on the Trash item in the side bar and select Permanently Delete Trash. 
But the bar under the influenced drive which indicates available disk space doesn't update. 
It does update when I press <kbd>ctrl</kbd>+<kbd>R</kbd>. 

### Steps to Reproduce

1. Move a significant amount of stuff to trash so that clearing it influences the disk space bar
2. In Files sidebar, right click on the Trash item
3. Select Permanently Delete Trash. 

### Expected Behavior

The disk space bar under the influenced drive updates. 

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #2836 

 - Introduce ViewWindowInterface in libcore so we can use required function
 -  Signal free space change after emptying trash in EmptyTrashJob


danirabbit:
This seems odd. Why are we requiring individual operations to signal disk space changes rather than listening to disk space changes directly?

jeremypw:
@danirabbit Hmm, presumably some historical reason.  I just followed existing code in this PR to keep it simple.  I'll review and see if its possible to change the whole method of keeping sidebar up to date.

jeremypw:
As I recall, there is no signal from the device to connect to that will tell when the space changes, you have to poll it at regular intervals to report its usage stats.  I guess it was considered not worth spending cpu cycles polling when most of the time the space is not changing.  The drawback is, off course, the current way does not detect changes made by other apps unless the folder containing the changes is displayed in Files.  It might be consider adding a global polling of devices at a suitable modest rate (every few seconds) I guess.  It would only require adding a continuous Timeout which fires off the `free_space_change` signal.  Probably worth keeping the existing code to get immediate response to Files own actions?

danirabbit:
It seems like maybe we can get this from UDisks? https://valadoc.org/udisks2/UDisks.Drive.size.html

jeremypw:
Not sure, but I don't think UDisks provides a signal when the value changes? In which case we would still have to poll. It would work better if we were ever to Flatpak Files I guess.

